PR #2, opened 2024-06-04
Add workflows for build and release

hallipr:
The build pipeline will add a dynamic version suffix to the pipeline and run build, test and pack.   On non-PR runs, it will publish the alpha versioned package to the internal GitHub NuGet feed.

The release pipeline is queued whenever a new release is created in GitHub.  It doesn't use any dynamic versioning, relying on the version committed to the repo.   It will build, test, pack and release the package to GitHub.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: Pack
+      - name: Build and Pack
         run: dotnet pack
           --configuration Release
           --output "${{github.workspace}}/artifacts/packages"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ jobs:
           dotnet-version: '8.x' # SDK Version to use.
 
       # Pack the client nuget package and include urls back to the repository and release tag
-      - name: Pack
+      - name: Build and Pack
         run: dotnet pack
           --configuration Release
           --output "${{github.workspace}}/artifacts/packages"


.github/workflows/main.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 8.x
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Build
+        run: dotnet build
```
weshaggard: Does github actions run command automatically default to single line across multiple?

.github/workflows/main.yml
```diff
@@ -0,0 +1,67 @@
...
+          ${{ env.version_suffix_args }}
+
+      - name: Test
+        run: dotnet test
+          --no-build
+          --configuration Release
+          --filter="TestCategory~${{ github.event_name == 'pull_request' && 'Offline' || 'Online' }}"
+          --logger "trx;LogFileName=${{github.workspace}}/artifacts/test-results/full.trx"
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
weshaggard, line 37: Given this isn't setup this is going to fail should we just comment out this step for now?

.github/workflows/main.yml
weshaggard: Any reason to not just call dotnet pack here instead of having build and pack in different steps?

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
+name: Release package
+
+on:
+  release:
+    types: [published]
```
weshaggard, line 5: What triggers this?

.github/workflows/main.yml
weshaggard: I guess if there is nothing marked Offline this will just not run anything yet.

.github/workflows/main.yml
```diff
@@ -0,0 +1,67 @@
...
+          name: build-artifacts
+          path: ${{github.workspace}}/artifacts
+
+      # - name: NuGet Autenticate
+      #   if: github.event_name != 'pull_request'
+      #   run: dotnet nuget add source
+      #     "https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json"
+      #     --name "github"
+      #     --username ${{ github.actor }}
+      #     --password ${{ secrets.GITHUB_TOKEN }}
```
weshaggard: Does this already exist for normal workflows? If so and it works I think we can enable this step.

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+        with:
+          name: build-artifacts
+          path: ${{github.workspace}}/artifacts
+
+      # - name: NuGet Autenticate
+      #   run: dotnet nuget add source
+      #     "https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json"
+      #     --name "github"
+      #     --username ${{ github.actor }}
+      #     --password ${{ secrets.GITHUB_TOKEN }}
```
weshaggard: This will need to be replaced with the real secret to nuget.org once we get that setup.

.github/workflows/main.yml
```diff
@@ -0,0 +1,67 @@
...
+      #     --store-password-in-clear-text
+
+      # - name: Publish
+      #   if: github.event_name != 'pull_request'
+      #   run: dotnet nuget push
+      #     ${{github.workspace}}/artifacts/packages/*.nupkg
```
weshaggard: I suspect we will want to also publish the snpukg which contains the symbols.

.github/workflows/main.yml
hallipr: It does parse the yaml strings that way.  Indented lines under a string appear to have their whitespace removed.   I need to do more tests to see if it keeps the line breaks in the command and how different shells handle the command.  i.e. if passed to pwsh, the multiline command may break.

.github/workflows/main.yml
hallipr: The current code has opt-in to PR testing and opt-in to CI testing using the Offline or Online categories respectively.

We do have examples with `[Test]` attributes that require the environment variable `OPENAI_API_KEY`.
https://github.com/openai/openai-dotnet/blob/main/examples/Chat/Example03_FunctionCalling.cs#L55

My assumption was that once we had a key, we would enable the live test in CI runs by either targeting all tests without the "Offline" category, or by adding the "Online" category to the test.

.github/workflows/main.yml
hallipr: I'm used to build and testing in debug, then packing and live testing in release. Being that this is all release, if the chances of build success with pack failure are low, we can combine them.

.github/workflows/release.yml
hallipr: This is triggered on creation of a release.

.github/workflows/main.yml
hallipr: We can enable it.   It's just unnecessary without the following publish step.  I wanted to confirm that we wanted to publish alpha packages to the local feed before enabling that step.

.github/workflows/main.yml
hallipr: We don't currently produce a snupkg.  Docs suggest that publishing a `nupkg` will automatically publish a corresponding `snupkg` if it's in the same directory.  Should we update the csproj or the `dotnet pack` command to create the package?
I'm leaning toward updating the csproj to keep the workflow commands as simple as possible

.github/workflows/release.yml
hallipr: In testing, I was manually creating a release in the GitHub UI when I wanted to produce a non-alpha package.  I've seen another flow that will use the CI builds to maintain a draft GitHub release, so the releaser only needs to publish the current draft release to cut a new release.

.github/workflows/main.yml
weshaggard: Yeah dotnet pack will do a build and package. So I would just combine them as the package part is not likely to fail.

.github/workflows/main.yml
weshaggard: Ok lets do that at a later time then.

.github/workflows/main.yml
weshaggard: We should do some testing to see if we get symbols after we get publishing setup as we will want debugging symbols and things like sourcelink to be working for the package.

.github/workflows/main.yml
weshaggard: OK I'm good with using filtering for this.

.github/workflows/release.yml
weshaggard: Ok I think triggering on the manual creation of a release should work. We should add that information the README

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+      #     --store-password-in-clear-text
+
+      # - name: Publish
+      #   run: dotnet nuget push
+      #     ${{github.workspace}}/artifacts/packages/*.nupkg
+      #     --source "github"
+      #     --api-key ${{ secrets.GITHUB_TOKEN }}
```
weshaggard:
```suggestion
      #     --api-key ${{ secrets.NUGET_API_KEY }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+      #     --skip-duplicate
+      #   working-directory: .dotnet
```
weshaggard: Why does this need to be in the .dotnet working directory?

.github/workflows/main.yml
```diff
@@ -22,9 +22,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: Build
-        run: dotnet build
+      - name: Pack
```
weshaggard:
```suggestion
      - name: Build and Pack
```

.github/workflows/release.yml
```diff
@@ -16,9 +16,13 @@ jobs:
         with:
           dotnet-version: '8.x' # SDK Version to use.
 
-      - name: Build
-        run: dotnet build
+      # Pack the client nuget package and include urls back to the repository and release tag
+      - name: Pack
```
weshaggard:
```suggestion
      - name: Build and Pack
```

.github/workflows/release.yml
hallipr: updated

.github/workflows/release.yml
hallipr: updated

PR #14, opened 2024-06-06
Revert AssemblyInfo.cs removal; file needed for Azure OpenAI support

trrwilson:
This `[InternalsVisibleTo]` attribute likely appeared to be a development-time accidental artifact. It's the interim mechanism that the Azure.AI.OpenAI package will use for Azure OpenAI Service support, though, so we'll need it back.


weshaggard:
After we enabled signing, we removed this because we need a full strong name key for this to work. So, if this is needed, we also need to include the key. However, I'd like to understand what our plan is for this.

src/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Azure.AI.OpenAI")]
```
weshaggard:
```suggestion
[assembly: InternalsVisibleTo("Azure.AI.OpenAI, PublicKey=0024000004800000940000000602000000240000525341310004000001000100097ad52abbeaa2e1a1982747cc0106534f65cfea6707eaed696a3a63daea80de2512746801a7e47f88e7781e71af960d89ba2e25561f70b0e2dbc93319e0af1961a719ccf5a4d28709b2b57a5d29b7c09dc8d269a490ebe2651c4b6e6738c27c5fb2c02469fe9757f0a3479ac310d6588a50a28d7dd431b907fd325e18b9e8ed")]
```
```suggestion
[assembly: InternalsVisibleTo("Azure.AI.OpenAI")]
```

weshaggard (approved):

PR #19, opened 2024-06-07
refactor: Moved InternalsVisibleTo to .csproj.

HavenDV:


weshaggard:
@HavenDV is there an advantage to doing this in the csproj over the cs file? IIRC the csproj just generate a file on your behalf which to me doesn't really add any value and in fact causes more temp files to be created.

HavenDV:
No big difference, except that we will get rid of this file and all project settings will be stored in one place
I see the benefit in that lately you almost never expect an AssemblyInfo file to be present in a project, because now it is almost always generated, and it is better to put it in the project file.
I don't think generation has even a small impact on build time, besides it is generated for other attributes anyway

Feel free to just decline this, it's just a suggestion

Ste1io:
+1 for this PR, as it is the more modern approach.

src/OpenAI.csproj
```diff
@@ -48,4 +48,10 @@
     <PackageReference Include="System.ClientModel" Version="1.1.0-beta.4" />
     <PackageReference Include="System.Text.Json" Version="8.0.2" />
   </ItemGroup>
+
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
```
Ste1io: Consider using the [`<InternalsVisibleTo>`](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#internalsvisibleto) item, introduced in dotnet 5.

src/OpenAI.csproj
```diff
@@ -48,4 +48,8 @@
     <PackageReference Include="System.ClientModel" Version="1.1.0-beta.4" />
     <PackageReference Include="System.Text.Json" Version="8.0.2" />
   </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="Azure.AI.OpenAI, PublicKey=0024000004800000940000000602000000240000525341310004000001000100097ad52abbeaa2e1a1982747cc0106534f65cfea6707eaed696a3a63daea80de2512746801a7e47f88e7781e71af960d89ba2e25561f70b0e2dbc93319e0af1961a719ccf5a4d28709b2b57a5d29b7c09dc8d269a490ebe2651c4b6e6738c27c5fb2c02469fe9757f0a3479ac310d6588a50a28d7dd431b907fd325e18b9e8ed" />
+  </ItemGroup>
```
weshaggard:
```suggestion
  <ItemGroup>
      <ItemGroup>
    <InternalsVisibleTo Include="Azure.AI.OpenAI">
      <PublicKey>0024000004800000940000000602000000240000525341310004000001000100097ad52abbeaa2e1a1982747cc0106534f65cfea6707eaed696a3a63daea80de2512746801a7e47f88e7781e71af960d89ba2e25561f70b0e2dbc93319e0af1961a719ccf5a4d28709b2b57a5d29b7c09dc8d269a490ebe2651c4b6e6738c27c5fb2c02469fe9757f0a3479ac310d6588a50a28d7dd431b907fd325e18b9e8ed</PublicKey>
    </InternalsVisibleTo>
  </ItemGroup>
  </ItemGroup>
```

weshaggard:
Thanks for the contribution and I buy the argument that this file is getting generated for other attributes already so there isn't any real reason to not add this one. I made one suggestion which makes separates the public key into its own property.

src/OpenAI.csproj
HavenDV: Not sure about the name for the metadata for PublicKey, according to the documentation it's just Key?
image

But I'll try to find examples of real use now

HavenDV: I tested, both options work

HavenDV: I wasn’t sure that Suggestion had the correct formatting (it seems not, judging by the number of spaces), so I made it a separate commit

weshaggard (approved):

PR #32, opened 2024-06-10
Use cached/compiled Regex in InternalChatCompletionRequestMessageContentPartImageImageUrl

stephentoub:


src/Custom/Chat/Internal/InternalChatCompletionRequestMessageContentPartImageImageUrl.cs
```diff
@@ -8,6 +8,8 @@ namespace OpenAI.Chat;
 [CodeGenSuppress("InternalChatCompletionRequestMessageContentPartImageImageUrl", typeof(string))]
 internal partial class InternalChatCompletionRequestMessageContentPartImageImageUrl
 {
+    private static readonly Regex s_parseDataUriRegex = new(@"^data:(?<type>.+?);base64,(?<data>.+)$", RegexOptions.Compiled);
```
stephentoub, line 16: Note that when the library multitargets .NET 8, this should be updated to use the regex source generator.

KrzysztofCwalina (approved):

PR #33, opened 2024-06-10
Replace custom SSE reader with source for System.Net.ServerSentEvents

stephentoub:
A stable package for the new System.Net.ServerSentEvents library isn't yet available, but until one is we can switch the code over to using a source copy.

The natural replacement for `ServerSentEvent` is `SseItem<string>`, but I used `SseItem<byte[]>` instead just to avoid an unnecessary UTF8->UTF16->UTF8 roundtrip. It could be taken further in the future to avoid allocation by doing more of the parsing inside of the callback passed to SseParser.


src/OpenAI.csproj
```diff
@@ -28,7 +28,12 @@
    
     <!-- Embed source files that are not tracked by the source control manager in the PDB -->
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
-</PropertyGroup>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <!-- Allow use of unsafe code, for System.Net.ServerSentEvents polyfill on netstandard2.0 -->
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
KrzysztofCwalina, line 48: Could you add a comment and file an issue to remove this once we switch to using the oob package?

stephentoub: Yup.

stephentoub: Done. Opened https://github.com/openai/openai-dotnet/issues/41 and updated the csproj with a comment.

KrzysztofCwalina (approved):

src/Utility/System.Net.ServerSentEvents.cs
```diff
@@ -0,0 +1,620 @@
...
+using System.Buffers;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
+using System.IO;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+using System.Threading;
+
```
trrwilson, line 26:
```suggestion

#nullable enable

```

src/Utility/System.Net.ServerSentEvents.cs
```diff
@@ -0,0 +1,620 @@
...
+// This file contains a source copy of:
+// https://github.com/dotnet/runtime/tree/2bd15868f12ace7cee9999af61d5c130b2603f04/src/libraries/System.Net.ServerSentEvents/src/System/Net/ServerSentEvents
+// Once the System.Net.ServerSentEvents package is available, this file should be removed and replaced with a package reference.
+//
+// The only changes made to this code from the original are:
+// - Put into a single file for ease of management (it should not be edited in this repo).
+// - Changed public types to be internal.
+// - Removed a use of a [NotNull] attribute to assist in netstandard2.0 compilation.
+// - Replaced a reference to a .resx string with an inline constant.
```
trrwilson, line 13:
```suggestion
// - Enabled nullable reference types at file scope
// - Replaced a reference to a .resx string with an inline constant.
```

PR #36, opened 2024-06-10
Fix non-thread-safe use of Random in MultipartFormDataBinaryContent

stephentoub:
MultipartFormDataBinaryContent has a static Random instance that's being used without a lock. This locks around each access (an alternative would be to use and lazily-initialize on each access a [ThreadStatic] Random).

While fixing that, also improved a few things:
- We don't need to create a char[] for the boundary values; the const string can be used directly.
- For netstandard2.0, we don't need to allocate a char[]; a stackalloc'd span can be used directly.
- For .NET, we don't need a stackalloc, either; we can use string.Create and write directly into the new string instance.
- For .NET, we don't need a dedicated Random instance, either; we can use Random.Shared.
- For .NET, Random allows writing into a byte span, so we don't need the byte[] allocation either and can stackalloc it.
- For .NET, just %'ing by the string length produces as good or better code than using a manually-computed mask.

### DIFF
--- a/src/Utility/MultipartFormDataBinaryContent.cs
+++ b/src/Utility/MultipartFormDataBinaryContent.cs
@@ -120,7 +120,7 @@ internal class MultipartFormDataBinaryContent : BinaryContent
         content.Headers.ContentType = header;
     }
 
-#if NET
+#if NET6_0_OR_GREATER
     private static string CreateBoundary() =>
         string.Create(BoundaryLength, 0, (chars, _) =>
         {
@@ -145,14 +145,12 @@ internal class MultipartFormDataBinaryContent : BinaryContent
             _random.NextBytes(random);
         }
 
-        // The following will sample evenly from the possible values.
-        // This is important to ensuring that the odds of creating a boundary
-        // that occurs in any content part are astronomically small.
+        // Instead of `% BoundaryValues.Length` as is used above, use a mask to achieve the same result.
+        // `% BoundaryValues.Length` is optimized to the equivalent on .NET Core but not on .NET Framework.
         const int Mask = 255 >> 2;
-
         Debug.Assert(BoundaryValues.Length - 1 == Mask);
 
-        for (int i = 0; i < random.Length; i++)
+        for (int i = 0; i < chars.Length; i++)
         {
             chars[i] = BoundaryValues[random[i] & Mask];
         }
@@ -177,19 +175,18 @@ internal class MultipartFormDataBinaryContent : BinaryContent
 
     public override void WriteTo(Stream stream, CancellationToken cancellationToken = default)
     {
-        // TODO: polyfill sync-over-async for netstandard2.0 for Azure clients.
-        // Tracked by https://github.com/Azure/azure-sdk-for-net/issues/42674
-
-#if NET
+#if NET5_0_OR_GREATER
         _multipartContent.CopyTo(stream, default, cancellationToken);
 #else
+        // TODO: polyfill sync-over-async for netstandard2.0 for Azure clients.
+        // Tracked by https://github.com/Azure/azure-sdk-for-net/issues/42674
         _multipartContent.CopyToAsync(stream).GetAwaiter().GetResult();
 #endif
     }
 
     public override async Task WriteToAsync(Stream stream, CancellationToken cancellationToken = default)
     {
-#if NET
+#if NET5_0_OR_GREATER
         await _multipartContent.CopyToAsync(stream, cancellationToken).ConfigureAwait(false);
 #else
         await _multipartContent.CopyToAsync(stream).ConfigureAwait(false);
--- a/src/Utility/MultipartFormDataBinaryContent.cs
+++ b/src/Utility/MultipartFormDataBinaryContent.cs
@@ -14,8 +14,8 @@ internal class MultipartFormDataBinaryContent : BinaryContent
 {
     private readonly MultipartFormDataContent _multipartContent;
 
-    private static Random _random = new();
-    private static readonly char[] _boundaryValues = "0123456789=ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz".ToCharArray();
+    private const int BoundaryLength = 70;
+    private const string BoundaryValues = "0123456789=ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz";
 
     public MultipartFormDataBinaryContent()
     {
@@ -120,27 +120,46 @@ internal class MultipartFormDataBinaryContent : BinaryContent
         content.Headers.ContentType = header;
     }
 
+#if NET
+    private static string CreateBoundary() =>
+        string.Create(BoundaryLength, 0, (chars, _) =>
+        {
+            Span<byte> random = stackalloc byte[BoundaryLength];
+            Random.Shared.NextBytes(random);
+
+            for (int i = 0; i < chars.Length; i++)
+            {
+                chars[i] = BoundaryValues[random[i] % BoundaryValues.Length];
+            }
+        });
+#else
+    private static readonly Random _random = new();
+
     private static string CreateBoundary()
     {
-        Span<char> chars = new char[70];
+        Span<char> chars = stackalloc char[BoundaryLength];
 
-        byte[] random = new byte[70];
-        _random.NextBytes(random);
+        byte[] random = new byte[BoundaryLength];
+        lock (_random)
+        {
+            _random.NextBytes(random);
+        }
 
         // The following will sample evenly from the possible values.
         // This is important to ensuring that the odds of creating a boundary
         // that occurs in any content part are astronomically small.
-        int mask = 255 >> 2;
+        const int Mask = 255 >> 2;
 
-        Debug.Assert(_boundaryValues.Length - 1 == mask);
+        Debug.Assert(BoundaryValues.Length - 1 == Mask);
 
-        for (int i = 0; i < 70; i++)
+        for (int i = 0; i < random.Length; i++)
         {
-            chars[i] = _boundaryValues[random[i] & mask];
+            chars[i] = BoundaryValues[random[i] & Mask];
         }
 
         return chars.ToString();
     }
+#endif
 
     public override bool TryComputeLength(out long length)
     {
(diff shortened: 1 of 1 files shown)


KrzysztofCwalina:
cc: @annelo-msft

src/Utility/MultipartFormDataBinaryContent.cs
```diff
@@ -161,19 +180,19 @@ public override void WriteTo(Stream stream, CancellationToken cancellationToken
         // TODO: polyfill sync-over-async for netstandard2.0 for Azure clients.
         // Tracked by https://github.com/Azure/azure-sdk-for-net/issues/42674
 
-#if NET6_0_OR_GREATER
+#if NET
```
annelo-msft: Curious: why is this pattern preferable to using `NET6_0_OR_GREATER`?

stephentoub: Just for consistency and brevity. It's fine to want to be precise, but then NET6_0_OR_GREATER is arguably "incorrect" here, as the guarded API was introduced earlier than that; it's just guarding it with NET6_0_OR_GREATER then because that's the oldest version of core the library targets, but then you can also just use the simpler/shorter "NET", which has the same effect. I can put it back if you'd prefer, though.

src/Utility/MultipartFormDataBinaryContent.cs
```diff
@@ -120,27 +120,46 @@ public static void AddContentTypeHeader(HttpContent content, string contentType)
...
+    private static string CreateBoundary() =>
+        string.Create(BoundaryLength, 0, (chars, _) =>
+        {
+            Span<byte> random = stackalloc byte[BoundaryLength];
+            Random.Shared.NextBytes(random);
+
+            for (int i = 0; i < chars.Length; i++)
+            {
+                chars[i] = BoundaryValues[random[i] % BoundaryValues.Length];
```
annelo-msft, line 132: I expect `random[i] % BoundaryValues.Length` is equivalent to the operation using `Mask` below -- is there an advantage to doing it one way or the other?  If so, would it make sense to use the same approach in both places to make the code easier to understand?

src/Utility/MultipartFormDataBinaryContent.cs
```diff
@@ -120,27 +120,46 @@ public static void AddContentTypeHeader(HttpContent content, string contentType)
...
+    private static readonly Random _random = new();
+
     private static string CreateBoundary()
     {
-        Span<char> chars = new char[70];
+        Span<char> chars = stackalloc char[BoundaryLength];
 
-        byte[] random = new byte[70];
-        _random.NextBytes(random);
+        byte[] random = new byte[BoundaryLength];
```
annelo-msft, line 142: Would it make sense to use `stackalloc` here as well?

src/Utility/MultipartFormDataBinaryContent.cs
```diff
@@ -120,27 +120,46 @@ public static void AddContentTypeHeader(HttpContent content, string contentType)
...
         // This is important to ensuring that the odds of creating a boundary
         // that occurs in any content part are astronomically small.
-        int mask = 255 >> 2;
+        const int Mask = 255 >> 2;
 
-        Debug.Assert(_boundaryValues.Length - 1 == mask);
+        Debug.Assert(BoundaryValues.Length - 1 == Mask);
 
-        for (int i = 0; i < 70; i++)
+        for (int i = 0; i < random.Length; i++)
```
annelo-msft:
```suggestion
        for (int i = 0; i < BoundaryLength; i++)
```

src/Utility/MultipartFormDataBinaryContent.cs
stephentoub: We can't on netstandard2.0 and still use Random.NextBytes, which on netstandard2.0 lacks an overload for span / pointers.

src/Utility/MultipartFormDataBinaryContent.cs
stephentoub: > I expect random[i] % BoundaryValues.Length is equivalent to the operation using Mask below

On .NET Core, yes. You can see here that the two methods produce identical assembly:
[SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4axAMxNSDIQwDeNBqYZiovBdgwwGkLrgwNePBgCEI3ACbYoATwAZGC4AcwwACwYAXgYAdjoNajNzS2tbewhHZ2ZGTx8/fwA1bEkOVRiGACI6clItFABWJDiADgBOaIBBdyEAEQBRADEAcQAJAEkAKQBpQIBZADkAeQAFAEU2WQAVAFUigHUADQBNAC0AfWxgMG8YADNQiN4AKwBrSXwuCDEARygnBwFAB3BD+ABeVSSKRMZm0TBQDHm5AAFLIxNguAAeMARPwAPnseIBAEpYaZjMkUmZ0ZiscB/LZCVBMd5RJUnNgwG9SpIIGAGAzbABtfJcXwBYJhSIAXWh1NMbFZoiGUkkKJZ4tEJPlCvu0AYKNczl4lUSLgYWKJflwLCl4QiahcAGpnWSqQrKQqFbibcLeDLKmKJcVSuVcMLNWz8P7AwBSDxecWFEplVR2kIOuXk6kAXxz+Y9DBz8OIiPmpDRGOxvqghNruHdKS93tp2KFMGZyvwHIwXJ5kj5Ao7oqTIftst11KVWvwqsHGu7OpzKX1UENxotsXNpqtDYz0sdLrdK7MLe9ZgbsaDY5TYdUke714AZAwkFps0WUoXvzmc8L5hgSIIG8CZxHVQDgNA8DljEPgsltRYIAmLhJFcVxQhJGUcwsKwbDsZgkARBgZ2jed1TbLF3EZTsGFwasSSMQtcyAA===)

That's unfortunately not the case with an older JIT as exists in .NET Framework, e.g. note the idiv on the left here:
[SharpLab](https://sharplab.io/#v2:EYLgHgbALANALiAhgZwLYB8ACAmAjAWAChMAGAAk1wDoAlAVwDs4BLVAUyoGEB7VAB2YAbNgCcAyqIBuzAMZtkAbiKYAzBWxlOZAN5Ey+snxHNJiOGzIzuDZHDLMmZAELdGAE0QiAngBk2DAHM4AAsyAF4yAHYSJUIDQ2NTc0trWwpcchd3Ty8ANURBOnlwsgAiElxsFSgAVghIgA4ATjCAQSdOABEAUQAxAHEACQBJACkAaR8AWQA5AHkABQBFGjEAFQBVXIB1AA0ATQAtAH1EYBk3NgAzAODmACsAa0FUBm4+AEcRWzpJAHcwF4AF6lWLxPQGVQUKBkKa4AAUYj4iAYAB4ZMFPAA+SyY74ASgh+l0cXiBiRKNRwC85hxIhRbl4JVsiBkjwKgm4MjI1PMAG0sgwPN4/IEQgBdMFk/Q0Bm8XpCQTw+lC3j4qXSq7cERkeEOOzMEoxexkVG4zzIKiioLBBT2ADU9sJpOlJOl0oxFr5zHFJUFwryBSKyD5KsZqG9voApM5XEKcvlCvIrf4bZKiWSAL4Z7MusgZqGYGFTbCI5Foz0iHGV5DO+Ju90UtG8th0uWoZlwVnswSc7ktgVxgPWiUasmy1WoBW95Xt9UZ+JanV6xyGiLGw1mmspsW2h1OhcGBvugw1yN+ocJoPyUPt88AMjIEBU6bz8Vz74zGb5UzYIW4bjDPwSq/v+gHAXMfAsKkVAzNwwwMIIDgOAE+LihmRgmGYFiUBA0JkBO4bTkqTaok4NKtmQyDlviOi5pmQA==)

I could make them both `% BoundaryValues.Length` if we're not concerned about the (arguably inconsequential) overhead on netfx

src/Utility/MultipartFormDataBinaryContent.cs
stephentoub: Another one of those "Steve is trying to microoptimize... do we care?" scenarios.

The .NET Framework JIT doesn't know how to remove the bounds check on BoundaryValues here (the .NET Core one could). There are three bounds checks possible here: one on indexing into chars, one on indexing into BoundaryValues, and one on indexing into random. As it was written before, none could be removed on netfx. By changing it to iterate over the full length of random, at least the bounds check from `random[i]` is removed.

But, it's really not a big deal. If you prefer the source code consistency, I can certainly change it.

KrzysztofCwalina: > Another one of those "Steve is trying to microoptimize... do we care?" scenarios.

@annelo-msft, I think Steve is giving you a green light to push for adding Variant to the BCL :-)

annelo-msft: No, we like optimizations, and thanks for explaining so we can learn!  If you wouldn't mind adding a comment to that effect so the next person who wonders the same thing can have their question answered inline in the code, then I think we're happy to have the optimization.  Many thanks!

src/Utility/MultipartFormDataBinaryContent.cs
annelo-msft: Looks like I'm not able to resolve comments on this PR, but @joseharriaga, I am fine with this, feel free to resolve.

src/Utility/MultipartFormDataBinaryContent.cs
stephentoub: > If you wouldn't mind adding a comment to that effect so the next person who wonders the same thing can have their question answered inline in the code, then I think we're happy to have the optimization. Many thanks!

Okey dokey

src/Utility/MultipartFormDataBinaryContent.cs
annelo-msft: Ah, no, very cool.  If you wouldn't mind adding a comment to that effect, I'm happy to keep the optimization.  Thanks!

src/Utility/MultipartFormDataBinaryContent.cs
annelo-msft: I don't feel strongly, and overall I'm happy to learn.  @KrzysztofCwalina, do you have an opinion?  Our SDK libraries typically only target netstandard2.0 and net6.0, so it's correct in that context, and if we stick with NET6_0_OR_GREATER, it could lead to fewer preprocessor symbols in client code, but I'm open to revisiting what we standardize around with new SCM-based libraries.

KrzysztofCwalina: I don't feel strongly either way, but: a) I prefer being precise with the minimum version we need, i.e. 5.0+, b) I slightly prefer being explicit, i.e. NET_5_0_OR_GREATER; it makes it easier to read the code.

trrwilson (approved):

PR #37, opened 2024-06-10
Remove seemingly unnecessary string allocation from ClientUriBuilder.AppendPath

stephentoub:
There aren't any tests in the repo currently to help me validate this, but it looks like this PathBuilder.ToString() call in ClientUriBuilder.AppendPath is superfluous. Every use I see of ClientUriBuilder ends with a call to ToUri, which will itself call PathBuilder.ToString() and overwrite the UriBuilder's Path. Further, the corresponding ClientUriBuilder.AppendQuery doesn't similarly write to UriBuilder.Query.

### DIFF
--- a/src/Generated/Internal/ClientUriBuilder.cs
+++ b/src/Generated/Internal/ClientUriBuilder.cs
@@ -47,7 +47,6 @@ namespace OpenAI
             }
 
             PathBuilder.Append(value);
-            UriBuilder.Path = PathBuilder.ToString();
         }
 
         public void AppendPath(bool value, bool escape = false)


stephentoub:
> this is a generated file. we need to fix this in the generator. @m-nash

I see now it's in a generated folder. What's it generated from?

trrwilson:
> > this is a generated file. we need to fix this in the generator. @m-nash
> 
> I see now it's in a generated folder. What's it generated from?

I believe it's from this line in the code emitter project:
https://github.com/Azure/autorest.csharp/blob/bd37a920d0a545e06ebb754b6c2fb90b565bc571/src/AutoRest.CSharp/Common/Output/Models/Types/HelperTypeProviders/System/ClientUriBuilderProvider.cs#L162

@m-nash should confirm.

The optimization will be picked up automatically once we regenerate with updated tooling.

stephentoub:
Thanks. I'm curious why this is generated. Does the generated code end up differing from project to project?

PR #38, opened 2024-06-10
Improve Embedding.ConvertToVectorOfFloats

stephentoub:
It's currently incorrect on big endian systems (e.g. IBM s390x), as it blits the little-endian bytes into an array of floats. On big endian the values need to be reversed.

It also assumes the data being sent back from the service is always correct. If it's corrupted in certain ways, such as not being as long as was expected, not being quoted, etc., we can silently get bad data.

It also more allocation than is necessary. It first allocates a string for the base64 string data. Then it allocates a new string with the quotes stripped off. Then it allocates a byte[] with the decoded bytes. And finally it allocates the resulting float[]. We can avoid the initial string by getting the raw memory from the BinaryData. We can avoid the substring by just slicing those bytes. And we can generally avoid the temporary byte[] by renting one from the array pool. That leaves just the required float[].

### DIFF
--- a/src/Custom/Embeddings/Embedding.cs
+++ b/src/Custom/Embeddings/Embedding.cs
@@ -1,5 +1,9 @@
 using System;
+using System.Buffers;
+using System.Buffers.Binary;
+using System.Buffers.Text;
 using System.Collections.Generic;
+using System.Runtime.InteropServices;
 
 namespace OpenAI.Embeddings;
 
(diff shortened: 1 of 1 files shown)


src/Custom/Embeddings/Embedding.cs
```diff
@@ -89,13 +93,43 @@ internal Embedding(int index, BinaryData embeddingProperty, InternalEmbeddingObj
...
+        OperationStatus status = Base64.DecodeFromUtf8(base64, bytes.AsSpan(), out int bytesConsumed, out int bytesWritten);
+        if (status != OperationStatus.Done || bytesWritten % sizeof(float) != 0)
+        {
+            ThrowInvalidData();
+        }
+
+        // Interpret bytes as floats
+        float[] vector = new float[bytesWritten / sizeof(float)];
+        bytes.AsSpan(0, bytesWritten).CopyTo(MemoryMarshal.AsBytes(vector.AsSpan()));
+        if (!BitConverter.IsLittleEndian)
```
KrzysztofCwalina, line 116: Are you sure the data comes big-endian from the server? When I was prototyping EmbeddingVector, I thought I saw it being little-endian. I was surprised myself. @ShivangiReja, do you remember.

stephentoub: I believe it is little-endian. That's why if the machine is big endian we need to reverse the data.

KrzysztofCwalina: Ah, sorry. Working on a small laptop screen without glasses and I missed the "!"

KrzysztofCwalina (approved):

PR #47, opened 2024-06-11
Address some more analyzer warnings

stephentoub:


src/Custom/OpenAIError.cs
```diff
@@ -35,10 +35,10 @@ OpenAIErrorResponse errorResponse
     public string ToExceptionMessage(int httpStatus)
     {
         StringBuilder messageBuilder = new();
-        messageBuilder.AppendLine($"HTTP {httpStatus} ({Type}: {Code})");
+        messageBuilder.Append("HTTP ").Append(httpStatus).Append(" (").Append(Type).Append(": ").Append(Code).AppendLine(")");
```
KrzysztofCwalina, line 38: what's the reason for this change?

stephentoub: On netstandard2.0 it's allocating a new string only to then append that string.

On .NET Core it'll use the more efficient overload to avoid that, but an analyzer points out it should ideally be passing in a culture, and I can't do that without an ifdef.

If you're ok with an ifdef, I could do that and accept the lesser perf on netstandard (after all, this appears to only be exceptional). Or I could undo and if/when we enable that analyzer permanently we could suppress it here.

trrwilson (approved):

PR #48, opened 2024-06-12
Added sample showing how to do cancellations

KrzysztofCwalina:


examples/Chat/Example01_SimpleChat_Cancellations.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        CancellationTokenSource ct = new CancellationTokenSource();
+        RequestOptions options = new() { CancellationToken = ct.Token };
+
+        // The following code will be simplified in the future.
+        var wireFormat = new ModelReaderWriterOptions("W");
+        ChatMessage message = ChatMessage.CreateUserMessage("Say 'this is a test.'");
+        BinaryData json = ModelReaderWriter.Write(message, wireFormat);
+
+        ClientResult result = client.CompleteChat(BinaryContent.Create(json), options);
```
joseharriaga: This `BinaryContent` contains the JSON for one message:

```json
{
    "role": "user",
    "content": "Say \u0027this is a test.\u0027"
}
```

This will result in a 400 error.

The appropriate format should be:

```json
{
    "model": "gpt-4o",
    "messages": [
        {
            "role": "user",
            "content": "Say \u0027this is a test.\u0027"
        }
    ]
}
```

KrzysztofCwalina: Yeah, I forgot about this. Now I get the context of your question about models not corresponding 1-1 to the payloads returned from protocol methods. I also don't think that we can get to 100% of 1-1 mappings, unless we introduce the concept of "protocol models", which i am not sure is worth it.

joseharriaga (approved):

PR #53, opened 2024-06-12
Added CancellationToken parameters to the most complete convenience overloads. 

### LINKED ISSUE #35: CancellationToken support
Simple, really:

We should be able to pass `CancellationToken`s to any methods that make API calls.

KrzysztofCwalina:
fixes #35

### DIFF
--- a/src/Custom/Assistants/AssistantClient.cs
+++ b/src/Custom/Assistants/AssistantClient.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.Threading;
 using System.Threading.Tasks;
 using static OpenAI.InternalListHelpers;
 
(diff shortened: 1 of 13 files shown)


src/Custom/Common/RequestOptionsExtensions.cs
```diff
@@ -0,0 +1,22 @@
+using System.ClientModel.Primitives;
+using System.Threading;
+
+internal static class RequestOptionsExtensions
```
trrwilson: Minor: should this be `CancellationTokenExtensions` given that's the `this` type being extended?

src/Custom/Assistants/AssistantClient.cs
```diff
@@ -915,7 +980,4 @@ private static ClientResult<T> CreateResultFromProtocol<T>(ClientResult protocol
         T deserializedResultValue = responseDeserializer.Invoke(pipelineResponse);
         return ClientResult.FromValue(deserializedResultValue, pipelineResponse);
     }
-
-    private static RequestOptions StreamRequestOptions => _streamRequestOptions ??= new() { BufferResponse = false };
-    private static RequestOptions _streamRequestOptions;
 }
```
trrwilson, line 983: If Ctrl+C/Ctrl+V aren't worn out on the keyboard yet, could we also mirror this in the companion `AssistantClient.Convenience.cs` file's overloads? I know we have a pending discussion item on whether we should even have those, but in the interim it'd be great to retain parity if at all possible.

src/Custom/Chat/ChatClient.cs
```diff
@@ -126,16 +130,21 @@ public virtual ClientResult<ChatCompletion> CompleteChat(params ChatMessage[] me
...
+    public virtual AsyncResultCollection<StreamingChatCompletionUpdate> CompleteChatStreamingAsync(IEnumerable<ChatMessage> messages, ChatCompletionOptions options = null, CancellationToken cancellationToken = default)
     {
         Argument.AssertNotNull(messages, nameof(messages));
 
         options ??= new();
         CreateChatCompletionOptions(messages, ref options, stream: true);
 
         using BinaryContent content = options.ToBinaryContent();
-        RequestOptions requestOptions = new() { BufferResponse = false };
+        RequestOptions requestOptions = new() {
```
trrwilson: With the extension helpers, can this just be `cancellationToken.ToRequestOptions(true)` ?

tests/Assistants/VectorStoreTests.cs
```diff
@@ -73,7 +73,7 @@ public void CanCreateGetAndDeleteVectorStores()
         Assert.That(deleted, Is.True);
         _vectorStoresToDelete.RemoveAt(_vectorStoresToDelete.Count - 1);
 
-        vectorStore = client.CreateVectorStore(new()
+        vectorStore = client.CreateVectorStore(new VectorStoreCreationOptions()
```
trrwilson: Is there any reasonable approach for us to avoid the overload collisions that make implicit `new()` non-viable? It's a small thing but it's unfortunate if we miss an opportunity to make instantiation of these trivial options containers simpler. Would it e.g. be problematic for us to have three declarations:
- Foo(FooOptions, CancellationToken)
- Foo(FooOptions) => Foo(FooOptions, default(CancellationToken))
- Foo(BinaryContent, RequestOptions)

I believe that'd then make `new()` non-ambiguous since there'd be only one single-parameter option.

src/Custom/Assistants/AssistantClient.cs
KrzysztofCwalina: I don't think we want to do that. These are just simple call throughs, and if I add more and more overloads, the possibility of ambiguities at the call sites just goes up. That's why the title of the PR says "most complete convenience overloads".

src/Custom/Chat/ChatClient.cs
KrzysztofCwalina: Yeah, will do. I added the streaming helper after I changed the ChatClient.

tests/Assistants/VectorStoreTests.cs
KrzysztofCwalina: Yeah, we could do this. But ultimately, it means we should not use optional parameters in public APIs. Even that does not solve the evolution problem completely, but it does mitigate it.

trrwilson (approved):

src/Custom/Assistants/AssistantClient.cs
joseharriaga: I noticed that, for example, in the `AudioClient`, the methods that take a `Stream` do have a `CancellationToken`, but the overloads that take a `string` file path do not. Also noticed the same with other similar overloads in the `FileClient` and the `ImageClient`. I think this is a little unexpected because it forces users that need a `CancellationToken` into a specific paradigm.

If the main concern of adding `CancellationToken` is that things are more likely to break due to ambiguities in the future, I wonder if there are mitigation strategies that we could consider. Off the top of my head, what if any client method only ever had two optional parameters: The `CancellationToken` parameter and an options class parameter (and then the options class holds all the other optional parameters, present and future). I recognize that preemptively adding an options class for methods that only have one or two optional parameters today is perhaps not ideal, but personally I think that the benefit of knowing that this would never break due to ambiguities in the future is a good tradeoff.

joseharriaga (approved):

src/Custom/Assistants/AssistantClient.cs
KrzysztofCwalina: Yeah, let's look at the APIs wholistically and see if we want to add more of these overloads.

PR #70, opened 2024-06-18
Fixes to Assistant, VectorStore client methods

### LINKED ISSUE #61: GetRunStepsAsync does not yet take a CancellationToken parameter
I just got the latest NuGet package and I love all the new CancellationToken parameters. GetRunStepsAsync is one of the calls I make where it is missing.

### LINKED ISSUE #62: GetAssistantAsync does not return a useful value
AssistantClient.GetAssistantAsync only returns a `ClientResult`, not a `ClientResult<Assistant>`. Also, it should take a CancellationToken optional parameter, rather than just a RequestOptions optional parameter.

trrwilson:
- Fixes #62 -- the remaining convenience methods in `AssistantClient` and `VectorStoreClient` have optional `CancellationToken`s added as method parameters, including the typed instance convenience overloads
- Closes #61 -- adds missing `[EditorBrowsable(EditorBrowsableState.Never)]` attributes to `AssistantClient` protocol methods, improving strongly typed convenience method discoverability
- Addresses a yet-unreported problem with the new `max_num_results` property for `file_search` not being serialized/deserialized


api/api.md
```diff
@@ -47,142 +47,148 @@ namespace OpenAI.Assistants {
...
+        public virtual Task<ClientResult<ThreadMessage>> CreateMessageAsync(AssistantThread thread, IEnumerable<MessageContent> content, MessageCreationOptions options = null, CancellationToken cancellationToken = default);
+        public virtual Task<ClientResult<ThreadMessage>> CreateMessageAsync(string threadId, IEnumerable<MessageContent> content, MessageCreationOptions options = null, CancellationToken cancellationToken = default);
         public virtual Task<ClientResult> CreateMessageAsync(string threadId, BinaryContent content, RequestOptions options = null);
-        public virtual ClientResult<ThreadRun> CreateRun(AssistantThread thread, Assistant assistant, RunCreationOptions options = null);
-        public virtual ClientResult<ThreadRun> CreateRun(string threadId, string assistantId, RunCreationOptions options = null);
+        public virtual ClientResult<ThreadRun> CreateRun(AssistantThread thread, Assistant assistant, RunCreationOptions options = null, CancellationToken cancellationToken = default);
```
KrzysztofCwalina: This is not exactly what I did in my PR. In my PR, I added cancellation tokens only to the most general/flexible overload. In case of this method, it would mean adding the CT only to the overload that takes string threadId and string assistantId. If the caller have AssistantThread and Assistant instances, they would simply call CreateRun(thread.Id, assistant.Id). In other words, I did not add CTs to helper overloads that simply delegate to other more general/flexible overloads.

api/api.md
```diff
@@ -879,6 +890,7 @@ namespace OpenAI.Chat {
         public ChatFunctionCall FunctionCall { get; init; }
         public string ParticipantName { get; init; }
         public IList<ChatToolCall> ToolCalls { get; }
+        protected override void WriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
```
joseharriaga, line 888: This file needs to be regenerated before merging.

joseharriaga (approved):

api/api.md
joseharriaga: Based on our conversation offline, we have decided to remove the overloads that take complex objects (such as `ThreadRun run`) and keep only the ones that take plain string IDs (such as `string runId`) in favor of exploring the option of making these models into subclients. For example, rather than having `AssistantClient.CancelRun(ThreadRun run)`, we could instead have `ThreadRun.Cancel()`.

With these overloads gone, Kryzsztof's concern here will be addressed too.

trrwilson: Reverted all changes to the convenience file (which will be removed in a later PR).

api/api.md
trrwilson: Regenerated, hand-copying AssistantClient and VectorStoreClient. Note that this still looks a little "dirty" because it's not fresh in main -- this WriteCore [does exist in code](https://github.com/openai/openai-dotnet/blob/59c8433084d9bdb4bac62a48ab39942be0c54060/src/Custom/Assistants/ToolDefinition.Serialization.cs#L18) despite [not currently being in the .md](https://github.com/openai/openai-dotnet/blob/59c8433084d9bdb4bac62a48ab39942be0c54060/api/api.md?plain=1#L638).

joseharriaga (approved):

PR #75, opened 2024-06-19
Address non-ASCII filename support for files, audio, etc.

### LINKED ISSUE #72: When I upload a file with a Chinese txt format or other format in the file name, it prompts that the uploaded file does not support it
When I upload a file with a Chinese txt format or other format in the file name, it prompts that the uploaded file does not support it
企业微信截图_17187767209882
企业微信截图_17187768056251

trrwilson:
Fixes #72.

The library's `multipart/form-data` implementation uses custom logic to bypass the use of the `filename*` `Content-Disposition` parameter, which is [RFC 6266](https://www.rfc-editor.org/rfc/rfc6266#section-4.3)'s method of encoding non-ASCII filenames in `multipart/form-data` requests. This bypass was made in deference to an interpretation of [RFC 7578](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) wherein the use of `filename*` is disallowed.

This PR removes the restriction by directly using `System.Net.Http.MultipartFormDataContent.Add()`'s `filename*` behavior.

Although pending discussion with the Base Common Library team is still needed to clarify best interoperability practices, this is a better current state because:

- It aligns with the existing language type in `System.Net.Http`
- The interpretation of the competing RFC is ambiguous; it appears it may not be making a statement about the `filename*` parameter itself, but rather on the use of the *encoding scheme* of the newer `filename*` in the older `filename`
- Most importantly, it fixes the readily observable issue against the OpenAI V1 endpoint -- and still leaves the older `filename` intact for any unanticipated backwards compatibility issues with API gateways or other intermediaries

### DIFF
--- a/src/Utility/MultipartFormDataBinaryContent.cs
+++ b/src/Utility/MultipartFormDataBinaryContent.cs
@@ -75,14 +75,14 @@ internal class MultipartFormDataBinaryContent : BinaryContent
         Add(new ByteArrayContent(content.ToArray()), name, fileName);
     }
 
-    private void Add(HttpContent content, string name, string filename)
+    private void Add(HttpContent content, string name, string fileName)
     {
         Argument.AssertNotNull(content, nameof(content));
         Argument.AssertNotNull(name, nameof(name));
 
-        if (filename is not null)
+        if (fileName is not null)
         {
-            _multipartContent.Add(content, name, filename);
+            _multipartContent.Add(content, name, fileName);
         }
         else
         {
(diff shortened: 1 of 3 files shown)


src/Utility/MultipartFormDataBinaryContent.cs
```diff
@@ -76,48 +75,19 @@ public void Add(BinaryData content, string name, string fileName = default)
         Add(new ByteArrayContent(content.ToArray()), name, fileName);
     }
 
-    private void Add(HttpContent content, string name, string filename, string contentType)
+    private void Add(HttpContent content, string name, string filename)
```
glecaros: nit: we seem to use fileName and filename, maybe we should update to pick one.

glecaros (approved):

joseharriaga (approved):

PR #76, opened 2024-06-20
Allow ChatCompletionOption properties to be set/changed after instantiation

RogerBarreto:
The same way we can change Tools and other Properties in the Options class for long running Chats we normally need to change the `ToolChoice`, and as well as other properties of the chat iteration.

Currently that is only possible allocating (deep cloning) a new Option instance with those few options with the desired value.


trrwilson:
Thanks for this, @RogerBarreto -- as mentioned on the pinged issue, we agree with this need and we're going to make the change across all top-level request option types like this; there's really no good reason for request option properties to be treated like client option properties.

We have to make a small change to the generated layer to catch everything (hence why it's not just done yet) but it won't be a long wait.

PR #80, opened 2024-06-20
Added support for OrganizationId and ProjectId

### LINKED ISSUE #68: Missing support for OrganizationID and ProjectID
When I generate an API key, it requires me to use an OrganizationID and ProjectID in conjunction with the API key. There appears to be no mechanism to specify these when creating a client.

https://platform.openai.com/docs/api-reference/authentication

KrzysztofCwalina:
fixes #68 

@trrwilson, we will need to coordiante this with AOAI package update. As we talked, AOAI library will now need its own client options.

### DIFF
--- a/src/Custom/OpenAIClient.cs
+++ b/src/Custom/OpenAIClient.cs
@@ -6,17 +6,13 @@ using OpenAI.Embeddings;
 using OpenAI.Files;
 using OpenAI.FineTuning;
 using OpenAI.Images;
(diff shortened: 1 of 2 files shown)


trrwilson (approved):

trrwilson:
I believe we can safely merge this even before the Azure breaking change -- the properties will be present but take no effect, as the Azure library manages its own policy logic.

PR #92, opened 2024-06-25
Update paging APIs to use updated SCM types

annelo-msft:
Currently prototype work-in-progress.

Uses SCM paging types updated in https://github.com/Azure/azure-sdk-for-net/pull/44729


README.md
```diff
@@ -132,10 +132,10 @@ foreach (StreamingChatCompletionUpdate update in updates)
 }
 ```
 
-Alternatively, you can do this asynchronously by calling the `CompleteChatStreamingAsync` method to get an `AsyncResultCollection<StreamingChatCompletionUpdate>` and enumerate it using `await foreach`:
+Alternatively, you can do this asynchronously by calling the `CompleteChatStreamingAsync` method to get an `AsyncResultValueCollection<StreamingChatCompletionUpdate>` and enumerate it using `await foreach`:
```
KrzysztofCwalina: this collection type name is long. I forgot we also need to prefix async. I think it's shipable, but maybe somebody can come up with a name that is only 3 words long.

annelo-msft: Some other options I considered:
- ClientValueCollection
- CollectionResult  // Problem with this is that it's not a single result in all cases and if not the result can change
- Remove the type in favor of returning IAsyncEnumerable\<ClientResult\<T\>\> // cons here are that the same response would be associated with a T that is contained with other values in the same response
- ValueCollection // I expect this to be too general

annelo-msft: Actually, I am starting to favor CollectionResult.  The return value from GetRawResponse might change, but the fact that a result is returned from a GetValues method is true and that Collection is a reasonable adjective to describe the result.  I worry that any adjective modifying Collection would need to have to words to be sufficiently non-general, and then we have failed.  CollectionResult has the nice property that then the LRO base type can be OperationResult and there is nice symmetry there.

annelo-msft: Snapped to CollectionResult per offline discussion -- also changed ClientPage\<T\> to PageResult\<T\> to follow new precedent that types inheriting from ClientResult have a Result suffix.

annelo-msft:
Closing in favor of https://github.com/openai/openai-dotnet/pull/105

PR #98, opened 2024-06-29
Remove net6.0 multitarget for improved compatibility

trrwilson:
As described at:
https://github.com/Azure/azure-sdk-for-net/issues/44709

These problems are specific to using the OpenAI library via an intermediate class library targeting `netstandard2.0`; in that case, the multi-target is causing the executing project to pick up the uplevel framework target that's incompatible with downlevel intermediate library.

Although there may be other approaches to mitigate, it seems reasonable to not expressly multi-target net6.0 if netstandard2.0 is sufficient.

### DIFF
--- a/src/OpenAI.csproj
+++ b/src/OpenAI.csproj
@@ -7,7 +7,7 @@
     <VersionPrefix>2.0.0</VersionPrefix>
     <VersionSuffix>beta.7</VersionSuffix>
 
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <LangVersion>latest</LangVersion>
 
     <!-- Sign the assembly with the specified key file. -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
 # Release History
 
-## 2.0.0-beta.7 (Unreleased)
+## 2.0.0-beta.8 (Unreleased)
 
 ## Bugs Fixed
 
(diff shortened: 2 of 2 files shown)


src/OpenAI.csproj
```diff
@@ -7,7 +7,7 @@
     <VersionPrefix>2.0.0</VersionPrefix>
     <VersionSuffix>beta.7</VersionSuffix>
 
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
rogerbarreto, line 10: Should this target 8.0 instead?

annelo-msft, line 10: This may cause issues with sync-over-async in the SCM pipeline, which relies on newer net6.0 sync APIs being available for HttpClient APIs to avoid sync-over-async for the majority case.  Removing this target would require sync-over-async for all sync HttpClient calls.

KrzysztofCwalina: I don't think we will start doing sync over async, unless this OpenAI dll calls the sync HTTP APIs directly. The System.ClientModel implementation (6+ or NS2) is picked based on the runtime of the program, not based on framework target of OpenAI. @ericstj, could you confirm or deny?

KrzysztofCwalina: Having said that, I don't think we should be removing NETx target. I think we should always cross target to the latest LTS target.

joy-sarkar-bain:
Hello all, 

Is there any likelihood of this PR being merged? The targeting issue seems to be a blocker for me; if there is some resolution (either via this PR or some other method?), it might unlock a lot of use cases. 

Thanks in advance
JS

joseharriaga:
We removed the initters, which means that the original issue that prompted this change is no longer applicable. Therefore, I'm closing this PR. 🙂

PR #105, opened 2024-07-01
Update paging APIs to use updated SCM types

annelo-msft:
In [System.ClientModel 1.1.0-beta.5](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/System.ClientModel/CHANGELOG.md#110-beta5-2024-07-11), `PageableCollection<T>` was changed to `PageCollection<T>` to better align convenience and protocol methods.

This PR upgrades the OpenAI library to SCM 1.1.0-beta.5, and manually implements internal types that will be generated from TypeSpec API specification in the future.  Specifically, it:

- Updates return types that have been renamed for streaming service methods
- For service methods corresponding to paginated-collection endpoints
  - Returns `PageCollection<T>` or `AsyncPageCollection<T>` from convenience methods
  - Takes `XxCollectionOptions` types that describe the paginated collection to be returned
  - Adds "rehydration" methods that return a collection that starts where another collection type left off via a persisted rehydration token
  - Returns `IEnumerable<ClientResult>` or `IAsyncEnumerable<ClientResult>` from protocol methods, representing a collection of service responses, where each corresponds to a page of results returned from the paginated endpoint
  - Implements internal page enumerator and page token types for each paginated-endpoint operation

Types that are intended to be output by the generator have been added in a temporary folder called `To.Be.Generated`.


README.md
```diff
@@ -528,7 +528,7 @@ Finally, you can use the `AssistantClient`'s `GetMessages` method to retrieve th
 For illustrative purposes, you could print the messages to the console and also save any images produced by the assistant to local storage:
 
 ```csharp
-PageableCollection<ThreadMessage> messages = assistantClient.GetMessages(threadRun.ThreadId, ListOrder.OldestFirst);
+PageCollection<ThreadMessage> messages = assistantClient.GetMessages(threadRun.ThreadId, ListOrder.OldestFirst);
```
trrwilson, line 531: If I'm looking at ApiView correctly, `ListOrder` is now replaced by operation/collection-specific options -- I don't see the old `ListOrder` type used here or anywhere else. Should we update the references here and retire the old class entirely?

annelo-msft: Sure - happy to remove that type as part of this PR.

src/Custom/Assistants/MessageCollectionOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Creates a new instance of <see cref="MessageCollectionOptions"/>.
+    /// </summary>
+    public MessageCollectionOptions() { }
+
+    /// <summary>
+    /// The <c>order</c> that results should appear in the list according to
+    /// their <c>created_at</c> timestamp.
+    /// </summary>
+    public ListOrder? Order { get; init; }
```
trrwilson, line 17: Per design guidance on loosening up accessors, should make all of these `set` and not `init`?

examples/Assistants/Example03_ListAssistantsWithPaginationAsync.cs
```diff
@@ -17,7 +18,8 @@ public async Task Example03_ListAssistantsWithPaginationAsync()
 
         int count = 0;
 
-        AsyncPageableCollection<Assistant> assistants = client.GetAssistantsAsync();
+        AsyncPageCollection<Assistant> assitantPages = client.GetAssistantsAsync();
```
trrwilson: nit: stubborn `s` key (`assitantPages` -> `assistantPages`) -- you must have the same keyboard I do 😆

examples/Assistants/Example04_AllTheTools.cs
```diff
@@ -171,8 +172,8 @@ PageableCollection<ThreadMessage> messages
             #endregion
 
             #region List run steps for details about tool calls
-            PageableCollection<RunStep> runSteps = client.GetRunSteps(run, resultOrder: ListOrder.OldestFirst);
-            foreach (RunStep step in runSteps)
+            PageCollection<RunStep> runSteps = client.GetRunSteps(run, new() { Order = ListOrder.OldestFirst });
```
trrwilson: Ignorable/opinion-only: even though it's more verbose, I like either using the explicit constructor named method parameter in example code like this; until you've learned the patterns, it's to infer what's being `new()`'d.

examples/Assistants/Example04_AllTheTools.cs
```diff
@@ -171,8 +172,8 @@ PageableCollection<ThreadMessage> messages
...
+            foreach (RunStep step in runSteps.GetAllValues())
```
trrwilson, line 180: Is it not possible to iterate directly over the collection anymore? Is the idea that we don't want to unwittingly hide the paging?

annelo-msft: That is correct.  I wrote down some of the design considerations that drove this decision [here](https://github.com/Azure/azure-sdk-for-net/pull/44825#issue-2384746666).  Essentially, though, this one is driven by the need for convenience method return types to inherit from protocol method return types in SCM-based clients, and to have conceptual alignment across those return types.

examples/Assistants/Example04_AllTheTools.cs
annelo-msft: Yep, happy to. I was optimizing for characters, but if we feel this isn't a familiar pattern yet, I'm happy to go with what we think will be more helpful to users.

examples/Assistants/Example03_ListAssistantsWithPaginationAsync.cs
annelo-msft: TY!  It's the moving from left to right on the left hand motion ...

src/Custom/Assistants/MessageCollectionOptions.cs
annelo-msft: I'm happy to move to whatever convention we decide on for the library on this.  I don't know if this one would have the same effect on the ASP.NET initialization sequence that using `init` for client options does, but I also don't feel strongly one way or another.  Have we finalized the decision around the convention for non-client options types?

README.md
annelo-msft: Per offline discussion, I was thinking this was a general list options and not something we want to keep to have a way to specify item order within a collection.  We will keep `ListOrder`!

src/To.Be.Generated/AssistantsPageEnumerator.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.ClientModel;
+using System.ClientModel.Primitives;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+#nullable enable
+
+namespace OpenAI.Assistants;
+
+internal partial class AssistantsPageEnumerator : PageEnumerator<Assistant>
```
ShivangiReja: Is the plan for `To.be.Generated` types to be added to the TypeSpec specification, or do these already exist in the TypeSpec and this is just a generator gap?"

trrwilson (approved):

src/To.Be.Generated/AssistantsPageEnumerator.cs
annelo-msft: TypeSpec has a proposed design for pagination, but it is still under development.  I've been working with the team to ensure that the SCM types can be used for anything the TypeSpec can represent, and last I checked, they are looking into unifying the 3p pagination design with the 3p LRO design, at least at lower levels, around the observation that the "get next page" method for pagination is similar conceptually to the "get status update" method for LROs.

To answer your question - to my knowledge, the generator is currently unable to implement generation of these types due to the TypeSpec pieces not being available yet.

src/Custom/Assistants/MessageCollectionOptions.cs
annelo-msft: I'll merge this as-is since other options properties haven't been updated yet -- just so we don't end up with two approaches in a single library version.  Should be easy to blanket-address in a single project-wide search-replace when we close on that convention.

src/Utility/PageEnumerator.cs
```diff
@@ -0,0 +1,60 @@
...
+    PageResult<T> IAsyncEnumerator<PageResult<T>>.Current
+    {
+        get
+        {
+            if (Current is null)
+            {
+                return default!;
+            }
+
+            return GetPageFromResult(Current);
```
annelo-msft, line 57: If someone calls `Current` multiple times, this will deserialize again instead of using a cached value.  Tracking with https://github.com/openai/openai-dotnet/issues/136

src/Custom/Assistants/AssistantsPageEnumerator.cs
```diff
@@ -0,0 +1,133 @@
...
+
+    public override ClientResult GetFirst()
+        => GetAssistants(_limit, _order, _after, _before, _options);
+
+    public override async Task<ClientResult> GetNextAsync(ClientResult result)
+    {
+        PipelineResponse response = result.GetRawResponse();
+
+        using JsonDocument doc = JsonDocument.Parse(response.Content);
+        _after = doc.RootElement.GetProperty("last_id"u8).GetString()!;
```
annelo-msft, line 51: This was implemented this way to illustrate what an implementation might look like for protocol-only, but once we have added convenience APIs and are deserializing values to return to end-users, we don't need to parse the JsonDocument multiple times.  Tracking with https://github.com/openai/openai-dotnet/issues/136

src/Custom/Assistants/MessageCollectionOptions.cs
joseharriaga: Sounds good. FYI: I synced with @ShivangiReja earlier today and she will be making the fix to change all `init` to `set`.

joseharriaga (approved):

PR #107, opened 2024-07-02
Add instrumentation for simple convenience chat calls

lmolkova:
This PR builds foundation for OpenAI SDK tracing and metrics instrumentation (using Otel-compatible .NET primitives).

It's limited to convenience `ChatClient` methods without streaming. The PR implements instrumentation according to [OpenTelemetry GenAI semantic conventions](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/gen-ai). 

The intention is to add instrumentation to other methods and client types and evolve it along with OTel GenAI semantic conventions. 

TODO (in this PR):
- [x] add samples/docs
- [x] add experimental feature-flag required to enable instrumentation - we don't know when OTel semantic conventions will be stable and expect breaking changes.

TODO (in next PRs):
- [ ] add instrumentation to streaming calls and protocol methods
- [ ] track prompts and completions in events
- ...


src/Custom/Common/Instrumentation/InstrumentationScope.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        return null;
+    }
+
+    private void StartChat(ChatCompletionOptions options)
+    {
+        _duration = Stopwatch.StartNew();
+        _commonTags = new TagList
+        {
+            { Constants.GenAiSystemKey, Constants.GenAiSystemValue },
```
cijothomas: is this fully static? Perhaps a better fit for Resource?

lmolkova: resource attributes are per-all-tracers + only otel SDK can set them up. Instrumentation scope attributes are not available on activity sources and in many other languages.

So it'd be nice to optimize it in the future, but it'd need changes in otel beyond gen-ai or even semconv.

README.md
```diff
@@ -726,3 +730,57 @@ By default, the client classes will automatically retry the following errors up
...
+
+### How to enable
+
+The instrumentation is **experimental** - names of activity sources and meters, volume and semantics of the telemetry items may change.
+
+To enable the instrumentation:
+
+1. Set instrumentation feature-flag using one of the following options:
+
+   - set the `AZURE_EXPERIMENTAL_ENABLE_ACTIVITY_SOURCE` environment variable to `"true"`
```
cijothomas: Curious, why its named "AZURE_..." ? Is this library connected to Azure SDKs in general?

README.md
```diff
@@ -726,3 +730,57 @@ By default, the client classes will automatically retry the following errors up
...
+            ...
+            .AddOtlpExporter();
+       });
+   ```
+
+Consider enabling [HTTP client instrumentation](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.Http) to see all HTTP client
+calls made by your application including those done by the OpenAI SDK.
+Check out [OpenTelemetry documentation](https://opentelemetry.io/docs/languages/net/getting-started/) for more details.
+
+### Available sources and meters
```
cijothomas: it might not be obvious what are "sources" and "meters"..

src/Custom/Common/Instrumentation/InstrumentationScope.cs
```diff
@@ -0,0 +1,218 @@
...
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+
+namespace OpenAI.Custom.Common.Instrumentation;
+
+internal class InstrumentationScope : IDisposable
+{
+    private static readonly ActivitySource s_chatSource = new ActivitySource("OpenAI.ChatClient");
+    private static readonly Meter s_chatMeter = new Meter("OpenAI.ChatClient");
+    private static readonly Histogram<double> s_duration = s_chatMeter.CreateHistogram<double>(Constants.GenAiClientOperationDurationMetricName, "s", "Measures GenAI operation duration.");
```
cijothomas: please add a TODO to provide the explicit buckets, when DS 9.0 can be used.

src/Custom/Common/Instrumentation/InstrumentationScope.cs
```diff
@@ -0,0 +1,218 @@
...
+
+    private void RecordMetrics(string responseModel, string errorType, int? inputTokensUsage, int? outputTokensUsage)
+    {
+        TagList tags = ResponseTagsWithError(responseModel, errorType);
+        s_duration.Record(_duration.Elapsed.TotalSeconds, tags);
+
+        if (inputTokensUsage != null)
+        {
+            // tags is a struct, let's copy them
+            TagList inputUsageTags = tags;
```
cijothomas: Token usage metrics has no attribute for error.type, but tags here will have error.type.

tests/Instrumentation/TestActivityListener.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace OpenAI.Tests.Instrumentation;
+
+internal class TestActivityListener : IDisposable
+{
+    private readonly ActivityListener _listener;
+    private readonly ConcurrentQueue<Activity> stoppedActivities = new ConcurrentQueue<Activity>();
+
+    public TestActivityListener(string sourceName)
+    {
+        _listener = new ActivityListener()
```
cijothomas: if the tests can take OTel dependency sdk, then consider using InMemoryExporter for these tests.

README.md
trrwilson: Agreed -- this bears a bit more of a concept-level introduction. We've already heard feedback that the top-level README is very big, though, so just adding more seems to exacerbate that; perhaps we can use this topic as a pilot for separating detailed topic information into a sub-page? E.g., as a pattern-match, check out how the Python SDK splits out streaming helpers: https://github.com/openai/openai-python?tab=readme-ov-file#streaming-helpers

I'd propose that we put this in a `/docs` folder rather than repo root, but we could then put a dedicated header section into the top-level README for instrumentation explaining what it is and why you'd use it, then a link to the `/docs/instrumentation.md` for follow-up.

src/Custom/Chat/ChatClient.cs
```diff
@@ -1,3 +1,4 @@
+using OpenAI.Custom.Common.Instrumentation;
```
joseharriaga: nit: Could you move the "Instrumentation" folder under "src/Utility". And the namespace can be `OpenAI.Instrumentation`.

README.md
```diff
@@ -26,6 +27,9 @@ It is generated from our [OpenAPI specification](https://github.com/openai/opena
 - [Advanced scenarios](#advanced-scenarios)
   - [Using protocol methods](#using-protocol-methods)
   - [Automatically retrying errors](#automatically-retrying-errors)
+- [Observability with OpenTelemetry](#observability-with-opentelemetry)
+  - [How to enable](#how-to-enable)
+  - [Available sources and meters](#available-sources-and-meters)
```
joseharriaga: nit: Can this be under the "Advanced scenarios" sub-header above?

README.md
joseharriaga: +1. There should be no references to Azure, which I believe is what you did in the actual implementation, so I believe is just about cleaning up these docs.

README.md
joseharriaga: I like the idea of putting this under a `/docs` folder and keeping the info in the README as short as possible in favor of linking to the other docs.

src/Custom/Common/Instrumentation/InstrumentationFactory.cs
```diff
@@ -0,0 +1,60 @@
+﻿using OpenAI.Chat;
+using System;
+
+namespace OpenAI.Custom.Common.Instrumentation;
+
+internal class InstrumentationFactory
```
annelo-msft: @lmolkova, do you have a sense of whether our plan for generating third-party clients will be to generate these factory and scope types internally in the client?  Or are we planning to provide types to support this in System.ClientModel, similar to the ones in Azure.Core today?

src/Custom/Chat/ChatClient.cs
```diff
@@ -62,6 +64,7 @@ protected internal ChatClient(ClientPipeline pipeline, string model, Uri endpoin
         _model = model;
         _pipeline = pipeline;
         _endpoint = endpoint;
+        _instrumentation = new InstrumentationFactory(model, endpoint);
```
annelo-msft: Will we need options in OpenAIClientOptions to configure distributed tracing, e.g. to turn it on or off and or other options like providing an allow-list to prevent secrets from being logged?

README.md
lmolkova: oops, great catch

README.md
lmolkova: done, PTAL

src/Custom/Common/Instrumentation/InstrumentationScope.cs
lmolkova: it should be impossible to have an error AND usage data (it comes from response body), but I refactored it a bit to be more obvious

tests/Instrumentation/TestActivityListener.cs
lmolkova: it's possible, are there any specific benefits? I don't mind writing validation against `ActivityListener` since it makes things really easy to debug and understand.

src/Custom/Chat/ChatClient.cs
lmolkova: done

src/Custom/Common/Instrumentation/InstrumentationFactory.cs
lmolkova: I think in the very long-term we will need a way to generate at least some of instrumentation code. But:
- it's very specific to the SDK and we'd need some support for it in typespec
- it's unlikely that we'll need anything in `System.ClientModel` - the Azure.Core stuff existed to bridge different generation of tracing API in .NET. Currently there is just one relevant set of APIs which client libraries can and should use directly.

I.e. I don't envision any changes to `System.ClientModel`, but maybe there will be a way to auto-gen instrumentation wrappers for public APIs.

src/Custom/Chat/ChatClient.cs
lmolkova: Yes, great question. 

I believe we'll need addition to base `ClientOptions` to enable/disable/configure tracing/metrics/logging for specific instances of clients. `OpenAIClientOptions` should be able to extend diagnostics options if necessary. 

In this PR tracing/metrics can be enabled globally for all OpenAI clients which is necessary regardless of per-instance configuration/overrides. There are no secrets/sensitive data to redact (yet).

So `ClientOptions` API addition would be purely additive. Do you want me to create issues to follow up?

annelo-msft: That would be great, and thank you!  @m-redding is working on public APIs for logging options in SCM now - it would be great to make sure we have a unified plan here if there are any overlaps in options for logging and distributed tracing.  @jsquire, @KrzysztofCwalina  FYI.

lmolkova: thanks! done - https://github.com/Azure/azure-sdk-for-net/issues/45267

src/Custom/Chat/ChatClient.cs
```diff
@@ -14,6 +15,7 @@ namespace OpenAI.Chat;
 public partial class ChatClient
 {
     private readonly string _model;
+    private readonly InstrumentationFactory _instrumentation;
```
KrzysztofCwalina: Would it be acceptable to call the type "OpenTelemetrySource" and the field "_telemetry"?

src/Utility/Instrumentation/Constants.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace OpenAI.Instrumentation;
+
+internal class Constants
```
KrzysztofCwalina: Could we rename to OpenTelemertyConstants"?

src/Utility/Instrumentation/Constants.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    // follows OpenTelemetry GenAI semantic conventions:
+    // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai
+
+    public const string ErrorTypeKey = "error.type";
```
KrzysztofCwalina: Why don't we have otel APIs that operate on UTF8 strings?

src/Utility/Instrumentation/InstrumentationFactory.cs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public InstrumentationScope StartChatScope(ChatCompletionOptions completionsOptions)
+    {
+        return IsInstrumentationEnabled
+            ? InstrumentationScope.StartChat(_model, ChatOperationName, _serverAddress, _serverPort, completionsOptions)
+            : null;
+    }
+
+    internal static class AppContextSwitchHelper
```
KrzysztofCwalina: Could we move this to a separate file?

src/Utility/Instrumentation/InstrumentationScope.cs
```diff
@@ -0,0 +1,219 @@
+﻿using OpenAI.Chat;
+using System;
+using System.Buffers;
+using System.ClientModel;
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+
+namespace OpenAI.Instrumentation;
+
+internal class InstrumentationScope : IDisposable
```
KrzysztofCwalina: Can we call it TelemetryScope

src/Utility/Instrumentation/InstrumentationScope.cs
```diff
@@ -0,0 +1,219 @@
+﻿using OpenAI.Chat;
+using System;
+using System.Buffers;
+using System.ClientModel;
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+
+namespace OpenAI.Instrumentation;
```
KrzysztofCwalina: Can we call the subnamespace "OpenTelemetry" or "Telemetry"?

src/Utility/Instrumentation/InstrumentationScope.cs
```diff
@@ -0,0 +1,219 @@
...
+        _requestModel = model;
+        _operationName = operationName;
+        _serverAddress = serverAddress;
+        _serverPort = serverPort;
+    }
+
+    public static InstrumentationScope StartChat(string model, string operationName,
+        string serverAddress, int serverPort, ChatCompletionOptions options)
+    {
+        if (s_chatSource.HasListeners() || s_tokens.Enabled || s_duration.Enabled)
```
KrzysztofCwalina: Should this check be refactored into a separate routine such that it can be reused in other methods.

src/Utility/Instrumentation/Constants.cs
lmolkova: OTel allows unicode attribute names and uses unicode strings on the wire format.

KrzysztofCwalina (approved):

tests/Chat/ChatSmokeTests.cs
```diff
@@ -470,4 +475,38 @@ public void SerializeChatMessageContentPartAsImageBytes(bool fromRawJson)
             Assert.That(additionalPropertyProperty.ValueKind, Is.EqualTo(JsonValueKind.True));
         }
     }
+
+    [Test]
+    [NonParallelizable]
+    public async Task HelloWorldChatWithTracingAndMetrics()
```
joseharriaga: Since this one is running against the live service, it doesn't fall in the "Smoke" category. Could you move this to ChatTests.cs file?

tests/Telemetry/ChatTelemetryTests.cs
```diff
@@ -0,0 +1,314 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using static OpenAI.Tests.Telemetry.TestMeterListener;
+using static OpenAI.Tests.Telemetry.TestActivityListener;
+
+namespace OpenAI.Tests.Telemetry;
+
+[TestFixture]
+[NonParallelizable]
+public class ChatTelemetryTests
```
joseharriaga, line 22: If I'm reading these correctly, these are basically mock tests, right? If so, could you add the attribute `[Category("Smoke")]` to this `ChatTelemetryTests class?

tests/Chat/ChatSmokeTests.cs
joseharriaga: This change will fix the CI pipeline.

lmolkova: done, thanks!

joseharriaga (approved):

PR #122, opened 2024-07-12
Add simple embedding example using protocol method

joseharriaga:
- Add an example illustrating how to use the `EmbeddingClient`'s `GenerateEmbeddings` protocol method
- Modify `Example06_SimpleChatProtocol` and `Example06_SimpleChatProtocolAsync` to use UTF-8 string literals for correctness

### DIFF
--- a/examples/Chat/Example06_SimpleChatProtocol.cs
+++ b/examples/Chat/Example06_SimpleChatProtocol.cs
@@ -13,7 +13,7 @@ public partial class ChatExamples
     {
         ChatClient client = new("gpt-4o", Environment.GetEnvironmentVariable("OPENAI_API_KEY"));
 
-        BinaryData input = BinaryData.FromString("""
+        BinaryData input = BinaryData.FromBytes("""
             {
                "model": "gpt-4o",
                "messages": [
@@ -23,7 +23,7 @@ public partial class ChatExamples
                    }
                ]
             }
-            """);
+            """u8.ToArray());
 
         using BinaryContent content = BinaryContent.Create(input);
         ClientResult result = client.CompleteChat(content);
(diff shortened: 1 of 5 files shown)


examples/Chat/Example06_SimpleChatProtocol.cs
```diff
@@ -23,17 +23,17 @@ public void Example06_SimpleChatProtocol()
...
         using BinaryContent content = BinaryContent.Create(input);
         ClientResult result = client.CompleteChat(content);
         BinaryData output = result.GetRawResponse().Content;
 
         using JsonDocument outputAsJson = JsonDocument.Parse(output.ToString());
         string message = outputAsJson.RootElement
-            .GetProperty("choices")[0]
-            .GetProperty("message")
-            .GetProperty("content")
+            .GetProperty("choices"u8)[0]
```
trrwilson, line 34: Aside, but do we know if/when we'll get a `ToDynamicFromJson` extension equivalent?

examples/Embeddings/Example04_SimpleEmbeddingProtocol.cs
```diff
@@ -0,0 +1,43 @@
...
+using System;
+using System.ClientModel;
+using System.Text.Json;
+
+namespace OpenAI.Examples;
+
+public partial class EmbeddingExamples
+{
+    [Test]
+    public void Example04_SimpleEmbeddingProtocol()
```
trrwilson, line 12: Should we add (maybe just as a link) this sample to the README?

joseharriaga: I do think we should improve discovery of our examples by linking them in the README. I would like to do it as a bigger sweep in a separate PR.

examples/Chat/Example06_SimpleChatProtocol.cs
joseharriaga: It has been agreed that we would like to have this functionality or something equivalent, but the corresponding work item is currently sitting in the backlog in the face of higher priorities like LROs, pagination, etc. so we don't have an ETA at the moment.

PR #124, opened 2024-07-13
Add serialization/deserialization example with chat completions

joseharriaga:
Added example illustrating how to serialize and deserialize a list of `ChatMessage`. We are also working to make this simpler in the future.

This change also refactors a few examples and tests.

### DIFF
--- a/examples/Chat/Example06_ChatSerialization.cs
+++ b/examples/Chat/Example06_ChatSerialization.cs
@@ -1,0 +1,102 @@
+﻿using NUnit.Framework;
+using OpenAI.Chat;
+using System;
+using System.ClientModel.Primitives;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Text.Json;
+
+namespace OpenAI.Examples;
+
+public partial class ChatExamples
+{
+    #region
+    public static List<ChatMessage> DeserializeMessages(BinaryData data)
+    {
+        List<ChatMessage> messages = [];
+        using JsonDocument messagesAsJson = JsonDocument.Parse(data.ToString());
+
+        foreach (JsonElement jsonElement in messagesAsJson.RootElement.EnumerateArray())
+        {
+            BinaryData jsonElementAsData = BinaryData.FromString(jsonElement.ToString());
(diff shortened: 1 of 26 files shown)


examples/Chat/Example06_ChatSerialization.cs
```diff
@@ -0,0 +1,104 @@
...
+{
+    #region
+    public static List<ChatMessage> DeserializeMessages(BinaryData data)
+    {
+        List<ChatMessage> messages = [];
+        using JsonDocument messagesAsJson = JsonDocument.Parse(data.ToString());
+
+        foreach (JsonElement jsonElement in messagesAsJson.RootElement.EnumerateArray())
+        {
+            BinaryData jsonElementAsData = BinaryData.FromString(jsonElement.ToString());
```
annelo-msft: Once [this ships](https://github.com/dotnet/runtime/issues/77666), I believe we should be able to improve perf here by not round-tripping UTF-8 through UTF-16 with strings ...

examples/Chat/Example06_ChatSerialization.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+        writer.WriteEndArray();
+        writer.Flush();
+
+        return BinaryData.FromBytes(stream.ToArray());
+    }
+    #endregion
+
+    [Test]
+    public void Example06_ChatSerialization()
```
trrwilson: Likewise to the Embeddings example, it seems like it'd be very valuable to link to this (with mention) from the README!

trrwilson (approved):

examples/Chat/Example06_ChatSerialization.cs
```diff
@@ -0,0 +1,104 @@
...
+
+namespace OpenAI.Examples;
+
+public partial class ChatExamples
+{
+    #region
+    public static List<ChatMessage> DeserializeMessages(BinaryData data)
+    {
+        List<ChatMessage> messages = [];
+        using JsonDocument messagesAsJson = JsonDocument.Parse(data.ToString());
```
KrzysztofCwalina: Why not parse the UTF8 payload, i.e. data.ToMemory()

PR #137, opened 2024-07-23
Add live tests to release, main and new live-test workflows

hallipr:


.github/workflows/release.yml
```diff
@@ -5,8 +5,10 @@ on:
     types: [published]
 
 jobs:
-  deploy:
+  build:
+    name: Build
     runs-on: ubuntu-latest
+    environment: Live Testing
```
weshaggard, line 18: Why not split this into build and test as well?

.github/workflows/live-test.yml
```diff
@@ -0,0 +1,39 @@
+name: Live Test
```
weshaggard, line 2: Is this one needed to be separate? Can we not just use the main.yml with a manual or scheduled trigger?

hallipr: Main runs build, unit test and deploy.   We will have already run these for any PR.   We're only lacking the ability to run live tests against the PR.

.github/workflows/release.yml
hallipr: Release will depend on live tests and live test should only run if unit tests have passed. As there's no parallel work, I didn't see the need to run this in a separate job. 

The same could be said for the Deploy job.   I don't see any reason for it to be its own job.

hallipr: ... except for retries.   I would be handy to be able to retry a deployment after a long live test run

.github/workflows/main.yml
```diff
@@ -1,7 +1,14 @@
-name: Build and Test
+# This worflow is triggered on push to main, pull request, or by manual invocation.
+# It builds and unit tests the codebase.
+name: Build and Unit Test
 
 on:
   workflow_dispatch:
+    inputs:
```
weshaggard: I don't think this is needed given they can trigger against a PR already and they can trigger against a branch as needed.

.github/workflows/release.yml
```diff
@@ -20,34 +31,69 @@ jobs:
...
+      contents: write
+    env:
+      NUGET_API_KEY: ${{ github.event_name == 'release' && secrets.NUGET_API_KEY || secrets.GITHUB_TOKEN }}
+      PUBLISH_TARGET: ${{ github.event_name == 'release' && 'https://api.nuget.org/v3/index.json' || 'github' }}
+    steps:
+      - uses: actions/download-artifact@v2
+
       # Append the nuget package to the github release that triggered this workflow
       - name: Upload release asset
+        if: github.event_name == 'release'
```
weshaggard, line 78: I think we always want to publish the artifact.

.github/workflows/release.yml
```diff
@@ -20,34 +31,69 @@ jobs:
...
+        uses: actions/upload-artifact@v2
+        if: ${{ !cancelled() }}
+        with:
+          name: build-artifacts
+          path: ${{ github.workspace }}/artifacts
+
+  deploy:
+    name: Publish Package
+    needs: build
+    if: github.event_name != 'pull_request'
```
weshaggard: is it even possible for this to be triggered from a PR?

.github/workflows/release.yml
```diff
@@ -20,34 +31,69 @@ jobs:
...
+        run: dotnet nuget add source
+          "https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json"
+          --name "github"
+          --username ${{ github.actor }}
+          --password ${{ secrets.GITHUB_TOKEN }}
+          --store-password-in-clear-text
 
-      - name: Publish
+      - name: Publish NuGet package
         run: dotnet nuget push
```
weshaggard, line 100: This needs a condition of ==release

.github/workflows/release.yml
```diff
@@ -20,34 +31,69 @@ jobs:
...
+    name: Publish Package
+    needs: build
+    if: github.event_name != 'pull_request'
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: write
+    env:
+      NUGET_API_KEY: ${{ github.event_name == 'release' && secrets.NUGET_API_KEY || secrets.GITHUB_TOKEN }}
+      PUBLISH_TARGET: ${{ github.event_name == 'release' && 'https://api.nuget.org/v3/index.json' || 'github' }}
```
weshaggard: Where is this used?

.github/workflows/release.yml
hallipr: This is attaching the artifacts to the Release resource.   Build artifacts are always upload as the last step in the build job

.github/workflows/release.yml
hallipr: condition removed

.github/workflows/release.yml
```diff
@@ -1,53 +1,101 @@
...
 
-      - name: Publish
+      - name: Publish NuGet package
         run: dotnet nuget push
-          ${{github.workspace}}/artifacts/packages/*.nupkg
-          --source https://api.nuget.org/v3/index.json
-          --api-key ${{ secrets.NUGET_API_KEY }}
-          --skip-duplicate
+            ${{github.workspace}}/build-artifacts/packages/*.nupkg
+            --source ${{ env.NUGET_API_KEY }}
```
weshaggard: Should this be `PUBLISH_TARGET`?

.github/workflows/release.yml
hallipr: We publish the scheduled runs to the local feed and release to nuget.org.   The conditionals in the job variables pick the feed and token to use

.github/workflows/release.yml
hallipr: in the Publish NuGet package step.   I copy pasted wrong before, but it's corrected now

.github/workflows/release.yml
hallipr: corrected

.github/workflows/main.yml
hallipr: removed

.github/workflows/release.yml
```diff
@@ -1,53 +1,104 @@
...
+          ${{ github.workspace }}/build-artifacts/packages/*.*nupkg
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Upload artifact
-        uses: actions/upload-artifact@v2
-        with:
-          name: build-artifacts
-          path: ${{github.workspace}}/artifacts
+      - name: NuGet autenticate
```
weshaggard:
```suggestion
      - name: NuGet authenticate
```

.github/workflows/live-test.yml
```diff
@@ -0,0 +1,45 @@
+# This workflow is triggered by the user and runs live tests on the codebase.
+name: Live Test
+
+on:
+  workflow_dispatch:
```
weshaggard, line 5: Didn't we decide to switch this trigger to be label based?

.github/workflows/live-test.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  test:
+    name: Live Test
+    runs-on: ubuntu-latest
+    environment: Live Testing
+    env:
+      version_suffix_args: ${{ format('--version-suffix="alpha.{0}"', github.run_number) }}
+    steps:
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v1
```
weshaggard:
```suggestion
        uses: actions/setup-dotnet@v3
```

weshaggard (approved):
One last question about the live-test trigger but otherwise looks good.

.github/workflows/live-test.yml
hallipr: I was just about to push that change :)

PR #139, opened 2024-07-24
[ModelFactory] Implemented Audio, Embedding, and Image factories

kinelski:
Adding Model Factories for Audio, Embedding, and Image models. These factories can be used to instantiate models for mocking.

At the moment there's no way to have the factories generated automatically since they are split across multiple namespaces.

I tried to keep them as close to generated code as possible. In case we manage to have the code generator generating them automatically for us, it would be ideal if we could keep changes to a minimum when migrating from manually-written to generated code.


src/Custom/Embeddings/Embedding.cs
```diff
@@ -83,6 +83,16 @@ internal Embedding(int index, BinaryData embeddingProperty, InternalEmbeddingObj
         Vector = ConvertToVectorOfFloats(embeddingProperty);
     }
 
+    // CUSTOM: Entirely custom constructor used by the Model Factory.
```
kinelski, line 86: Required to set `Vector` directly. The other constructor parses it from a `BinaryData` parameter.

src/Custom/Audio/OpenAIAudioModelFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Linq;
+
+namespace OpenAI.Audio;
+
+/// <summary> Model factory for models. </summary>
+public static partial class OpenAIAudioModelFactory
+{
+    /// <summary> Initializes a new instance of <see cref="OpenAI.Audio.AudioTranscription"/>. </summary>
+    /// <param name="duration"> The duration of the input audio. </param>
+    /// <returns> A new <see cref="OpenAI.Audio.AudioTranscription"/> instance for mocking. </returns>
```
kinelski: I'm only adding docstrings to a parameter if its corresponding property also has a docstring, from where I copied the text.

src/Custom/Audio/OpenAIAudioModelFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+            text,
+            words.ToList(),
+            segments.ToList(),
+            serializedAdditionalRawData: null);
+    }
+
+    /// <summary> Initializes a new instance of <see cref="OpenAI.Audio.AudioTranslation"/>. </summary>
+    /// <param name="duration"> The duration of the input audio. </param>
+    /// <returns> A new <see cref="OpenAI.Audio.AudioTranslation"/> instance for mocking. </returns>
+    public static AudioTranslation AudioTranslation(string language = null, TimeSpan? duration = null, string text = null, IEnumerable<TranscribedSegment> segments = null)
```
kinelski, line 29: Could be a nice idea to double check the order of parameters chosen by the generator (same order as defined in the typespec?). I've been following the parameter order defined by the constructor of each model. If we manage to get the generator to output model factories for us, we don't want it messing up parameter order.

src/Custom/Embeddings/OpenAIEmbeddingModelFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+            index,
+            vector.ToArray());
+    }
+
+    /// <summary> Initializes a new instance of <see cref="OpenAI.Embeddings.EmbeddingCollection"/>. </summary>
+    /// <param name="items"> The list of embeddings generated by the model. </param>
+    /// <param name="model"> The name of the model used to generate the embedding. </param>
+    /// <param name="usage"> The usage information for the request. </param>
+    /// <returns> A new <see cref="OpenAI.Embeddings.EmbeddingCollection"/> instance for mocking. </returns>
+    public static EmbeddingCollection EmbeddingCollection(IEnumerable<Embedding> items = null, string model = null, EmbeddingTokenUsage usage = null)
```
kinelski: I'm using the name `items` for elements in the collection. I'm not entirely sure whether this is a good idea or not. Any chance of these models getting a property named `Items` in the future?

src/Custom/Audio/OpenAIAudioModelFactory.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OpenAI.Audio;
+
+/// <summary> Model factory for models. </summary>
+public static partial class OpenAIAudioModelFactory
```
kinelski, line 8: Naming them with the pattern `OpenAI<client-name>ModelFactory`. Another alternative would be `OpenAI<namespace>ModelFactory`. For example, `OpenAIEmbeddingModelFactory` vs. `OpenAIEmbeddingsModelFactory`.

kinelski: Now that I'm thinking more about it, maybe going with the namespace makes more sense, since the factory is supposed to instantiate objects of a given namespace. `OpenAIModelsModelFactory` sounds more reasonable than `OpenAIModelModelFactory`.

src/Custom/Audio/OpenAIAudioModelFactory.cs
joseharriaga: Don't worry about the docstrings for the parameters. \<summary\> and \<returns\> are more than enough for now.

src/Custom/Audio/OpenAIAudioModelFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary> Initializes a new instance of <see cref="OpenAI.Audio.AudioTranscription"/>. </summary>
+    /// <param name="duration"> The duration of the input audio. </param>
+    /// <returns> A new <see cref="OpenAI.Audio.AudioTranscription"/> instance for mocking. </returns>
+    public static AudioTranscription AudioTranscription(string language = null, TimeSpan? duration = null, string text = null, IEnumerable<TranscribedWord> words = null, IEnumerable<TranscribedSegment> segments = null)
+    {
+        words ??= new List<TranscribedWord>();
+        segments ??= new List<TranscribedSegment>();
+
+        return new AudioTranscription(
+            task: default,
```
joseharriaga: What does default mean in this case? Does correctly set it to `InternalCreateTranscriptionResponseVerboseJsonTask.Transcribe` (a.k.a. "transcribe")?

src/Custom/Audio/OpenAIAudioModelFactory.cs
joseharriaga: I think I like the idea of going with the namespace (at least as much as someone can like calling it "ModelsModel" 🙃)

src/Custom/Audio/OpenAIAudioModelFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+
+    /// <summary> Initializes a new instance of <see cref="OpenAI.Audio.AudioTranslation"/>. </summary>
+    /// <param name="duration"> The duration of the input audio. </param>
+    /// <returns> A new <see cref="OpenAI.Audio.AudioTranslation"/> instance for mocking. </returns>
+    public static AudioTranslation AudioTranslation(string language = null, TimeSpan? duration = null, string text = null, IEnumerable<TranscribedSegment> segments = null)
+    {
+        segments ??= new List<TranscribedSegment>();
+
+        return new AudioTranslation(
+            task: default,
```
joseharriaga: Same here. Does this set the internal property to `InternalCreateTranslationResponseVerboseJsonTask.Translate` (a.k.a. "translate")?

src/Custom/Embeddings/OpenAIEmbeddingModelFactory.cs
joseharriaga: `items` is good. `EmbeddingCollection` inherits from `ReadOnlyCollection`, which has the `Items` property.

src/Custom/Images/OpenAIImageModelFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <param name="imageBytes">
+    /// The binary image data received from the response, provided when
+    /// <see cref="ImageGenerationOptions.ResponseFormat"/> is set to <see cref="GeneratedImageFormat.Bytes"/>.
+    /// </param>
+    /// <param name="imageUri">
+    /// A temporary internet location for an image, provided by default or when
+    /// <see cref="ImageGenerationOptions.ResponseFormat"/> is set to <see cref="GeneratedImageFormat.Uri"/>.
+    /// </param>
+    /// <returns> A new <see cref="OpenAI.Images.GeneratedImage"/> instance for mocking. </returns>
+    public static GeneratedImage GeneratedImage(BinaryData imageBytes = null, Uri imageUri = null, string revisedPrompt = null)
```
joseharriaga: `GeneratedImage` is a funny one. In practice, an instance of `GeneratedImage` will either have `imageBytes` or `imageUri` (in other words, one or the other will be `null`). I wonder if we should have two factory methods: One that takes the `imageBytes` and `revisedPrompt`, and then second one that takes `imageUri` and `revisedPrompt`. 🤔

src/Custom/Audio/OpenAIAudioModelFactory.cs
kinelski: I'll be consistent and remove all docstrings for parameters then, so we don't forget any parameter when we add them all.

src/Custom/Audio/OpenAIAudioModelFactory.cs
kinelski: It means the internal members will be set to their default values. In this case, the internal value will be `null`. I can set it to `InternalCreateTranscriptionResponseVerboseJsonTask.Transcribe` if you think we may eventually benefit from that. I didn't bother doing it because it was an internal property that didn't seem to have any particular use.

src/Custom/Images/OpenAIImageModelFactory.cs
kinelski: This is a tricky question. There are some maintainability benefits we may lose by having two methods due to the fact that they must be custom code, so we would need to remember to update them manually every time the `GeneratedImage` class changes (new properties, docstring changes, property renames, etc). We're only talking about a simple scenario (two possibilities), but ensuring a valid "state" of an object may get trickier for larger classes. We did [something similar](https://github.com/Azure/azure-sdk-for-net/blob/a247e40f374a099b2df042969db90a1fed2c8f41/sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentAnalysisModelFactory.cs#L236-L354) in the Form Recognizer SDK and ended up with 15 methods for a single model (luckily that one never changed after that).

The question we want to answer is: what's the actual benefit in ensuring the object has a valid state? How do our customers benefit from that? The only thing that comes to mind is we're preventing them from doing something like:
```cs
var image = OpenAIImagesModelFactory.GeneratedImage(imageBytes, imageUri);
```
However:
- How likely is this to happen? That would happen if they assumed both properties are returned without checking documentation nor trying to reproduce this behavior before. I can see that happening, frankly speaking, but I can't say how often.
- What's the negative impact on customers if they set the object wrongly? The two main scenarios that come to mind are: they will eventually realize this can't happen once they test against the service; or they'll spend a long time with a meaningless unit test. These are not catastrophic as far as I can tell.

In the end, I don't feel like it's the model factory's responsibility to ensure the generated objects are valid. That's why we don't enforce passing in all "required" parameters, only the ones they're interested in testing, setting all others to `default` or `null`.

Any thoughts?

src/Custom/Audio/OpenAIAudioModelFactory.cs
joseharriaga: Sounds good!

src/Custom/Audio/OpenAIAudioModelFactory.cs
joseharriaga: I think for "correctness" and because it's a simple change, let's set it to `Transcribe`. It's true that it is an internal property, but it is technically visible if the mock object is serialized.

src/Custom/Images/OpenAIImageModelFactory.cs
joseharriaga: That's a good point. I think we can leave it as is for now. If we receive feedback that this is confusing, we can always add the other overloads in the future and hide this one.

src/Custom/Audio/OpenAIAudioModelFactory.cs
kinelski: Updated both.

joseharriaga (approved):

PR #140, opened 2024-07-25
Updates to CHANGELOG and README following SCM collection type update

annelo-msft:
Follow-ups to https://github.com/openai/openai-dotnet/pull/105

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### Breaking Changes
 
+- Changed name of return types from methods returning streaming collections from `ResultCollection` to `CollectionResult` [#105](https://github.com/openai/openai-dotnet/pull/105).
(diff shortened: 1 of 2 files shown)


eliasandronicou:
Hello, is the issue solved? I am getting this error in every Assistant's example.
Error (active)	CS7069	Reference to type 'AsyncPageableCollection<>' claims it is defined in 'System.ClientModel', but it could not be found

joseharriaga:
Hello, @eliasandronicou . What version of System.ClientModel are you using?

eliasandronicou:
1.1.0-beta.5

joseharriaga:
@eliasandronicou  We're preparing to release a new version of the library in the next few days that will include the upgrade of the System.ClientModel dependency to version `1.1.0-beta.5`. For now, version `2.0.0-beta.7` of the OpenAI library uses System.ClientModel version `1.1.0-beta.4`, so if you switch to that version, the error should go away.

PR #161, opened 2024-08-07
Updated the 'How to work with Azure OpenAI' section with …

AngelosP:
… instructions to use Azure OpenAI client library for .NET and where to find it + some sample code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -678,7 +678,7 @@ The first image shows a red apple with a smooth skin and a single leaf, while th
 
 ## How to work with Azure OpenAI
 
-For Azure OpenAI scenarios we recommend using the [Azure SDK](https://github.com/Azure/azure-sdk-for-net) and more specifically the [Azure OpenAI client library for .NET](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/README.md). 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -678,7 +678,29 @@ The first image shows a red apple with a smooth skin and a single leaf, while th
 
 ## How to work with Azure OpenAI
 
-Details for using the OpenAI .NET library with [Azure OpenAI](https://learn.microsoft.com/azure/ai-services/openai/overview) are coming soon. Please watch here and the [Azure.AI.OpenAI project](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/openai/Azure.AI.OpenAI) for updates.
+For Azure OpenAI scenarios we recommend using the [Azure SDK](https://github.com/Azure/azure-sdk-for-net) and more specifically the [Azure OpenAI client library for .NET](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/README.md).
```
KrzysztofCwalina: "recommend" together with the paragraphs below sounds like it's not absolutely necessary but just a recommendation. I think it should be worded more strongly.

AngelosP: > "recommend" together with the paragraphs below sounds like it's not absolutely necessary but just a recommendation. I think it should be worded more strongly.

Good point, I've worded it more like an instruction and less like a recommendation now.

KrzysztofCwalina (approved):

PR #185, opened 2024-08-26
Bugfix missing internal options set

### LINKED ISSUE #184: OpenAIClient Constructor not persisting `_options` internally, fails for custom options (+endpoints).
### Confirm this is not an issue with the underlying OpenAI API

- [X] This is an issue with the Python library

### Confirm this is not an issue with Azure OpenAI

- [X] This is not an issue with Azure OpenAI

### Describe the bug

On `OpenAIClient` `options`, constructor is missing set of internal `_options` = options`.

Here: 
https://github.com/openai/openai-dotnet/blob/4897e23abfa68aa51853ccfd4e21220d4ea7f333/src/Custom/OpenAIClient.cs#L101

The impact of those changes is big as any specialized clients created from `OpenAIClient` lose specific options, including custom endpoints.

Suggest adding UnitTests to ensure the behavior.

### To Reproduce

When running the test below, it will never reach localhost, instead will go for public OpenAI API with a 401.

```csharp
private sealed class HttpMessageHandlerStub : DelegatingHandler
{
    public HttpRequestHeaders RequestHeaders { get; private set; }

    public HttpContentHeaders ContentHeaders { get; private set; }

    public byte[] RequestContent { get; private set; }

    public Uri RequestUri { get; private set; }

    public HttpMethod Method { get; private set; }

    public HttpResponseMessage ResponseToReturn { get; set; }

    public HttpMessageHandlerStub()
    {
        InnerHandler = new HttpClientHandler();
        this.ResponseToReturn = new HttpResponseMessage(HttpStatusCode.OK)
        {
            Content = new StringContent("{}", Encoding.UTF8, "application/json")
        };
    }

    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
    {
        this.Method = request.Method;
        this.RequestUri = request.RequestUri;
        this.RequestHeaders = request.Headers;
        if (request.Content is not null)
        {
            this.RequestContent = await request.Content.ReadAsByteArrayAsync(cancellationToken);
        }

        this.ContentHeaders = request.Content?.Headers;
        return await Task.FromResult(this.ResponseToReturn);
    }
}

[Test]
public void CustomEndpoint()
{
    using var handler = new HttpMessageHandlerStub();
    using var httpClient = new HttpClient(handler, false);
    Uri fakeUri = new("https://localhost:1234");
    OpenAIClient client = new("sk-not-a-real-key", new OpenAIClientOptions()
    {
        Endpoint = fakeUri,
        ApplicationId = "my-app-id",
        NetworkTimeout = TimeSpan.FromSeconds(10),
        RetryPolicy = new ClientRetryPolicy(),
        Transport = new HttpClientPipelineTransport(httpClient)
    });

    client.GetChatClient("gpt-4o").CompleteChat(["Hello, world!"]);

    Assert.That(handler.RequestUri, Is.EqualTo(fakeUri));
}

```

### Code snippets

_No response_

### OS

Windows 11

### .NET version

.net 8

### Library version

beta.9

RogerBarreto:
# Motivation
When creating `OpenAIClient` using `ApiKey` + `Options` the options are not persisted and passed when creating the specialized clients `Chat`, `Embedding`, etc resetting unintentionally the options for the specialized clients, including the `Endpoint`.

# Fix

The fix set the internal options restoring the desired behavior.

Resolve #184

### DIFF
--- a/src/Custom/OpenAIClient.cs
+++ b/src/Custom/OpenAIClient.cs
@@ -98,6 +98,7 @@ public partial class OpenAIClient
 
         _pipeline = OpenAIClient.CreatePipeline(credential, options);
         _endpoint = OpenAIClient.GetEndpoint(options);
+        _options = options;
     }
 
     // CUSTOM: Added protected internal constructor that takes a ClientPipeline.


trrwilson:
Thank you, @RogerBarreto! This fix is rolled up into #186; I wanted to be sure we also had a test that validated. We'll get this released shortly.

PR #187, opened 2024-08-26
Telemetry: update enablement (experimental source instead of app context switch) and docs improvements

lmolkova:
Based on the discussion in https://github.com/dotnet/aspire/issues/5451, suggesting a new way to onboard:
- ActivitySource and Meter names start with `Experimental.*`  which make their status obvious
- App context switch is removed

This PR also adds docs on how to view telemetry


examples/OpenTelemetryExamples.cs
```diff
@@ -0,0 +1,51 @@
...
+            .SetResourceBuilder(resourceBuilder)
+            .AddSource("OpenAI*")
+            .AddHttpClientInstrumentation()
+            .AddConsoleExporter()
+            .AddOtlpExporter()
+            .Build();
+
+        using MeterProvider meterProvider = OpenTelemetry.Sdk.CreateMeterProviderBuilder()
+            .SetResourceBuilder(resourceBuilder)
+            .AddView("gen_ai.client.operation.duration", new Base2ExponentialBucketHistogramConfiguration { MaxSize = 10 })
```
lmolkova: @cijothomas do you know if there is a nicer way to do it?  Without it histogram boundaries start at 5 sec, but you probably know 😢 

We have to depend on DiagnosticSource 6.0 here and can't use advisory parameters at least yet.

cijothomas: Without Advisory Parameter, the only way is with Views. But I am unsure why was Exponential histogram chosen? You can just use ExplicitBuckets itself with custom buckets: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics/customizing-the-sdk#explicit-bucket-histogram-aggregation

ExponentialHistogram would work fine, but I am not sure if many backends support. Also, [prometheus exporters](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Exporter.Prometheus.AspNetCore) don't support them as well.

Once OTel .NET supports selecting between Explicit vs Exponential via [env variable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md#additional-environment-variable-configuration), that can be shown here to pick Exponential Buckets.

docs/observability.md
```diff
@@ -14,44 +14,45 @@ The instrumentation is **experimental** - volume and semantics of the telemetry
...
-   - set the `OPENAI_EXPERIMENTAL_ENABLE_OPEN_TELEMETRY` environment variable to `"true"`
-   - set the `OpenAI.Experimental.EnableOpenTelemetry` context switch to true in your application code when application
-     is starting and before initializing any OpenAI clients. For example:
-
-     ```csharp
-     AppContext.SetSwitch("OpenAI.Experimental.EnableOpenTelemetry", true);
-     ```
-
-2. Enable OpenAI telemetry:
+1. Enable experimental OpenAI telemetry:
```
cijothomas: nit: Add some note about what is experimental means here. Its good to call out that the name of span/metric and each attributes are subject to change /similar.

docs/observability.md
```diff
@@ -14,44 +14,45 @@ The instrumentation is **experimental** - volume and semantics of the telemetry
...
 
    ```csharp
    builder.Services.AddOpenTelemetry()
        .WithTracing(b =>
        {
-           b.AddSource("OpenAI.*")
+           b.AddSource("Experimental.OpenAI.*", "OpenAI.*")
```
cijothomas: nit: I think its best to mention in comment that "if you want to avoid changing code when things move to stable, add the m both now.", instead of showing them.

examples/OpenTelemetryExamples.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace OpenAI.Examples;
+
+public partial class ChatExamples
+{
+    [Test]
+    public async Task OpenTelemetryExamples()
+    {
+        // Let's configure OpenTelemetry to collect OpenAI and HTTP client traces and metrics
+        // and export them to console and also to the local OTLP endpoint.
```
cijothomas, line 17: nit: Maybe shown in comment how to run a otel collector OR show how to run aspire dashboard?

tests/Telemetry/ChatTelemetryTests.cs
```diff
@@ -175,11 +155,10 @@ public void ChatTracingException()
     [Test]
     public async Task ChatTracingAndMetricsMultiple()
     {
-        using var _ = TestAppContextSwitchHelper.EnableOpenTelemetry();
         var source = new OpenTelemetrySource(RequestModel, new Uri(Endpoint));
 
-        using var activityListener = new TestActivityListener("OpenAI.ChatClient");
-        using var meterListener = new TestMeterListener("OpenAI.ChatClient");
+        using var activityListener = new TestActivityListener("Experimental.OpenAI.ChatClient");
```
cijothomas, line 162: I wish everyone uses OTel's InMemoryExporter rather than writing own listeners, but it is upto the owners of this project!

cijothomas (approved):
LGTM!

tests/Telemetry/ChatTelemetryTests.cs
lmolkova: can you explain why?

docs/observability.md
lmolkova: It's called out in the paragraph before this, I'll expand it a bit

docs/observability.md
lmolkova: Added some info below where I also explain what happens on these lines

examples/OpenTelemetryExamples.cs
```diff
@@ -0,0 +1,49 @@
...
+            .SetResourceBuilder(resourceBuilder)
+            .AddSource("Experimental.OpenAI.*", "OpenAI.*")
+            .AddHttpClientInstrumentation()
+            .AddConsoleExporter()
+            .AddOtlpExporter()
+            .Build();
+
+        using MeterProvider meterProvider = OpenTelemetry.Sdk.CreateMeterProviderBuilder()
+            .SetResourceBuilder(resourceBuilder)
+            .AddView("gen_ai.client.operation.duration", new ExplicitBucketHistogramConfiguration { Boundaries = [0.01, 0.02, 0.04, 0.08, 0.16, 0.32, 0.64, 1.28, 2.56, 5.12, 10.24, 20.48, 40.96, 81.92] })
```
cijothomas, line 36: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/CHANGELOG.md#1100-beta1 added support for Hints, so Views maybe replaced with the hint API itself. 
(Assuming this can take a dependency on preview DS,OTel nugets)

tests/Telemetry/ChatTelemetryTests.cs
cijothomas: Listeners are not extensively documented given its generally only used by small set of people like people authoring OpenTelemetry SDK etc.
OTOH, InMemoryExporter is documented, and is intended to be used by end-users to validate their own instrumentation.  Changing from InMemoryExporter to anything else like Console/OTLP is often the one-line change. End-users, who are familiar with OTLP Exporter find it easy to use InMemoryExporter with OTel for testing, as opposed to learning ActivityListener/MeterListener etc.

Just a suggestion only.

Issue #155, opened 2024-08-05
Move service methods for paginated endpoints to use PageCollection pattern


annelo-msft:
### Confirm this is not a feature request for the underlying OpenAI API.

- [X] This is not a feature request for the underlying OpenAI API

### Confirm this is not a feature request for Azure OpenAI.

- [X] This is not a feature request for Azure OpenAI

### Describe the feature or improvement you're requesting

The following service endpoints return paginated lists of results and should move to the SCM-based client PageCollection pattern:

- [ ] [List fine-tuning jobs](https://platform.openai.com/docs/api-reference/fine-tuning/list)
- [ ] [List fine-tuning events](https://platform.openai.com/docs/api-reference/fine-tuning/list-events)
- [ ] [List fine-tuning checkpoints](https://platform.openai.com/docs/api-reference/fine-tuning/list-checkpoints)
- [ ] [List batch](https://platform.openai.com/docs/api-reference/batch/list)

The pattern is illustrated for the [list assistants](https://platform.openai.com/docs/api-reference/assistants/listAssistants) and [list vector stores](https://platform.openai.com/docs/api-reference/vector-stores/list) endpoints in [this PR](https://github.com/openai/openai-dotnet/pull/105).

PR #188, opened 2024-08-26
Move service methods for paginated endpoints to use SCM based pagination pattern 

ShivangiReja:
This PR updates the protocol methods for the following service endpoints, which return paginated lists of results, and moves them to the SCM-based client pagination pattern:

- [x] [List fine-tuning jobs](https://platform.openai.com/docs/api-reference/fine-tuning/list)
- [x] [List fine-tuning events](https://platform.openai.com/docs/api-reference/fine-tuning/list-events)
- [x] [List fine-tuning checkpoints](https://platform.openai.com/docs/api-reference/fine-tuning/list-checkpoints)
- [x] [List batches](https://platform.openai.com/docs/api-reference/batch/list)

Fixes: https://github.com/openai/openai-dotnet/issues/155

### DIFF
--- a/src/Custom/Batch/Internal/Pagination/BatchesPageEnumerator.cs
+++ b/src/Custom/Batch/Internal/Pagination/BatchesPageEnumerator.cs
@@ -1,7 +1,6 @@
 using System;
 using System.ClientModel;
 using System.ClientModel.Primitives;
-using System.Collections.Generic;
 using System.Text.Json;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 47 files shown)


src/Custom/Batch/Internal/Pagination/BatchesPageEnumerator.cs
```diff
@@ -0,0 +1,109 @@
...
+
+    internal PipelineMessage CreateGetBatchesRequest(string after, int? limit, RequestOptions options)
+    {
+        var message = _pipeline.CreateMessage();
+        message.ResponseClassifier = PipelineMessageClassifier200;
+        var request = message.Request;
+        request.Method = "GET";
+        var uri = new ClientUriBuilder();
+        uri.Reset(_endpoint);
+        uri.AppendPath("/batches", false);
```
joseharriaga: A change just went in last week that removed the "v1" version path parameter from the default endpoint. Which means that this parameter must be added to the URI in each operation instead. Here, for example, instead "/batches", it should now be "/v1/batches". Same in the fine-tuning URIs.

joseharriaga (approved):

PR #206, opened 2024-09-01
Fix deserialization of run steps when using file search and mitigate .NET runtime issue when serializing `ChatResponseFormat`

joseharriaga:
* Fixed the deserialization of run steps when using file search. The internal representation of File Search as an `IReadOnlyDictionary<string, string>` is incorrect and causing issues during deserialization after the latest service. This fix changes this internal representation to `IReadOnlyDictionary<string, BinaryData>` instead.
* Mitigated a .NET runtime issue that prevented `ChatResponseFormat` from serializing correct on targets including Unity.
* Renamed `SpeechGenerationOptions`'s `Speed` property to `SpeedRatio`.
* Changed `GeneratedSpeechFormat` from an enum to an extensible enum.


Pauluz26 (approved):
Would be great if this can be fixed, as we are running in production application problems because we use the FileSearch a lot.

Pauluz26 (approved):

trrwilson (approved):

joseharriaga:
@Pauluz26 This fix is now live starting with version 2.0.0-beta.11:
🔗 https://www.nuget.org/packages/OpenAI/2.0.0-beta.11

Thank you!

Pauluz26:
Great! I  will do some testings and update our deployments!

Issue #244, opened 2024-10-04
(Beta) Websocket closing unexpectedly on ReceiveUpdatesAsync


camelCase12:
### Service

OpenAI

### Describe the bug

I was setting up the realtime API for voice chatting, but I found after my first question and reply from the model, the web socket would close and terminate the conversation. (This was while using `CreateServerVoiceActivityTurnDetectionOptions`)

When calling `ReceiveUpdatesAsync` to receive server messages, I didn't expect this to close the `_clientWebSocket`, but it did.

 It took me a while to figure out that the `AsyncWebsocketMessageResultEnumerator` is automatically disposing of the `_clientWebSocket` when it finishes iterating, causing `ReceiveUpdatesAsync` to terminate the WebSocket. Commenting out the disposing call within `AsyncWebsocketMessageResultEnumerator `resulted in the expected behavior for me:

```csharp
public ValueTask DisposeAsync()
{
    //_clientWebSocket?.Dispose();
    return new ValueTask(Task.CompletedTask);
}
```

This allowed me to have my expected 2-way conversation. If there's another intended method to receive server events without terminating the socket, or to keep the socket alive for more than one request-response, please let me know.

### Steps to reproduce

1. Initialize a `RealtimeConversationSession` with server voice activity turn detection:
```csharp
var client = new RealtimeConversationClient(model: "gpt-4o-realtime-preview-2024-10-01", new(apiKey));


CancellationTokenSource cts = new();

var session = await client.StartConversationSessionAsync(cts.Token);

var options = new ConversationSessionOptions()
{
    Instructions = "<system prompt>",
    TurnDetectionOptions = ConversationTurnDetectionOptions.CreateServerVoiceActivityTurnDetectionOptions(0.5f, TimeSpan.FromMilliseconds(300), TimeSpan.FromMilliseconds(200)),
    Voice = ConversationVoice.Alloy,
    OutputAudioFormat = ConversationAudioFormat.Pcm16,
    InputTranscriptionOptions = new ConversationInputTranscriptionOptions()
    {
        Model = "whisper-1"
    }
};

await session.ConfigureSessionAsync(options);
```

2. Begin sending audio through `SendAudioAsync` (in my case with NAudio Wave):
```csharp
waveIn.DataAvailable += (s, a) =>
{
    using var memoryStream = new MemoryStream();
    memoryStream.Write(a.Buffer, 0, a.BytesRecorded);
    memoryStream.Position = 0;
[...]

PR #261, opened 2024-10-18
Fix issue #244

isaac-j-miller:
- Don't dispose the client websocket in the AsyncWebsocketMessageEnumerator
- Bump version suffix to beta.2 (not sure if this is correct)


isaac-j-miller:
#261

PR #262, opened 2024-10-20
fix: the GetBatch by id should be available

mickaelMamani:
as the GetBatch by Id if not available then we must browse all the batches

### DIFF
--- a/src/Custom/Batch/BatchClient.Protocol.cs
+++ b/src/Custom/Batch/BatchClient.Protocol.cs
@@ -102,7 +102,7 @@ public partial class BatchClient
     /// <exception cref="ArgumentException"> <paramref name="batchId"/> is an empty string, and was expected to be non-empty. </exception>
     /// <exception cref="ClientResultException"> Service returned a non-success status code. </exception>
     /// <returns> The response returned from the service. </returns>
-    internal virtual async Task<ClientResult> GetBatchAsync(string batchId, RequestOptions options)
+    public virtual async Task<ClientResult> GetBatchAsync(string batchId, RequestOptions options)
     {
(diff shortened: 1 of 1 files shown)


LuisM000:
Hi,
Is there any reason why this PR (#262) is stalled?
Do we have an estimate on when this change will be available or if it needs further review?

Thanks!

shargon:
Seems that the project is unactive

legendaryleo7:
Seems like this entire project is generated so it’s likely no one is watching the PRs

KrzysztofCwalina:
@mickaelMamani, sorry for a delay in this PR. The reason for the delay is that we would like to support this scenario with what we call "operation rehydration". We already have a method CreateBatchOperation.Rehydrate that allows you to retrieve a previously initiated batch operation. The problem is that there is currently no way to rehydrate from batchId. We would like to add CreateBatchOperation.Rehydrate(BatchClient client, int batchId, ...) overload, but the addition got postponed due to other work. I will bump the priority of the addition to get it done soon.

I filed issue: https://github.com/openai/openai-dotnet/issues/463

KrzysztofCwalina:
the issue I filed has been fixed and merged: https://github.com/openai/openai-dotnet/pull/464

jsquire (approved):
Hi @mickaelMamani. Thanks very much for your contribution and interest in improving the OpenAI developer experience. We apologize for the long wait for someone to review.

As Krzysztof mentioned, there are still ongoing design conversations around the experience that we'd like to offer for the batch API - including at the protocol level.  Since this was opened, we've changed our approach to evolving the API which will allows us to accept this and continue to iterate on the design over time.

I'll get this merged in. Thank you, again, for your efforts.

PR #267, opened 2024-10-24
add eos finish reason that maps to Stop for llama models

kcrandall:
I got this error below when using SemanticKernel with Together.ai as an open ai compatible endpoint. It seems like they return `eos` instead of the normal `stop` string thowing an exception just before completion.

System.ArgumentOutOfRangeException: Unknown ChatFinishReason value. (Parameter 'value')\nActual value was eos.\n   at OpenAI.Chat.ChatFinishReasonExtensions.ToChatFinishReason(String value)\n   at OpenAI.Chat.InternalCreateChatCompletionStreamResponseChoice.DeserializeInternalCreateChatCompletionStreamResponseChoice(JsonElement element, ModelReaderWriterOptions options)\n   at OpenAI.Chat.StreamingChatCompletionUpdate.DeserializeStreamingChatCompletionUpdate(JsonElement element, ModelReaderWriterOptions options)\n   at

### DIFF
--- a/src/Generated/Models/ChatFinishReason.Serialization.cs
+++ b/src/Generated/Models/ChatFinishReason.Serialization.cs
@@ -25,6 +25,7 @@ namespace OpenAI.Chat
             if (StringComparer.OrdinalIgnoreCase.Equals(value, "tool_calls")) return ChatFinishReason.ToolCalls;
             if (StringComparer.OrdinalIgnoreCase.Equals(value, "content_filter")) return ChatFinishReason.ContentFilter;
             if (StringComparer.OrdinalIgnoreCase.Equals(value, "function_call")) return ChatFinishReason.FunctionCall;
+            if (StringComparer.OrdinalIgnoreCase.Equals(value, "eos")) return ChatFinishReason.Stop;
             throw new ArgumentOutOfRangeException(nameof(value), value, "Unknown ChatFinishReason value.");
         }
     }


sergeyleliukh:
the same problem, please add eos finish reason

shargon (approved):

jsquire (changes requested):
Hi @kcrandall.  Thank you for your contribution and for your interest in improving the OpenAI developer experience.   The OpenAI library is intended to conform to the OpenAI REST API and work with the OpenAI service.  While we appreciate that it can be used with other services/models, it is not a goal of the project to support other services where they deviate from the OpenAI implementation.

In this scenario, the [platform documentation](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices) is explicit in the set of supported values for `finish_reason` under `choice` for a chat completion object:

```text
finish_reason  (string)

The reason the model stopped generating tokens. This will be stop if the model hit a natural stop point or a provided stop sequence, length if the maximum number 
of tokens specified in the request was reached, content_filter if content was omitted due to a flag from our content filters, tool_calls if the model called a tool, or 
function_call (deprecated) if the model called a function.
```

The REST API specification defines this more explicitly as:
```yml
 choices:
   type: array
   description: A list of chat completion choices. Can be more than one if `n` is
     greater than 1.
   items:
     type: object
     required:
       - finish_reason
       - index
       - message
       - logprobs
     properties:
       finish_reason:
         type: string
         description: >
           The reason the model stopped generating tokens. This will be
           `stop` if the model hit a natural stop point or a provided
           stop sequence,
           `length` if the maximum number of tokens specified in the
           request was reached,
           `content_filter` if content was omitted due to a flag from our
           content filters,
           `tool_calls` if the model called a tool, or `function_call`
           (deprecated) if the model called a function.
         enum:
           - stop
           - length
           - tool_calls
           - content_filter
           - function_call
```

Unfortunately, we cannot accept this submission, as we do not want to set the precedent that we offer explicit support for non-OpenAI services/models.  I'm going to close this out. Thank you, again, for your efforts.

Issue #272, opened 2024-10-29
Add a net8.0 target


m-redding:
### Confirm this is a feature request for the .NET library and not the underlying OpenAI API

- [X] This is a feature request for the .NET library

### Describe the feature or improvement you are requesting

It would be great if OpenAI could add a net8.0 target, in order to unblock adding a net8.0 target to Azure.AI.OpenAI. There are more details in this issue: https://github.com/Azure/azure-sdk-for-net/issues/46891

Alternatively, resolving https://github.com/openai/openai-dotnet/issues/259 would also unblock adding this target to Azure.AI.OpenAI

Frulfump:
Target was added in https://github.com/openai/openai-dotnet/pull/277 so available since OpenAI_2.2.0-beta.1 (no GA release yet)

m-redding:
Yeah looks like this was fixed at some point since Azure OpenAI was able to add the net8.0 target https://github.com/Azure/azure-sdk-for-net/pull/48350

PR #277, opened 2024-11-01
Added net8.0 as a supported target framework for OpenAI #272

armanossiloko:
I honestly do not know why this has not been done by now. NET 8 is the latest dotnet version and is also LTS.

All the examples and tests projects are already targeting `net8.0` so basically the only the thing needing to be done is add it to the supported target frameworks in the OpenAI.csproj

Besides, from what I understood, this would help with #272 and the change proposed in this PR was also suggested by @RogerBarreto in #98, however that PR was never updated since August.

### DIFF
--- a/src/OpenAI.csproj
+++ b/src/OpenAI.csproj
@@ -9,7 +9,7 @@
     <VersionPrefix>2.1.0</VersionPrefix>
     <VersionSuffix>beta.1</VersionSuffix>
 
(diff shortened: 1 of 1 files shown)


Barry-Eventio (approved):

joseharriaga (approved):

AlecVermeer:
👀

joseharriaga:
Thank you for your contribution, @armanossiloko !

Issue #268, opened 2024-10-27
SendAudioAsync does not set 'isSendingAudio' back to false


onionhammer:
### Service

OpenAI

### Describe the bug

https://github.com/openai/openai-dotnet/blob/b8ec620316c66ce9f7807c2efe9e2058ada1c791/src/Custom/RealtimeConversation/RealtimeConversationSession.cs#L95

In the 'BinaryData' overload, subsequent calls will fail with exception since `_isSendingAudio` is not set back to `false` as it does in the memorystream overload

### Steps to reproduce

This is appears to be a logical/code error

### OS

*

### .NET version

.NET 8

### Library version

2.1.0-*

onionhammer:
How is this still not fixed? Isn't there a failing test suite somewhere?

onionhammer:
Bump.

joseharriaga:
Thank you for your contribution, @onionhammer ! Have you tried the latest stable 2.1.0 version from early December? I believe this issue is fixed in that one, but let me know if you're still seeing it. 

Here's a link:
🔗 https://www.nuget.org/packages/OpenAI/2.1.0

onionhammer:
Looks like the issue is that the only (working) version of 'Azure.AI.OpenAI' is referencing the broken version of your nuget package.

joseharriaga:
Ah, so this is an issue with the Azure OpenAI library (Azure.AI.OpenAI) rather than the OpenAI library? Got it. I think I see what the problem is now:

Azure OpenAI versions a little differently than OpenAI. In particular, a stable version of the Azure OpenAI service does not include beta APIs (such as the Realtime API). Since version 2.1.0 of the Azure OpenAI library is a stable version and targets the latest stable version of the Azure service, it does not include beta APIs either, and that is why the fix is missing there. 

We're preparing to publish new 2.2.0-beta.1 versions of both libraries in a few days. This beta library of Azure OpenAI will target the beta versions of the service, and therefore it will once again include the Realtime APIs (along with this fix). I will post back here as soon as it is out.

onionhammer:
Ah that makes sense, thanks!

joseharriaga:
There is a new version of both the Azure OpenAI library:
🔗 https://www.nuget.org/packages/Azure.AI.OpenAI/2.2.0-beta.1

Let me know if the fix works correctly for you. 🙂

onionhammer:
@joseharriaga looks fixed, thanks!

PR #298, opened 2024-12-02
fix: should set _isSendingAudio to false after sending audio BinaryData

sunnyden:
`_isSendingAudio` was not set back to `false` and will cause exception when calling the `SendInputAudioAsync` with `BinaryData` repeatedly.


joseharriaga:
Thank you for your contribution, @sunnyden ! This issue has been fixed starting with version 2.1.0 for OpenAI and with version 2.2.0-beta.1 for Azure OpenAI.
🔗 https://www.nuget.org/packages/OpenAI/AbsoluteLatest
🔗 https://www.nuget.org/packages/Azure.AI.OpenAI/AbsoluteLatest

Here is the link to the issue: 
🔗 https://github.com/openai/openai-dotnet/issues/268

PR #324, opened 2025-01-14
Fixed access modifier for file ids

walkeryan:
File ID's should be able to be set as they are in the python package.

### DIFF
--- a/src/Custom/Assistants/CodeInterpreterToolResources.cs
+++ b/src/Custom/Assistants/CodeInterpreterToolResources.cs
@@ -14,7 +14,7 @@ public partial class CodeInterpreterToolResources
     public IList<string> FileIds
     {
         get => _fileIds;
-        internal set
+        set
         {
             _fileIds = new ChangeTrackingList<string>();
             foreach (string fileId in value)


walkeryan:
Good morning @romainhuet , @trevorcreech , @PaulMcMillan , @schnerd . I've been using this code reliably for the past week or so. Can we get this merged in so I don't have to continue maintaining the project and can switch back to the Nuget package? <3

joseharriaga:
Thank you for your contribution, @walkeryan ! Based on our current design guidelines for .NET, our collection properties do not have setters. Instead, they are guaranteed to be initialized, and users can add or remove items directly without having to `new` up the collection. For example:

```csharp
CodeInterpreter = new() 
{ 
    FileIds = { myFileId1, myFIleId2 } 
}
```

If needed, users can replace the contents of the collection by clearing the collection first and then adding the new contents. For example:

```csharp
toolResources.CodeInterpreter.Clear();
toolResources.CodeInterpreter.Add(myFile1);
toolResources.CodeInterpreter.Add(myFile2);
```

One big benefit of this approach is that users can safely use collection properties without having to check if the property is null each time.

If we were to make a change related to this, we would like to be able to apply it consistently across the entire library. As such, I will close this PR for now, but your contribution here is a very helpful data point for us as we continue evolving our current design guidelines and polishing the library. So once again thank you!

walkeryan:
Thanks for the clarification, not sure why, but I tried it this way and had to switch over to the python package.

wjdavis5:
I struggled with this for way longer than I'd care to admit.

PR #338, opened 2025-02-07
Update RealtimeConversationClient.cs

BlinkSun:
uriBuilder already concat the ? into the uri;
that make double ?? and cause problem;

### DIFF
--- a/src/Custom/RealtimeConversation/RealtimeConversationClient.cs
+++ b/src/Custom/RealtimeConversation/RealtimeConversationClient.cs
@@ -87,7 +87,7 @@ public partial class RealtimeConversationClient
             uriBuilder.Path += uriBuilder.Path[uriBuilder.Path.Length - 1] == '/' ? "realtime" : "/realtime";
         }
 
-        uriBuilder.Query = $"?model={model}";
+        uriBuilder.Query = $"model={model}";
 
         return uriBuilder.Uri;
     }


BlinkSun:
image

image

joseharriaga:
Thank for your contribution, @BlinkSun ! Question: What version of .NET are you using?

BlinkSun:
Framework 4.8 !

BlinkSun:
Hey @joseharriaga with that fix ? I need to recompile manually with my fix the lib to use it in my framework 4.8 app :(

joseharriaga:
@trrwilson : I recall you mentioned that there are currently limitations in the RealtimeConversationClient related to .NET Framework. Could you provide more information and how it relates to this issue?

BlinkSun:
The fix work but if you want to stay in your codeline, you can fix it by using your ClientUriBuilder class that you use everywhere !

jsquire (changes requested):
Hi @BlinkSun. Thanks very much for your contribution and interest in improving the OpenAI developer experience. We apologize for the long wait for someone to review.

Unfortunately, we cannot accept this submission, as the client type, namespace, and implementation have changed since this was opened.  The [current implementation](https://github.com/openai/openai-dotnet/blob/main/src/Custom/Realtime/RealtimeClient.cs#L127) uses a different service path that has no querystring component.

I'm going to close this out. Thank you, again, for your efforts.

PR #353, opened 2025-02-19
* GenericActionPipelinePolicy - ConfigureAwait(false) for ProcessAsync method

vanek021:
If this was not intentional, it might be better to add ConfigureAwait(false) to this method. In my case, there was dedlock with Blazor Server.
I think it would be better if the synchronization context was not captured inside the library


joseharriaga (approved):

joseharriaga:
Thank you for your contribution, @vanek021 ! Great catch!

PR #389, opened 2025-03-25
some implement of streaming api may return "" as finish_reason 

migeyusu:
some implement of streaming api may return "" as finish_reason such as o3 (https://o3.fan/).

### DIFF
--- a/src/Custom/Chat/Streaming/InternalCreateChatCompletionStreamResponseChoice.Serialization.cs
+++ b/src/Custom/Chat/Streaming/InternalCreateChatCompletionStreamResponseChoice.Serialization.cs
@@ -83,7 +83,7 @@ internal partial class InternalCreateChatCompletionStreamResponseChoice : IJsonM
             {
                 var propertyValue = property.Value;
                 string propertyValueString;
-                if (propertyValue.ValueKind == JsonValueKind.Null||(propertyValueString=propertyValue.GetString())?.Trim()==string.Empty)
+                if (propertyValue.ValueKind == JsonValueKind.Null||string.IsNullOrEmpty((propertyValueString=propertyValue.GetString())?.Trim()))
                 {
                     finishReason = null;
                     continue;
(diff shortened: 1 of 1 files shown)


migeyusu:
QQ20250325-122743
as you see , some implement will use "" as finish_reason, this works well in cherry studio, but throw exception in OpenAI:
"Unknown ChatFinishReason value."
It should be regarded as null and goto continue.

jsquire:
Hi @migeyusu.  Thank you for your contribution and for your interest in improving the OpenAI developer experience.   The OpenAI library is intended to conform to the OpenAI REST API and work with the OpenAI service.  While we appreciate that it can be used with other services/models, it is not a goal of the project to support other services where they deviate from the OpenAI implementation.

In this scenario, the [platform documentation](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices) is explicit in the set of supported values for `finish_reason` under `choice` for a chat completion object:

```text
finish_reason  (string)

The reason the model stopped generating tokens. This will be stop if the model hit a natural stop point or a provided stop sequence, length if the maximum number 
of tokens specified in the request was reached, content_filter if content was omitted due to a flag from our content filters, tool_calls if the model called a tool, or 
function_call (deprecated) if the model called a function.
```

The REST API specification defines this more explicitly as:
```yml
 choices:
   type: array
   description: A list of chat completion choices. Can be more than one if `n` is
     greater than 1.
   items:
     type: object
     required:
       - finish_reason
       - index
       - message
       - logprobs
     properties:
       finish_reason:
         type: string
         description: >
           The reason the model stopped generating tokens. This will be
           `stop` if the model hit a natural stop point or a provided
           stop sequence,
           `length` if the maximum number of tokens specified in the
           request was reached,
           `content_filter` if content was omitted due to a flag from our
           content filters,
           `tool_calls` if the model called a tool, or `function_call`
           (deprecated) if the model called a function.
         enum:
           - stop
           - length
           - tool_calls
           - content_filter
           - function_call
```

While I wish that explicitly stated `null` as a valid value, it is consistent with OpenAI patterns in the spec and demonstrated by all examples showing `null` as the appropriate value for the "no finish_reason" scenario:
```text
{"id":"chatcmpl-123","object":"chat.completion.chunk","created":1694268190,"model":"gpt-4o-mini", "system_fingerprint": "fp_44709d6fcb", "choices":[{"index":0,"delta":{"role":"assistant","content":""},"logprobs":null,"finish_reason":null}]}
```

Unfortunately, we cannot accept this submission, as we do not want to set the precedent that we offer explicit support for non-OpenAI services/models.  I'm going to close this out. Thank you, again, for your efforts.

PR #422, opened 2025-05-02
Add ChatTools and ResponseTools helper classes

christothes:
This PR adds functionality for managing OpenAI function tools in chat completions and responses, implementing two utility classes: `ChatTools` and `ResponseTools`.

### Key Features
- Unified Tool Management
  - Both classes handle local and remote MCP (Model Context Protocol) tools
  - Support for both static method tools and remote MCP server tools
  - Vectorized tool lookup using embeddings for smart tool filtering
- Tool Registration
  - Local tool registration via reflection from static methods
  - Remote tool registration from MCP servers
- Tool Filtering
  - Vector-based tool filtering when an `EmbeddingClient` is provided
  - Configurable similarity thresholds
  - Customizable maximum tool limits


src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,316 @@
...
+        _client = client;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the ChatTools class with the specified tool types.
+    /// </summary>
+    /// <param name="tools">Additional tool types to add.</param>
+    public ChatTools(params Type[] tools) : this((EmbeddingClient)null)
+    {
+        foreach (var t in tools)
```
KrzysztofCwalina, line 44: nit: rename t to tool

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,316 @@
...
+            AddLocalTool(t);
+    }
+
+    /// <summary>
+    /// Gets the list of defined tools.
+    /// </summary>
+    public IList<ChatTool> Tools => _tools;
```
KrzysztofCwalina, line 51: should this be IReadOnlyList?

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /// <summary>
+    /// Converts the tools collection to <see cref="ChatCompletionOptions"/>, filtered by relevance to the given prompt.
+    /// </summary>
+    /// <param name="prompt">The prompt to find relevant tools for.</param>
+    /// <param name="maxTools">The maximum number of tools to return. Default is 3.</param>
+    /// <param name="minVectorDistance">The similarity threshold for including tools. Default is 0.29.</param>
+    /// <returns>A new <see cref="ChatCompletionOptions"/> containing the most relevant tools.</returns>
+    public ChatCompletionOptions CreateCompletionOptions(string prompt, int maxTools = 3, float minVectorDistance = 0.29f)
```
KrzysztofCwalina: should we increase the maxTools default?

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,316 @@
...
+    {
+        if (!CanFilterTools)
+            return _tools;
+
+        return FindVectorMatches(prompt, maxTools, minVectorDistance).Select(e => ParseToolDefinition(e.Data));
+    }
+
+    private IEnumerable<VectorbaseEntry> FindVectorMatches(string prompt, int maxTools, float minVectorDistance)
+    {
+        var vector = ToolsUtility.GetEmbedding(_client, prompt).GetAwaiter().GetResult();
```
KrzysztofCwalina: we should provide an async version of the CreateCompletionOptions so users dont need to sync over async in async context

src/Utility/MCP/McpSession.cs
```diff
@@ -0,0 +1,516 @@
...
+        var streamReader = new StreamReader(responseStream);
+        _ = Task.Run(async () =>
+        {
+            try
+            {
+                await ProcessSseStreamAsync(streamReader).ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                DebugPrint($"SSE processing failed: {ex.Message}");
```
KrzysztofCwalina, line 111: I wonder if we should not integrate with OTel or ILogger at some point.

src/Utility/MCP/McpSession.cs
```diff
@@ -0,0 +1,516 @@
...
+                _isInitialized = false;
+            }
+        }, _cancellationSource.Token);
+    }
+
+    private async Task ProcessSseStreamAsync(StreamReader streamReader)
+    {
+        string eventName = string.Empty;
+        var dataBuilder = new StringBuilder();
```
KrzysztofCwalina, line 120: At some point it would be nice if this was done in UTF8

src/Utility/MCP/McpSession.cs
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    private PipelineMessage CreateHandshakeMessage()
+    {
+        var message = _pipeline.CreateMessage();
+        var request = message.Request;
+        request.Uri = _serverEndpoint;
+        request.Method = "GET";
+        request.Headers.Add("Accept", "text/event-stream");
+        message.BufferResponse = false;
```
KrzysztofCwalina, line 280: this is just the handshake. why are we not buffering?

src/Utility/MCP/McpSession.cs
```diff
@@ -0,0 +1,516 @@
...
+
+    public void Dispose()
+    {
+        _cancellationSource.Cancel();
+        CleanupCurrentSession();
+        _initializationLock.Dispose();
+        _cancellationSource.Dispose();
+    }
+
+    public struct SseEvent
```
KrzysztofCwalina: should this be internal. Safer if we ever make the outer class public.

src/Utility/ResponseTools.cs
```diff
@@ -0,0 +1,313 @@
...
+    /// </summary>
+    /// <param name="mcpEndpoint">The URI endpoint of the MCP server.</param>
+    /// <returns>A task representing the asynchronous operation.</returns>
+    public async Task AddMcpToolsAsync(Uri mcpEndpoint)
+    {
+        var client = new McpClient(mcpEndpoint);
+        await AddMcpToolsAsync(client).ConfigureAwait(false);
+    }
+
+    private async Task AddToolsAsync(BinaryData toolDefinitions, McpClient client)
```
KrzysztofCwalina: could we share some of these complex routines between chat and responses? i.e. put them in the utility class?

src/Utility/ResponseTools.cs
```diff
@@ -0,0 +1,313 @@
...
+                var item = toolsToVectorize[embedding.Index];
+                var toolDefinition = SerializeTool(item);
+                _entries.Add(new VectorbaseEntry(vector, toolDefinition));
+            }
+        }
+    }
+
+    private string GetDescription(ResponseTool tool) => (tool as InternalResponsesFunctionTool)?.Description ?? "";
+
+    private BinaryData SerializeTool(ResponseTool tool)
```
KrzysztofCwalina, line 144: even this one could be shared if it took 3 parameters (name, description, parameters)

src/Utility/ToolsUtility.cs
```diff
@@ -0,0 +1,110 @@
...
+        var attr = param.GetCustomAttribute<DescriptionAttribute>();
+        if (attr != null)
+            description = attr.Description;
+        return description;
+    }
+
+    internal static ReadOnlySpan<byte> ClrToJsonTypeUtf8(Type clrType) =>
+        clrType switch
+        {
+            Type t when t == typeof(double) => "number"u8,
```
KrzysztofCwalina, line 40: we should add more types here

KrzysztofCwalina: we should also write a good doc remarks describing the type system we support.

src/Utility/ChatTools.cs
christothes: done

src/Utility/MCP/McpSession.cs
christothes: This is establishing the SSE session. We can't buffer here since we need to recieve events as they arrive.

src/Utility/VectorBaseEntry.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+
+namespace OpenAI;
+
+/// <summary>
+/// A vectorbase entry.
+/// </summary>
+internal readonly struct VectorbaseEntry
```
joseharriaga: nit: The filename does not match the class name.
nit: Consider renaming this one to "VectorDatabaseEntry", since "vectorbase" is a little niche.

src/Utility/ToolsUtility.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        writer.WriteEndObject(); // properties
+        writer.WriteStartArray("required");
+        foreach (string param in required)
+            writer.WriteStringValue(param);
+        writer.WriteEndArray();
+        writer.WriteEndObject();
+        writer.Flush();
+        stream.Position = 0;
+        return BinaryData.FromStream(stream);
```
joseharriaga, line 80: Are we planning to support structured outputs (a.k.a. "strict" mode)?

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,316 @@
...
+        if (!CanFilterTools)
+            return CreateCompletionOptions();
+
+        var completionOptions = new ChatCompletionOptions();
+        foreach (var tool in FindRelatedTools(prompt, maxTools, minVectorDistance))
+            completionOptions.Tools.Add(tool);
+        return completionOptions;
+    }
+
+    private IEnumerable<ChatTool> FindRelatedTools(string prompt, int maxTools, float minVectorDistance)
```
joseharriaga: What is the expectation for how this should work when combined with the `ToolChoice` property of `ChatCompletionOptions`? If I tell the model "I want you to call this specific function", but that function gets filtered out, the server will return an error.

tests/Utility/ChatToolsTests.cs
```diff
@@ -0,0 +1,329 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using OpenAI.Agents;
+
+namespace OpenAI.Tests.Utility;
+
+[TestFixture]
+[Category("Utility")]
+public class ChatToolsTests
```
joseharriaga: Consider adding a test for the following scenario: I have an ongoing chat where the model called "Function Foo" at some point, but then I remove "Function Foo" from the list of available tools and continue talking to the model. Confirm that when the model sees "Function Foo" in the chat history but not in its list of available tools, it still works correctly.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,326 @@
...
+    private readonly EmbeddingClient _client;
+    private readonly List<VectorbaseEntry> _entries = [];
+    private readonly List<McpClient> _mcpClients = [];
+    private readonly Dictionary<string, McpClient> _mcpClientsByEndpoint = [];
+
+    /// <summary>
+    /// Initializes a new instance of the ChatTools class with an optional embedding client.
+    /// </summary>
+    /// <param name="client">The embedding client used for tool vectorization, or null to disable vectorization.</param>
+    public ChatTools(EmbeddingClient client = null)
```
joseharriaga, line 33: Would this API look "friendlier" if this class had a default constructor? i.e.:

```csharp
public ChatTools();
public ChatTools(EmbeddingClient client);
```

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,326 @@
...
+    {
+        _client = client;
```
joseharriaga, line 35: Should this also receive `EmbeddingGenerationOptions` as an optional parameter (e.g., to allow users to customize the dimension of the embeddings)?

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+    /// <summary>
+    /// Gets whether tools can be filtered using embeddings provided by the provided <see cref="EmbeddingClient"/> .
+    /// </summary>
+    public bool CanFilterTools => _client != null;
+
+    /// <summary>
+    /// Adds local tool implementations from the provided types.
+    /// </summary>
+    /// <param name="tools">Types containing static methods to be used as tools.</param>
+    public void AddLocalTools(params Type[] tools)
```
joseharriaga: I think this should be called `AddFunctionTools`. This is because the term "local" feels a little overloaded with there being "local" and "remote" MCP servers too. In fact, I had to read the implementation to be sure that this was not actually related to MCP.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+    private async Task<IEnumerable<ChatTool>> FindRelatedTools(bool async, string prompt, int maxTools, float minVectorDistance)
+    {
+        if (!CanFilterTools)
+            return _tools;
+
+        return (await FindVectorMatches(async, prompt, maxTools, minVectorDistance).ConfigureAwait(false))
+            .Select(e => ParseToolDefinition(e.Data));
+    }
+
+    private async Task<IEnumerable<VectorbaseEntry>> FindVectorMatches(bool async, string prompt, int maxTools, float minVectorDistance)
```
joseharriaga, line 214: This feels a little bit like a "silent failure" to me, because this method doesn't do the expected thing unless this instance of `ChatTools` was created with an `EmbeddingClient`.

Considering that the `EmbeddingClient` is not used anywhere else, I think it would be a good improvement to remove the `EmbeddingClient` parameter from the `ChatTools` constructor and add it here as a required parameter instead. This fixes the "silent failure" and also reconciles all the functionality related to embeddings in a single place, which is very nice.

src/Utility/ChatTools.cs
joseharriaga: I added a comment below about removing the `EmbeddingClient` from the constructor and moving it to `CreateChatCompletionOptions` instead, so if we decide to include an `EmbeddingGenerationOptions` parameter, it should move there too.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+        var messages = new List<ToolChatMessage>();
+        foreach (ChatToolCall toolCall in toolCalls)
+        {
+            bool isMcpTool = false;
+            if (!_methods.ContainsKey(toolCall.FunctionName))
+            {
+                if (_mcpMethods.ContainsKey(toolCall.FunctionName))
+                {
+                    isMcpTool = true;
+                }
```
joseharriaga, line 282: We probably want to add support for awaiting async functions. For example, it's not uncommon to use functions to call web APIs.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+using OpenAI.Agents;
+using OpenAI.Chat;
+using OpenAI.Embeddings;
+
+namespace OpenAI.Chat;
+
+/// <summary>
+/// Provides functionality to manage and execute OpenAI function tools for chat completions.
+/// </summary>
+public class ChatTools
```
joseharriaga, line 19: Please add `[Experimental("OPENAIMCP001")]` on all the public APIs. 🙂

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+            bool isMcpTool = false;
+            if (!_methods.ContainsKey(toolCall.FunctionName))
+            {
+                if (_mcpMethods.ContainsKey(toolCall.FunctionName))
+                {
+                    isMcpTool = true;
+                }
+                else
+                {
+                    throw new InvalidOperationException("Tool not found: " + toolCall.FunctionName);
```
joseharriaga, line 285: Just like in `CallMcpAsync`, this should throw an exception instead of returning a message.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+                }
+            }
+
```
joseharriaga, line 288: Instead of returning `string.Empty`, this should probably throw an exception explaining that the result of a function cannot be null and the user needs to fix their function.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+            root.GetProperty("name").GetString()!,
+            root.GetProperty("description").GetString()!,
+            BinaryData.FromString(root.GetProperty("inputSchema").GetRawText()));
+    }
+
+    /// <summary>
+    /// Converts the tools collection to chat completion options.
+    /// </summary>
+    /// <returns>A new ChatCompletionOptions containing all defined tools.</returns>
+    public ChatCompletionOptions CreateCompletionOptions()
```
joseharriaga: Since this method is not static, should we call it `ToChatCompletionOptions` instead of `CreateCompletionOptions`?

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+    /// Converts the tools collection to <see cref="ChatCompletionOptions"/>, filtered by relevance to the given prompt.
+    /// </summary>
+    /// <param name="prompt">The prompt to find relevant tools for.</param>
+    /// <param name="maxTools">The maximum number of tools to return. Default is 3.</param>
+    /// <param name="minVectorDistance">The similarity threshold for including tools. Default is 0.29.</param>
+    /// <returns>A new <see cref="ChatCompletionOptions"/> containing the most relevant tools.</returns>
+    public async Task<ChatCompletionOptions> CreateCompletionOptionsAsync(string prompt, int maxTools = 3, float minVectorDistance = 0.29f)
+    {
+        if (!CanFilterTools)
+            return CreateCompletionOptions();
```
joseharriaga: This feels a little bit like a "silent failure" to me, because this method doesn't do the expected thing unless this instance of `ChatTools` was created with an `EmbeddingClient`.

Considering that the `EmbeddingClient` is not used anywhere else, I think it would be a good improvement to remove the `EmbeddingClient` parameter from the `ChatTools` constructor and add it here as a required parameter instead. This fixes the "silent failure" and also reconciles all the functionality related to embeddings in a single place, which is very nice.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+        {
+            return ToolsUtility.GetClosestEntries(_entries, maxTools, minVectorDistance, vector);
+        }
+    }
+
+    /// <summary>
+    /// Implicitly converts ChatTools to <see cref="ChatCompletionOptions"/>.
+    /// </summary>
+    /// <param name="tools">The ChatTools instance to convert.</param>
+    public static implicit operator ChatCompletionOptions(ChatTools tools) => tools.CreateCompletionOptions();
```
joseharriaga: Personally, I would prefer not to have this implicit conversion operator. In my experience, customers struggle to discover them, and when they do, they struggle to understand how they work. I think the `CreateCompletionOptions` method is good enough.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+
+    internal string CallLocal(ChatToolCall call)
```
joseharriaga: We probably want to add support for awaiting async functions. For example, it's not uncommon to use functions to call web APIs.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+        {
+            ToolsUtility.ParseFunctionCallArgs(call.FunctionArguments, out arguments);
+        }
+        return CallLocal(call.FunctionName, [.. arguments]);
+    }
+
+    private string CallLocal(string name, object[] arguments)
+    {
+        if (!_methods.TryGetValue(name, out MethodInfo method))
+            return $"I don't have a tool called {name}";
```
joseharriaga: Just like in `CallMcpAsync`, this should throw an exception instead of returning a message.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        object result = method.Invoke(null, arguments);
+        return result?.ToString() ?? string.Empty;
```
joseharriaga: Instead of returning `string.Empty` (which is a bit of a "silent failure"), this should probably throw an exception explaining that the result of a function cannot be null and that the user needs to fix their function.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,296 @@
...
+            throw new NotImplementedException($"MCP tool {call.FunctionName} not found.");
+
+#if !NETSTANDARD2_0
+        var actualFunctionName = call.FunctionName.Split(ToolsUtility.McpToolSeparator, 2)[1];
+#else
+        var index = call.FunctionName.IndexOf(ToolsUtility.McpToolSeparator);
+        var actualFunctionName = call.FunctionName.Substring(index + ToolsUtility.McpToolSeparator.Length);
+#endif
+        var result = await method(actualFunctionName, call.FunctionArguments).ConfigureAwait(false);
+        return result.ToString();
```
joseharriaga, line 254: This should probably throw an exception if `result` or `ToString` return `null` explaining that the result of a function cannot be null and the user needs to fix their code.

src/Utility/ToolsUtility.cs
```diff
@@ -0,0 +1,194 @@
...
+        writer.WriteStartObject("properties"u8);
+
+        foreach (ParameterInfo parameter in parameters)
+        {
+            writer.WriteStartObject(parameter.Name!);
+            writer.WriteString("type"u8, ClrToJsonTypeUtf8(parameter.ParameterType));
+            writer.WriteString("description"u8, GetParameterDescription(parameter));
+            writer.WriteEndObject();
+
+            if (!parameter.IsOptional || (parameter.HasDefaultValue && parameter.DefaultValue is not null))
```
joseharriaga, line 68: Curious: Why is an optional parameter with a non-null default considered required?

src/Utility/ChatTools.cs
KrzysztofCwalina: do we really add it before GA? It seems like unneeded hurdle to try new features. Of course, if we GA, we need to do it. Maybe we add the attributes but commented out so we dont forget?

src/Utility/ChatTools.cs
KrzysztofCwalina: Are we sure about that? I heard it's better to send an error message to the model so it can retry

src/Utility/ChatTools.cs
KrzysztofCwalina: or return string "null"

KrzysztofCwalina: Also, dont we support tools that do things without returning anything?

src/Utility/ChatTools.cs
christothes: Can you say more about why you think this is friendlier?

src/Utility/ChatTools.cs
christothes: Do you think it would be common that users would want to customize the embedding options? I'd prefer to leave it out unless you think it is a mainline scenario.

src/Utility/ToolsUtility.cs
christothes: Should we? If so, what would be the motivation to do so?

src/Utility/ChatTools.cs
christothes: I don't think it would work currently. I'll look into this.

src/Utility/ChatTools.cs
christothes: I think this could be a good alternative - @KrzysztofCwalina - thoughts?

src/Utility/ChatTools.cs
christothes: > Also, dont we support tools that do things without returning anything?

Good question I would think that all tools return something, even if it is just to say what it did.

src/Utility/ChatTools.cs
KrzysztofCwalina: I am not sure this will work. The same embeddings client needs to be used for creating vectors for tools and for the prompt. If we accept the client here, we would need to recompute the tools vectors or risk the chance that the caller used different one for tools and different for prompt

src/Utility/ToolsUtility.cs
christothes: Because it effectively is always included with at least a default value.

src/Utility/ToolsUtility.cs
joseharriaga: Structured outputs offer better guarantees that the model will adhere to the JSON schema of the function parameters instead of being "best effort", which can result in the model hallucinating parameters that don't exist, returning malformed JSON, etc.

Here is more info from OpenAI's documentation:
> Structured Outputs is available in two forms in the OpenAI API:
> 1. When using [function calling](https://platform.openai.com/docs/guides/function-calling)
> 2. When using a json_schema response format
>
> Function calling is useful when you are building an application that bridges the models and functionality of your application.
>
> For example, you can give the model access to functions that query a database in order to build an AI assistant that can help users with their orders, or functions that can interact with the UI.
>
>Conversely, Structured Outputs via response_format are more suitable when you want to indicate a structured schema for use when the model responds to the user, rather than when the model calls a tool.
>
> For example, if you are building a math tutoring application, you might want the assistant to respond to your user using a specific JSON Schema so that you can generate a UI that displays different parts of the model's output in distinct ways.
>
> Put simply:
> * If you are connecting the model to tools, functions, data, etc. in your system, then you should use function calling
> * If you want to structure the model's output when it responds to the user, then you should use a structured response_format

🔗 https://platform.openai.com/docs/guides/structured-outputs

And also:

> Strict mode
>
> Setting strict to true will ensure function calls reliably adhere to the function schema, instead of being best effort. We recommend always enabling strict mode.

🔗 https://platform.openai.com/docs/guides/function-calling

Based on this, I expect many users would want to be able to use function calling in strict mode.

src/Utility/ChatTools.cs
joseharriaga: Maybe we can leave it out for now and add it later if customers request it.

src/Utility/ChatTools.cs
joseharriaga: Yes, all tools must return something. If a function has no "useful" return value, the recommendation is to return a string to indicate success or failure (e.g., "success" or "failure"). Ultimately, though, I think the specifics should be up to the user (so that they can tailor the implementation for their specific scenario) instead of us making a decision for them. This is why I think throwing an exception here instead of silently returning `string.Empty` is the correct thing to do.

src/Utility/ChatTools.cs
joseharriaga: Adding the attributes commented out sounds good.

src/Utility/ChatTools.cs
KrzysztofCwalina: the problem is that I am not sure users will know this. If the user wants to have a local function that does not return anything, they will be thinking about it as a void function. But I agree there is a tradeoff here. We would have to return something (empty or "done") which might be too much magic

src/Utility/MCP/McpSession.cs
christothes: will defer this to later.

src/Utility/MCP/McpSession.cs
christothes: will defer this to later.

src/Utility/ResponseTools.cs
```diff
@@ -0,0 +1,282 @@
...
+using OpenAI.Embeddings;
+using OpenAI.Responses;
+
+namespace OpenAI.Responses;
+
+/// <summary>
+/// Provides functionality to manage and execute OpenAI function tools for responses.
+/// </summary>
+//[Experimental("OPENAIMCP001")
+public class CallLocalAsync
```
KrzysztofCwalina: why public?

christothes: That was an accidentla rename. It should be `ResponseTools`

src/Utility/ToolsUtility.cs
```diff
@@ -0,0 +1,253 @@
...
+            Type t when t == typeof(float) => "number"u8,
+            Type t when t == typeof(string) => "string"u8,
+            Type t when t == typeof(bool) => "bool"u8,
+            _ => throw new NotImplementedException()
+        };
+
+    internal static string ClrToJsonTypeUtf16(Type clrType) =>
+        clrType switch
+        {
+            Type t when t == typeof(double) => "number",
```
KrzysztofCwalina: do we need to add long here?

KrzysztofCwalina (approved):

src/Utility/MCP/McpClient.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.ClientModel.Primitives;
+using System.Threading.Tasks;
+
+namespace OpenAI.Agents;
+
+/// <summary>
+/// Client for interacting with a Model Context Protocol (MCP) server.
+/// </summary>
+//[Experimental("OPENAIMCP001")]
+public class McpClient
```
stephentoub: I assume this is just a placeholder, right?

This should be depending on and using the ModelContextProtocol library. We should not be implementing and shipping another MCP client.

KrzysztofCwalina: ModelContextProtocol is pretty large. It has 8 namespaces and tens of public types. Almost all of them for the support of the server scenarios. I filed an issue to refactor ModelContextProtocol into client and server packages. Anyway, there is a tradeoff here; we will reevaluate options before stable release.

stephentoub: >  tens of public types. Almost all of them for the support of the server scenarios

First, that's not the case. The majority of the types are in support of protocol-level concerns that are applicable to both client and server.

But even if it were the case, it is not a reason to avoid using the library. It is _the_ official MCP library. It is the one we're collectively investing in. It is the defacto answer for anyone doing MCP with C# and .NET.

We can cull down the number of namespaces; we already plan to and have an issue to do so. But they're also all subnamespaces of ModelContextProtocol; I don't see why this should have any bearing on whether the library is used.

> I filed an issue to refactor ModelContextProtocol into client and server packages

Yes, and that issue is still open. You also offline cited the lack of it being split as valuable rather than blocking.

> Anyway, there is a tradeoff here; we will reevaluate options before stable release.

We should not be merging this. If the OpenAI library itself needs access to MCP support, it should be using the ModelContextProtocol library to get it.

christothes: I removed McpClient from the public API surface.

src/Utility/ToolsUtility.cs
christothes: added more types - will doc in a follow up PR.

src/Utility/ToolsUtility.cs
christothes: punting this to future.

src/Utility/ResponseTools.cs
```diff
@@ -0,0 +1,282 @@
...
+    {
+#pragma warning disable IL2070
+        foreach (MethodInfo function in tool.GetMethods(BindingFlags.Public | BindingFlags.Static))
+        {
+            AddFunctionTool(function);
+        }
+#pragma warning restore IL2070
+    }
+
+    internal void AddFunctionTool(MethodInfo function)
```
stephentoub, line 82: This is reinventing AIFunctionFactory from Microsoft.Extensions.AI, albeit with fewer features developers need. It'd be better to just use AIFunctionFactory. Static methods alone are insufficient to address real-world need. Developers need support instance methods, both on supplied instances and those manufactured from DI. Developers need to be able to inject state, such as satisfying parameters from dependency injection. Developers need to support flowing cancellation. Developers need support for complex parameters and return types, and with proper NativeAOT support. And so on.

src/Utility/ToolsUtility.cs
```diff
@@ -0,0 +1,243 @@
...
+            Type t when t == typeof(double) => "number"u8,
+            Type t when t == typeof(int) => "number"u8,
+            Type t when t == typeof(long) => "number"u8,
+            Type t when t == typeof(float) => "number"u8,
+            Type t when t == typeof(string) => "string"u8,
+            Type t when t == typeof(bool) => "bool"u8,
+            _ => throw new NotImplementedException()
+        };
+
+    internal static BinaryData BuildParametersJson(ParameterInfo[] parameters)
```
stephentoub, line 49: This should just use Microsoft.Extension.AI.Abstraction's AIJsonUtilities.CreateFunctionJsonSchema, which already has much broader support. Though you wouldn't need that either if you just use AIFunctionFactory.

stephentoub:
I appreciate that the MCP support was made internal instead of public, but that's only part of the concern [I raised](https://github.com/openai/openai-dotnet/pull/422#discussion_r2076788227). We should not be maintaining a separate MCP implementation. Are you going to invest in updating this with streamable HTTP support? Are you going to implement MCP's auth spec? Are you going to add support for sampling? What about OpenTelemetry MCP support, both the relevant spans/tags and metrics but also context propagation? Support for progress updates? Updating it when the next spec is ratified? And so on.

I'll reiterate my previous statement. We should not be shipping a custom MCP implementation. If this library is going to include MCP support, it should do so by depending on the official ModelContextProtocol C# SDK.

src/Utility/ChatTools.cs
```diff
@@ -0,0 +1,287 @@
...
+            AddFunctionTool(functionHolder);
+    }
+
+    /// <summary>
+    /// Adds all public static methods from the specified type as tools.
+    /// </summary>
+    /// <param name="tool">The type containing tool methods.</param>
+    internal void AddFunctionTool(Type tool)
+    {
+#pragma warning disable IL2070
```
stephentoub, line 74: We should not be suppressing trimmer warnings like this. The assembly claims to be AOT compatible, but as written this code isn't.

christothes: We are intentionally defering making this preview feature fully AOT compliant and will address this if it progresses to a GA release.

src/Utility/ChatTools.cs
christothes: defering this scenario for preview.

src/Utility/ResponseTools.cs
KrzysztofCwalina: We would love to use some reusable library for this. But currently, we are not ready take a hard dependency on MEAI. Let's keep discussing how to solve the problem offline (not in this PR)

src/Utility/ToolsUtility.cs
KrzysztofCwalina: We would love to use some reusable library for this. But currently, we are not ready take a hard dependency on MEAI. Let's keep discussing how to solve the problem offline (not in this PR)

KrzysztofCwalina:
> I appreciate that the MCP support was made internal instead of public, but that's only part of the concern [I raised](https://github.com/openai/openai-dotnet/pull/422#discussion_r2076788227). We should not be maintaining a separate MCP implementation. Are you going to invest in updating this with streamable HTTP support? Are you going to implement MCP's auth spec? Are you going to add support for sampling? What about OpenTelemetry MCP support, both the relevant spans/tags and metrics but also context propagation? Support for progress updates? Updating it when the next spec is ratified? And so on.
> 
> I'll reiterate my previous statement. We should not be implementing/maintaining/shipping a custom MCP implementation. If this library is going to include MCP support, it should do so by depending on the official ModelContextProtocol C# SDK. If there's a deficiency in the MCP SDK, we should collaborate to fix it rather than building another one.

We would love to use the official MCP library for this. But currently, we are not ready take a hard dependency on it. It brings MEAI dependency with it. Also, the official MCP library is very large, forces overhead of deserializing to MEAI representation of tools and then we would have to convert to the OPenAI implementation. Let's keep discussing how to solve the problem offline (not in this PR).

stephentoub:
Whether or not those are real concerns, I don't see how it affects a preview release. We should be starting under the assumption it uses the libraries it should be using and go from there.

rwoodley:
It would be more convenient for developers if the OpenAI SDK did in fact rely on MEAI. MEAI doesn't require us to write model-specific code. Right now in my code, in order to get the "The official .NET library for the OpenAI API" to talk to the "official C# SDK for the Model Context Protocol" I have to write ~60 lines of code to convert from OpenAI's ChatToolCall to the argument dictionary expected by the MCP SDK.

KrzysztofCwalina:
> It would be more convenient for developers if the OpenAI SDK did in fact rely on MEAI. MEAI doesn't require us to write model-specific code. Right now in my code, in order to get the "The official .NET library for the OpenAI API" to talk to the "official C# SDK for the Model Context Protocol" I have to write ~60 lines of code to convert from OpenAI's ChatToolCall to the argument dictionary expected by the MCP SDK.

@rwoodley, you can use Microsoft.Extensions.AI.OpenAI package to expose OpenAI ChatClient as IChatClient. If you then use IChatClient for the completion calls, you don't need any conversions while calling IMcpClient

```csharp
ChatClient openai = ..
IChatCLient chat = openai.AsIChatClient();
```

PR #423, opened 2025-05-02
Type-forward `IsExternalInit` on modern .NET.

teo-tsirpanis:
Fixes compatibility between the .NET Standard 2.0 and modern .NET assemblies.

See https://github.com/Sergio0694/PolySharp/issues/98#issuecomment-2183071987 for more details.

### DIFF
--- a/src/Utility/Polyfill/System.Runtime.CompilerServices.IsExternalInit.cs
+++ b/src/Utility/Polyfill/System.Runtime.CompilerServices.IsExternalInit.cs
@@ -1,4 +1,8 @@
-#if !NET5_0_OR_GREATER
+#if NET5_0_OR_GREATER
+using System.Runtime.CompilerServices;
+
+[assembly:TypeForwardedTo(typeof(IsExternalInit))]
+#else
 
 using System.ComponentModel;
 namespace System.Runtime.CompilerServices;


jsquire:
@KrzysztofCwalina, @christothes, @stephentoub:  Would appreciate thoughts on this one.

stephentoub:
@teo-tsirpanis, why? We don't do that in other places, e.g. System.Text.Json has an internal IsExternalInit when built for netstandard2.0 or net462 but doesn't include one when targeting net8.0, net9.0, or net10.0.

teo-tsirpanis:
`System.Text.Json` actually [does type-forward `IsExternalInit`](https://github.com/dotnet/runtime/blob/e1f6fbeb985c040108758cc0b017aa9d1c3d6be5/src/libraries/System.Text.Json/ref/System.Text.Json.Typeforwards.netcoreapp.cs) on modern frameworks. It needs that because it has one public property with an `init;` accessor. The OpenAI library does not have a public `init;` property yet, but once it gets one, it will require type-forwarding `IsExternalInit`. We can do this now or later.

stephentoub:
> actually [does type-forward IsExternalInit](https://github.com/dotnet/runtime/blob/e1f6fbeb985c040108758cc0b017aa9d1c3d6be5/src/libraries/System.Text.Json/ref/System.Text.Json.Typeforwards.netcoreapp.cs)

Thanks, I missed we added that.

In what situation do we expect the netstandard asset to be resolved when an app targets net core?

stephentoub:
Ah, I guess it's the opposite. A netstandard lib is built against this one, and will bake in that reference. Then that netstandard only library gets used on core. OK, this makes sense.

But I'd do it later only if / when it's actually needed; otherwise it's leaking implementation detail.

jsquire:
Thank you for your contribution, @teo-tsirpanis, and for your interest in improving the OpenAI developer experience.

> The OpenAI library does not have a public `init;` property yet, but once it gets one, it will require type-forwarding `IsExternalInit`. We can do this now or later.

We avoid `init` properties as part of the API surface because of these types of corner cases in `netstandard2.0` fallbacks.   If that's the scenario that is motivating this change, I don't expect that it will be needed.

jsquire:
Thank you, @teo-tsirpanis, for your contribution and interest in improving the OpenAI developer experience.  After discussion between the maintainers and confirming that we have no plans to make use of `init` properties due to the `netstandard2.0` compatibility concerns, I'm going to close this out.

PR #426, opened 2025-05-08
Add GenerateSpeechStreamingAsync() for streaming TTS responses.

bmdub:
This PR adds GenerateSpeechStreamingAsync(), which returns the stream of audio chunks from the TTS API.
This differs from the existing GenerateSpeechAsync(), which returns only the entire audio file after it has completed downloading.

Feel free to make it your own.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    //"version": "8.0.100",
     "rollForward": "feature"
   }
 }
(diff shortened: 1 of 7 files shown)


jsquire:
Hi @bmdub.   Thanks very much for your contribution and interest in improving the OpenAI developer experience.    We apologize for the long wait for someone to review.

Unfortunately, we cannot accept this submission, as you are editing files which are automatically generated from the REST API spec and would be overridden.   The good news is that the changes that you're looking to make all appear to have been included in the v2.2.0 release, which used an updated spec.

I'm going to close this out.  Thank you, again, for your efforts.

PR #435, opened 2025-05-20
Preliminary Support for gpt-image-1

benjamin-stern:
Adds Supported Resolutions of 1024x1536 and 1536x1024 
Adds Supported Qualities of High, Medium and Low. 

**Breaking Change**
High has been renamed to Hd reflecting the name of it's value.

Note: Addition of Quality has been added specifically _src/Generated/Models/GeneratedImageSize.cs_

### DIFF
--- a/examples/CombinationExamples.cs
+++ b/examples/CombinationExamples.cs
@@ -21,7 +21,7 @@ public partial class CombinationExamples
             new()
             {
                 Style = GeneratedImageStyle.Vivid,
-                Quality = GeneratedImageQuality.High,
+                Quality = GeneratedImageQuality.Hd,
                 Size = GeneratedImageSize.W1792xH1024,
             });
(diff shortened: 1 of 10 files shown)


benjamin-stern:
@joseharriaga , @trrwilson , @ShivangiReja any updates? Is some alternative in the works. Seems a little disappointing to leave OpenAI Developers in the lurch waiting for a simple extension of support.

jsquire (changes requested):
Hi @benjamin-stern.   Thanks very much for your contribution and interest in improving the OpenAI developer experience.    We apologize for the long wait for someone to review.

Unfortunately, we cannot accept this submission, as you are editing files which are automatically generated from the REST API spec and would be overridden.   The good news is that the changes that you're looking to make all appear to have been included in the v2.2.0 release, which used an updated spec.

I'm going to close this out.  Thank you, again, for your efforts.

PR #437, opened 2025-05-21
Fix bool types

shargon:
https://cswr.github.io/JsonSchema/spec/basic_types/


shargon:
Any feedback?

jsquire:
Confirmed with @joseharriaga that this does not conflict with any upstream work and is safe to merge once tests pass.

jsquire:
Failures are unrelated to these changes and indicate an issue in the live testing environment.   Merging based on units passing.

PR #459, opened 2025-06-18
Return shared ArrayPool

shargon:


src/Custom/Embeddings/OpenAIEmbedding.cs
```diff
@@ -121,6 +121,7 @@ private static ReadOnlyMemory<float> ConvertToVectorOfFloats(BinaryData binaryDa
         OperationStatus status = Base64.DecodeFromUtf8(base64, bytes.AsSpan(), out int bytesConsumed, out int bytesWritten);
         if (status != OperationStatus.Done || bytesWritten % sizeof(float) != 0)
         {
+            ArrayPool<byte>.Shared.Return(bytes);
```
jsquire: Can we please refactor the approach to move this to a `finally` block?  We're likely to run into additional missed corner cases over time and will already leak in the case of an unexpected exception.

jsquire (approved):
Hi @shargon.  Thank you for your contribution and interest in improving the OpenAI developer experience.   Thanks for catching these, as these appear to be legitimate leaks to me.    

@joseharriaga:  This looks to be a clean fix that is safe to make here.   Can you please confirm and, if so, complete the merge?

src/Custom/Embeddings/OpenAIEmbedding.cs
shargon: done!

jsquire (approved):
Thanks, @shargon!    I touched base with @joseharriaga earlier and confirmed that this won't cause issues with in-flight work upstream.   Merging.

src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs
```diff
@@ -26,6 +26,7 @@ public AsyncWebsocketMessageResultEnumerator(WebSocket webSocket, CancellationTo
 
     public ValueTask DisposeAsync()
     {
+        ArrayPool<byte>.Shared.Return(_receiveBuffer);
```
stephentoub, line 29: If disposal is performed multiple times, this will end up returning the buffer multiple times. That needs to be fixed, assuming this enumerator can be returned out to user code.

src/Custom/Embeddings/OpenAIEmbedding.cs
```diff
@@ -118,32 +118,36 @@ private static ReadOnlyMemory<float> ConvertToVectorOfFloats(BinaryData binaryDa
...
             {
                 ints[i] = BinaryPrimitives.ReverseEndianness(ints[i]);
             }
 #endif
+            }
+            return new ReadOnlyMemory<float>(vector);
+        }
+        finally
+        {
+            ArrayPool<byte>.Shared.Return(bytes);
```
stephentoub, line 148: This is OK. FWIW, though, we generally don't consider it a bug to not return an ArrayPool array in exceptional situations.

src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs
jsquire: This one is on me, as I was under the impression that it was safe to return multiple times.   However, [looks like I'm incorrect](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1.return?view=net-9.0#remarks).

stephentoub: This use of ArrayPool looks dangerous in general. The enumerator is yielding references to this array. If the consumer holds on to those yielded elements after disposal and reads from them (which is technically fine for them to do), that's a use-after-free bug.

I think the right answer here is to stop using ArrayPool here, i.e. revert the addition of the Return and change the Rent to just be a normal allocation. That won't be any worse than the state of allocation today because the array was already not being returned. If in the future perf problems demonstrate pooling is important, it can be re-evaluated holistically.

stephentoub: Actually, it may not be as bad as I feared... tracing through where _recieveBuffer goes, it looks like it may not actually be stored into the yielded object but instead just having its data copied out? If so, using an ArrayPool array would be ok. But we still want to ensure it's only returned to the pool once.

jsquire: Filed #474 for investigation.

Issue #470, opened 2025-06-21
Tools in Response API should not be readonly


gaspardpetit:
### Service

OpenAI

### Describe the bug

Thank you for adding samples with the Response API (https://github.com/openai/openai-dotnet/commit/dcf3838639d5a17162c7500aa22dfdeb8b49d1bb). 

I added a few changelist, I believe the documentation is incorrect in a few places.

In particular, I do not think it is currently possible to assign a Tool as described:

```
   new ResponseCreationOptions()
    {
        Tools = { ResponseTool.CreateWebSearchTool() },
    });
```

Since tools is only provided in the code with a getter, and no setter:

`ResponseCreationOptions`:
```
public IList<ResponseTool> Tools { get; }
```

Most likely, the setter should be added.

### Steps to reproduce

Try to compile the README.md sample, in particular:

   new ResponseCreationOptions()
    {
        Tools = { ResponseTool.CreateWebSearchTool() },
    });

Expect the code to compile;

Observe that the following error is generated:

```
Property or indexer 'ResponseCreationOptions.Tools' cannot be assigned to -- it is read only
```

### Code snippets

```C#

```

### OS

Windows

### .NET version

.Net 9.0

### Library version

2.2.0-beta.4

jsquire:
Hi @gaspardpetit.  Thank you for reaching out and we regret that you're experiencing difficulties.  As a convention across types in the library, collection types for models are intentionally read-only.   Their value can be set by mutating the existing collection.  This can be done by either adding members after the object has been constructed or using C#'s initializer syntax as part of construction.   

```csharp
// Example: adding items after construction.
var options = new ResponseCreationOptions();
options.Tools.Add( ResponseTool.CreateWebSearchTool());

// Example: initializer syntax
var options = new ResponseCreationOptions()
{
    Tools = { ResponseTool.CreateWebSearchTool() }
};
```

The snippet that you shared is syntactically correct and makes use of the initializer syntax.

gaspardpetit:
Thank you for the clarification. I now understand that the example works due to Tools being initialized by the default constructor. That said, I find this pattern awkward to work with in practice.

For example, if we already have:

```
List<ResponseTool> tools = GetAvailableTools();
```

We cannot simply pass it to the ResponseCreationOptions. Instead, we're forced to do something like:

```
var options = new ResponseCreationOptions();
foreach (var t in tools) options.Tools.Add(t);
```

or attempt a fragile cast:

```
[...]

PR #471, opened 2025-06-21
Fix to allow setting Tools in the Response API

gaspardpetit:
As discussed here: https://github.com/openai/openai-dotnet/issues/470


jsquire:
@gaspardpetit:  Thank you for your contribution and interest in improving the OpenAI developer experience.   As discussed in #470, the lack of setter is intentional and follows a convention for the library.   The collection is instantiated at construction and can be mutated.   While we appreciate your efforts, we will not be moving forward with these changes.

PR #472, opened 2025-06-21
Add support for filename in ResponseMessageAnnotation

gaspardpetit:


jsquire:
Hi @gaspardpetit.  Thank you for your contribution and interest in improving the OpenAI developer experience.   The content that you're attempting to change is automatically generated from the version of the OpenAI REST API spec that the library was built against.   Unfortunately, generated code cannot be hand-edited, as any changes will be overwritten the next time the generator is run.   When the library is regenerated from a version of the REST API spec with filename support in `ResponseMessageAnnotation` it will be included automatically.

PR #476, opened 2025-06-23
Improve disposal logic in AsyncWebsocketMessageResultEnumerator to prevent multiple disposals

### LINKED ISSUE #474: Evaluate ArrayPool usage in AsyncWebsocketMessageEnumerator
# Summary

The [AsyncWebsocketMessageEnumerator](https://github.com/openai/openai-dotnet/blob/main/src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs) makes use of the shared ArrayPool to rent a buffer that is used during enumeration.    As part of [#459](https://github.com/openai/openai-dotnet/pull/459) concerns were raised that this may be a dangerous choice for implementation and should be reconsidered.

## Scope

- Evaluate the implementation and take thoughts from the PR discussion into account.
- If the implementation needs to be adjusted, make the adjustments and ensure test coverage.

## References and resources

- [Return shared ArrayPool (#459)](https://github.com/openai/openai-dotnet/pull/459/files#r2162095389)

christothes:
closes #474

### DIFF
--- a/src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs
+++ b/src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs
@@ -31,7 +31,7 @@ internal partial class AsyncWebsocketMessageResultEnumerator : IAsyncEnumerator<
             ArrayPool<byte>.Shared.Return(toReturn);
         }
         _webSocket?.Dispose();
-        return new ValueTask(Task.CompletedTask);
+        return default;
     }
 
     public async ValueTask<bool> MoveNextAsync()


src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs
```diff
@@ -26,8 +27,12 @@ public AsyncWebsocketMessageResultEnumerator(WebSocket webSocket, CancellationTo
 
     public ValueTask DisposeAsync()
     {
-        ArrayPool<byte>.Shared.Return(_receiveBuffer);
-        _webSocket?.Dispose();
+        if (!_disposed)
```
KrzysztofCwalina: this is not thread safe.

christothes: Yes, but do we expect a given instance of the enumerator to be run concurrently?

src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs
```diff
@@ -26,8 +27,12 @@ public AsyncWebsocketMessageResultEnumerator(WebSocket webSocket, CancellationTo
...
+        {
+            ArrayPool<byte>.Shared.Return(_receiveBuffer);
+            _webSocket?.Dispose();
+            _disposed = true;
```
stephentoub: Rather than adding another piece of state to track, you could also null out the buffer, and then only return it when it's not null.

src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs
KrzysztofCwalina: probably not in real scenarios, but double return to a pool is such a severe bug. I think we should not be using the pool unless we can guarantee 100% correctness, even in contrived or bug cases or when the user has a bug in their code. Double returns to the pool mess up the whole app domain and can lead issues like data loss.

stephentoub: Wouldn't hurt to use an Interlocked, e.g.
```C#
if (Interlocked.Exchange(ref _receiveBuffer, null) is byte[] toReturn)
{
    ArrayPool<byte>.Shared.Return(toReturn);
}
```

src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs
```diff
@@ -26,7 +26,10 @@ public AsyncWebsocketMessageResultEnumerator(WebSocket webSocket, CancellationTo
 
     public ValueTask DisposeAsync()
     {
-        ArrayPool<byte>.Shared.Return(_receiveBuffer);
+        if (Interlocked.Exchange(ref _receiveBuffer, null) is byte[] toReturn)
+        {
+            ArrayPool<byte>.Shared.Return(toReturn);
+        }
         _webSocket?.Dispose();
         return new ValueTask(Task.CompletedTask);
```
stephentoub: Pre-existing, but this last line could also just be:
```suggestion
        return default;
```
What's there is fine, though.

stephentoub (approved):
Thanks

src/Custom/RealtimeConversation/Internal/AsyncWebsocketMessageEnumerator.cs
KrzysztofCwalina: @stephentoub, do we need to make the filed volatile so that there is no use after free?

stephentoub: Making it volatile won't really help if there's (erroneous) concurrent use of other members of the enumerator, as regardless of whether it's volatile or not they could have already grabbed a snapshot of the field's value. And non-concurrent use doesn't need a fence. Interlocked.Exchange itself is also a full fence.

KrzysztofCwalina: true

jsquire (approved):

KrzysztofCwalina (approved):

PR #477, opened 2025-06-23
Updating issue templates and configuration

jsquire:
# Summary

Updating the issue templates to:

  - Only add the `needs-triage` label to all submissions
  - Add a format slug to the title
  - Add a "question" type issue
  - Enable blank issues


jsquire:
Not impactful to the product; I'm going to merge directly without waiting on review so that we can fix the "what the heck needs to be triaged" problem.

PR #478, opened 2025-06-24
[EngSys] Update .NET SDK version

jsquire:
# Summary

Moving the repository version of the .NET SDK to the latest.


jsquire:
Closing out to move to a non-browser change.  Looks like the workflows are hardcoded and need to be updated in parallel.

PR #479, opened 2025-06-24
[EngSys] Update .NET SDK version

jsquire:
# Summary

Moving the repository version of the .NET SDK to the latest.


tests/OpenAI.Tests.csproj
```diff
@@ -9,7 +9,7 @@
     <NoWarn>$(NoWarn);OPENAI001;</NoWarn>
 
     <LangVersion>latest</LangVersion>
-  </PropertyGroup>
+  </PropertyGroup>
```
jsquire:
```suggestion
  </PropertyGroup> 
```

tests/OpenAI.Tests.csproj
```diff
@@ -9,7 +9,7 @@
...
```
jsquire:
```suggestion
  </PropertyGroup>
```

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    "version": "9.0.205",
```
Frulfump, line 3: 9.0.2xx SDKs are out of support since May https://learn.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs#supported-net-versions

Could consider Dependabot to handle updating the version in the future.

PR #483, opened 2025-06-25
Create dotnet-desktop.yml

solveforceapp:


jsquire (changes requested):
Hi @solveforceapp.  Thank you for your contribution, but this is not a submission that we'll be moving forward with.

PR #490, opened 2025-06-26
mcp usage sample

KrzysztofCwalina:


docs/guides/mcp/chat_with_mcp.cs
```diff
@@ -0,0 +1,89 @@
+// SAMPLE: call the official MCP library for .NET
```
jsquire, line 1: We should probably leave some kind of comment detailing what is needed to execute this, since C#-as-a-script files are not widely used.

KrzysztofCwalina: there is a readme in the parent folder describing it

jsquire (approved):

PR #507, opened 2025-07-03
Fix generation of `StreamingResponseTextAnnotationAddedUpdate`

joseharriaga:
Fixes an error in code generation that caused the `StreamingResponseTextAnnotationAddedUpdate` class to not be generated correctly as part of the set of possible handles for streaming Responses.


api/OpenAI.net8.0.cs
```diff
@@ -5499,7 +5499,20 @@ public class StreamingResponseRefusalDoneUpdate : StreamingResponseUpdate, IJson
         protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
     }
     [Experimental("OPENAI001")]
-    public class StreamingResponseTextAnnotationAddedUpdate {
+    public class StreamingResponseTextAnnotationAddedUpdate : StreamingResponseUpdate, IJsonModel<StreamingResponseTextAnnotationAddedUpdate>, IPersistableModel<StreamingResponseTextAnnotationAddedUpdate> {
+        public BinaryData Annotation { get; }
```
gaspardpetit, line 5503: Any chance this could be parsed as an `ResponseMessageAnnotation` ?

api/OpenAI.netstandard2.0.cs
```diff
@@ -4326,7 +4326,16 @@ public class StreamingResponseRefusalDoneUpdate : StreamingResponseUpdate, IJson
         protected override StreamingResponseUpdate PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
         protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
     }
-    public class StreamingResponseTextAnnotationAddedUpdate {
+    public class StreamingResponseTextAnnotationAddedUpdate : StreamingResponseUpdate, IJsonModel<StreamingResponseTextAnnotationAddedUpdate>, IPersistableModel<StreamingResponseTextAnnotationAddedUpdate> {
+        public BinaryData Annotation { get; }
```
gaspardpetit, line 4330: Same here, ideally this would be parsed as a `ResponseMessageAnnotation`

KrzysztofCwalina (approved):

api/OpenAI.net8.0.cs
joseharriaga: Yes, I will make this change in a separate PR.

api/OpenAI.netstandard2.0.cs
joseharriaga: Yes, I will make this change in a separate PR.

ericstj:
It looks like this change removed the public constructor on `StreamingResponseTextAnnotationAddedUpdate`  -- was that intentional?

joseharriaga:
> It looks like this change removed the public constructor on `StreamingResponseTextAnnotationAddedUpdate` -- was that intentional?

Yes, our output classes don't have public constructors by design. Is removing the constructor from `StreamingResponseTextAnnotationAddedUpdate` causing problems for you?

ericstj:
No, another internal breaking change broke my use of Azure.OpenAI, so I ran APICompat to see what else was broken and noticed this public breaking change.

jsquire:
Worth mentioning we only consider this an acceptable change because it's on an experimental surface.   We maintain strong back-compat guarantees on stable surfaces, but experimental should be expected to potentially have breaks between releases.

PR #527, opened 2025-07-10
Improve disposal logic in AsyncWebsocketMessageResultEnumerator to prevent multiple disposals

jsquire:
# Summary

The focus of this change is to reapply the logic from #476 that got inadvertently stomped during the last upstream sync.

### DIFF
--- a/src/Custom/Realtime/Internal/AsyncWebsocketMessageEnumerator.cs
+++ b/src/Custom/Realtime/Internal/AsyncWebsocketMessageEnumerator.cs
@@ -13,7 +13,7 @@ internal partial class AsyncWebsocketMessageResultEnumerator : IAsyncEnumerator<
     public ClientResult Current { get; private set; }
     private readonly CancellationToken _cancellationToken;
     private readonly WebSocket _webSocket;
-    private readonly byte[] _receiveBuffer;
+    private byte[] _receiveBuffer;
 
     public AsyncWebsocketMessageResultEnumerator(WebSocket webSocket, CancellationToken cancellationToken)
     {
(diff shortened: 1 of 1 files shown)


jsquire:
//cc: @christothes

src/Custom/Realtime/Internal/AsyncWebsocketMessageEnumerator.cs
```diff
@@ -26,12 +26,12 @@ public AsyncWebsocketMessageResultEnumerator(WebSocket webSocket, CancellationTo
...
     {
-        _webSocket?.Dispose();
-        if (_receiveBuffer is not null)
+        if (Interlocked.Exchange(ref _receiveBuffer, null) is byte[] toReturn)
         {
-            ArrayPool<byte>.Shared.Return(_receiveBuffer);
+            ArrayPool<byte>.Shared.Return(toReturn);
         }
-        return new ValueTask(Task.CompletedTask);
+        _webSocket?.Dispose();
```
jsquire:
```suggestion

```

jsquire: Removing disposal in the enumerator, because [RealtimeSession.Dispose](https://github.com/openai/openai-dotnet/blob/main/src/Custom/Realtime/RealtimeSession.cs#L389) also manages the websocket's lifespan.   

Transferring ownership to the enumerator invalidates the session and blocks usage scenarios as discussed in #244.

src/Custom/Realtime/Internal/AsyncWebsocketMessageEnumerator.cs
```diff
@@ -26,12 +26,11 @@ public AsyncWebsocketMessageResultEnumerator(WebSocket webSocket, CancellationTo
 
     public ValueTask DisposeAsync()
     {
-        _webSocket?.Dispose();
```
jsquire, line 29: Removed disposal of the socket here, as its lifespan is owned by `RealtimeSession` and managed in its [Dispose](https://github.com/openai/openai-dotnet/blob/main/src/Custom/Realtime/RealtimeSession.cs#L389).    Transferring ownership to the enumerator invalidates the session and blocks further use after observing any set of results, as detailed in #244.

Issue #513, opened 2025-07-03
Parsing streaming response silently fails when using "whisper" for transcription


stephentoub:
The latest release supports streaming audio transcription. Unfortunately the default model that folks have used ("whisper") doesn't work with streaming, but it doesn't just fail; OpenAI's docs highlight that if you use whisper, the stream parameter is simply ignored:
https://platform.openai.com/docs/api-reference/audio/createTranscription#audio-createtranscription-stream
>Note: Streaming is not supported for the whisper-1 model and will be ignored.

This means that when the .NET library goes to parse the response, it ends up not yielding anything out of the enumerable.

To me, the problem here is the service's silent ignoring of `stream`. But we might want to fix the client to recognize when it's getting back a non-streaming response and surface the data using the stream object model, or throw an exception indicating what's happening, rather than just silently not returning anything (would have saved me some time debugging figuring out what was going wrong).

jsquire:
@KrzysztofCwalina, @achandmsft :   Definitely sounds like service gap.   Thoughts on whether we should push for an upstream fix or guard in the client until that is addressed?

KrzysztofCwalina:
@jsquire, I think we should do both. If we patch it in the library, we should log an issue to remove once the service is fixed.

jsquire:
@ShivangiReja:  Can you please take this one?

PR #537, opened 2025-07-15
Throw exception when streaming is used with unsupported whisper-1 model

ShivangiReja:
Fixes: https://github.com/openai/openai-dotnet/issues/513

This PR adds a check to all `TranscribeAudioStreaming` methods to prevent usage with the `whisper-1` model, which does not support streaming.

According to [OpenAI's documentation](https://platform.openai.com/docs/api-reference/audio/createTranscription#audio-createtranscription-stream), the `stream` parameter is ignored when `whisper-1` is used, resulting in confusing behavior for users expecting streaming updates.

### DIFF
--- a/src/Custom/Audio/AudioClient.cs
+++ b/src/Custom/Audio/AudioClient.cs
@@ -229,10 +229,7 @@ public partial class AudioClient
         Argument.AssertNotNull(audio, nameof(audio));
         Argument.AssertNotNullOrEmpty(audioFilename, nameof(audioFilename));
 
-        if (string.Equals(_model, "whisper-1", StringComparison.OrdinalIgnoreCase))
-        {
-            throw new NotSupportedException($"The selected model {_model} does not support streaming transcription. Please use a compatible model.");
-        }
+        EnsureModelSupportsStreaming();
 
         MultiPartFormDataBinaryContent content
             = CreatePerCallTranscriptionOptions(options, stream: true)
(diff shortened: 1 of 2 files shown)


src/Custom/Audio/AudioClient.cs
```diff
@@ -229,6 +229,11 @@ public virtual AsyncCollectionResult<StreamingAudioTranscriptionUpdate> Transcri
         Argument.AssertNotNull(audio, nameof(audio));
         Argument.AssertNotNullOrEmpty(audioFilename, nameof(audioFilename));
 
+        if (string.Equals(_model, "whisper-1", StringComparison.OrdinalIgnoreCase))
```
jsquire: Should we consider extracting that into a method so that the message and model aren't hard-coded in a few places?

jsquire: We may also want to have this governed by an environment variable so that customers can opt-out if `whisper-1` ever does support it.

jsquire (approved):

src/Custom/Audio/AudioClient.cs
ShivangiReja: Thanks for the suggestions! I've updated the code to extract the check into a private method.

ShivangiReja: Thanks! Added support for the environment variable to allow opt-out if needed in the future.

Issue #542, opened 2025-07-16
[BUG] `MessageImageDetail.Auto` fails for `assistantClient.CreateMessageAsync` in version 2.2.0


moshe-ritterman:
### Describe the bug

I just upgraded from version 2.1.0 to 2.2.0 and this code (see below) started to fail with this error message
```
HTTP 400 (invalid_request_error: invalid_value)
Parameter: content[1].image_file.detail

Invalid value: 'Auto'. Supported values are: 'low', 'high', and 'auto'.
```
This is the code that fails
```csharp
await _assistantClient.CreateMessageAsync(threadId, MessageRole.User, [
        "<Some instructions>",
        MessageContent.FromImageFileId(fileId, MessageImageDetail.Auto),
        """
        Additional context: <Some more instructions>
        """
    ]
);
```

### Steps to reproduce

1. Create a chat assistant (using ` AssistantClient`)
2. Create a vector store (using `VectorStoreClient`)
3. Upload 2 files (using `OpenAIFileClient`)
4. Adding one of the files to the vector store (using `VectorStoreClient`)
5. Creating a thread (using `AssistantClient`) and attaching the vector store to the thread
6. Creating a message (using `AssistantClient`) passing in the thread id and the other file uploaded via `OpenAIFileClient`

It's at this step (where I attach the file) that I indicate that the image detail should be set to 'auto' via this code `MessageContent.FromImageFileId(fileId, MessageImageDetail.Auto)` and this is what fails (because 'Auto' is capitalized)

### Code snippets

```C#

```

### OS

macOS

### .NET version

8

### Library version

2.2.0

moshe-ritterman:
p.s. I doubt that this issue is limited to `MessageImageDetail.Auto` it's probably the same with `MessageImageDetail.High` and `MessageImageDetail.Low` since the issue is with casing

jsquire:
Hi @moshe-ritterman.   Thanks for reaching out and we regret that you're experiencing difficulties.    I can confirm this a serialization bug in the client where the enumeration is not being properly lower-cased.   We'll look at getting this fixed up.

In the meantime, you can work around this by creating the expected payload and deserializing, which would look something like:

```csharp
var fileId = "<< YOUR FILE ID >>";
var fileMessage = ModelReaderWriter.Read<MessageContent>(BinaryData.FromObjectAsJson(new
{
    type = "image_file",
    image_file = new
    {
        file_id = fileId,
        detail = "auto"
    }
[...]

PR #544, opened 2025-07-17
Fix casing of MessageImageDetail in MessageContent factory methods

ShivangiReja:
Lowercased `MessageImageDetail` enum values in all `MessageContent` factory methods to match API requirements and fix a 400 error. Also, added a test to verify.

Fixes: https://github.com/openai/openai-dotnet/issues/542

### DIFF
--- a/src/Custom/Assistants/MessageContent.cs
+++ b/src/Custom/Assistants/MessageContent.cs
@@ -19,7 +19,7 @@ public partial class MessageContent
             => new InternalMessageContentImageFileObject(
                 imageFile: new(
                     fileId: imageFileId,
-                    detail: detail?.ToString(),
+                    detail: detail?.ToString().ToLowerInvariant(),
                     additionalBinaryDataProperties: null));
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Custom/Assistants/MessageContent.cs
```diff
@@ -19,7 +19,7 @@ public static MessageContent FromImageFileId(
             => new InternalMessageContentImageFileObject(
                 imageFile: new(
                     fileId: imageFileId,
-                    detail: detail?.ToString(),
+                    detail: detail?.ToString().ToLowerInvariant(),
```
joseharriaga: Instead of calling `ToString().ToLowerInvariant`, it would be more appropriate to call the `ToSerialString()` extension method, so that this works regardless of renames, snake casing, or any other transformations (of which fortunately there are none today in this particular case... but just in case this changes in the future 🙂).

ShivangiReja: Thanks for pointing that out. I wasn’t aware of the `ToSerialString()` extension method. I've updated the code to use it.

joseharriaga (approved):

PR #545, opened 2025-07-18
Enable semantic voice activity detection option

ShivangiReja:
Fixes: https://github.com/openai/openai-dotnet/discussions/543

The method `TurnDetectionOptions.CreateSemanticVoiceActivityTurnDetectionOptions` was originally marked as `internal` to simplify the previous release and allow for iterative development.

This PR updates the method to `public`, making it accessible to users who want to configure conversation sessions using semantic VAD (Voice Activity Detection).


src/Custom/Realtime/GeneratorStubs.cs
```diff
@@ -11,3 +11,4 @@ namespace OpenAI.Realtime;
 [CodeGenType("RealtimeResponseUsage")] public partial class ConversationTokenUsage { }
 [CodeGenType("RealtimeToolType")] public readonly partial struct ConversationToolKind { }
 [CodeGenType("DotNetRealtimeVoiceIds")] public readonly partial struct ConversationVoice { }
+[CodeGenType("RealtimeSemanticVadTurnDetectionEagerness")] public readonly partial struct SemanticEagernessLevel { }
```
jsquire, line 14: I wonder if we can use a visitor to generate these rather than having to maintain stubs here.

src/Custom/Realtime/TurnDetectionOptions.cs
```diff
@@ -35,8 +35,8 @@ public static TurnDetectionOptions CreateServerVoiceActivityTurnDetectionOptions
         };
     }
 
-    internal static TurnDetectionOptions CreateSemanticVoiceActivityTurnDetectionOptions(
-        InternalRealtimeSemanticVadTurnDetectionEagerness? eagernessLevel = null,
+    public static TurnDetectionOptions CreateSemanticVoiceActivityTurnDetectionOptions(
```
jsquire, line 38: Should this migrate from an the `Internal` folder since the class and implementation are `public`?

src/Generated/Models/Realtime/SemanticEagernessLevel.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using System;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+using OpenAI;
+
+namespace OpenAI.Realtime
+{
+    [Experimental("OPENAI002")]
+    public readonly partial struct SemanticEagernessLevel : IEquatable<SemanticEagernessLevel>
```
jsquire, line 13: Yay!  Extensibile!

src/Custom/Realtime/GeneratorStubs.cs
ShivangiReja: I totally agree with this. We’ve been maintaining separate generator stubs per client, and switching to a visitor would make it much easier to manage these going forward. I’ve created an issue to track this: https://github.com/openai/openai-dotnet/issues/550, and we’ll look into addressing it as part of that. Thanks for the suggestion!

src/Custom/Realtime/TurnDetectionOptions.cs
ShivangiReja: It’s already outside the `Internal` folder.

jsquire: my bad.   The indent level in the UI didn't make that clear.   On double-check, it certainly is.

jsquire (approved):

PR #549, opened 2025-07-21
Generate ListChatCompletions with pagination

christothes:


codegen/generator/src/Visitors/MetadataQueryParamVisitor.cs
```diff
@@ -0,0 +1,130 @@
...
+                    }
+
+                    if (uri is not null &&
+                        statement is ForEachStatement foreachStatement &&
+                        foreachStatement.Enumerable is DictionaryExpression dictionaryExpression &&
+                        dictionaryExpression.Original is VariableExpression variable &&
+                        variable.Declaration.RequestedName == "metadata")
+                    {
+                        var formatString = new FormattableStringExpression("metadata[{0}]", [foreachStatement.ItemVariable.Property("Key")]);
+                        var appendQueryStatement = uri.Invoke("AppendQuery", [formatString, foreachStatement.ItemVariable.Property("Value"), new KeywordExpression("true", null)]);
```
jorgerangel-msft:
```suggestion
                        var appendQueryStatement = uri.Invoke("AppendQuery", [formatString, foreachStatement.ItemVariable.Property("Value"), Snippet.True]);
```
would this work here?

codegen/generator/src/Visitors/MetadataQueryParamVisitor.cs
```diff
@@ -0,0 +1,130 @@
...
+                        foreachStatement.Body.Clear();
+                        foreachStatement.Body.Add(new SingleLineCommentStatement("Plugin customization: Properly handle metadata query parameters"));
+                        foreachStatement.Body.Add(new ExpressionStatement(appendQueryStatement));
```
jorgerangel-msft: This can simply be
```suggestion
                        foreachStatement.Body.Add(appendQueryStatement.Terminate());
```
which appends a semicolon and does the same thing under the hood.

codegen/generator/src/Visitors/MetadataQueryParamVisitor.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// with:
+    /// <code>
+    /// foreach (var @param in metadata)
+    /// {
+    ///    uri.AppendQuery($"metadata[{@param.Key}]", @param.Value, true);
+    /// }
+    /// </summary>
+    /// <param name="method"></param>
+    /// <returns></returns>
+    protected override MethodProvider? VisitMethod(MethodProvider method)
```
jorgerangel-msft, line 44: nit: if you override `protected override ScmMethodProvider? VisitMethod(ScmMethodProvider method)` instead, you get a return type of `ScmMethodProvider` with an additional API `IsProtocolMethod`, and it might help with your check below. This implementation is called in our base SCM generator, which should be also called in this generator. @JoshLove-msft - I'm wondering if we should just go ahead and extend the APIs and add one for `IsCreateRequestMethod` ? I'm having to do a similar check in the visitor I'm working on for the MatchConditions header.

codegen/generator/src/Visitors/PaginationVisitor.cs
```diff
@@ -0,0 +1,267 @@
...
+                        newParameters.RemoveAt(i);
+                        lastRemovedIndex = i;
+                        i--;
+                    }
+                }
+                if (lastRemovedIndex >= 0)
+                {
+                    newParameters.Insert(
+                        lastRemovedIndex,
+                        new ParameterProvider("options", $"The pagination options", optionsType.Type, defaultValue: new KeywordExpression("default", null)));
```
jorgerangel-msft:
```suggestion
                        new ParameterProvider("options", $"The pagination options", optionsType.Type, defaultValue: Snippet.Default));
```

codegen/generator/src/Visitors/PaginationVisitor.cs
```diff
@@ -0,0 +1,267 @@
...
+
+                    var newSignature = new MethodSignature(
+                    methodSignature.Name,
```
jorgerangel-msft: minor nit: could we adjust the indentation?

codegen/generator/src/Visitors/PaginationVisitor.cs
```diff
@@ -0,0 +1,267 @@
...
+                                        newInstance.Type,
+                                        newParameters);
+
+                                    var newKeywordExpression = new KeywordExpression(
+                                        keyword.Keyword,
+                                        newNewInstanceExpression);
+
+                                    var newExpressionStatement = new ExpressionStatement(newKeywordExpression);
+
+                                    return newExpressionStatement;
```
jorgerangel-msft: Since we know they keyword expression is "return" could we simplify this to ?
```suggestion
                                    return Snippet.Return(Snippet.New(newInstance.Type, newParameters));
```

codegen/generator/src/Visitors/PaginationVisitor.cs
```diff
@@ -0,0 +1,267 @@
...
+                                if (binaryExpr.Left is VariableExpression leftVar &&
+                                    leftVar.Declaration.RequestedName == "nextToken" &&
+                                    binaryExpr.Right is KeywordExpression rightKeyword &&
+                                    rightKeyword.Keyword == "null")
+                                {
+                                    // Create "hasMore == null" condition
+                                    var hasMoreNullCheck = new BinaryOperatorExpression(
+                                        "==",
+                                        new MemberExpression(null, "hasMore"),
+                                        new KeywordExpression("false", null));
```
jorgerangel-msft: nit:
```suggestion
                                        Snippet.False);
```

codegen/generator/src/Visitors/PaginationVisitor.cs
```diff
@@ -0,0 +1,267 @@
...
+                                assignmentExpression.Value is MemberExpression memberExpression &&
+                                memberExpression.MemberName == "LastId")
+                            {
+                                // Create a new assignment for hasMore
+                                var hasMoreAssignment = new AssignmentExpression(
+                                    new DeclarationExpression(typeof(bool), "hasMore"),
+                                    new MemberExpression(memberExpression.Inner, "HasMore"));
+
+                                // Insert the new assignment before the existing one
+                                statementList.Insert(i, new ExpressionStatement(hasMoreAssignment));
```
jorgerangel-msft:
```suggestion
                                statementList.Insert(i, hasMoreAssignment.Terminate());
```

codegen/generator/src/Visitors/PaginationVisitor.cs
christothes: It looks like Snippet.New is not available as a method (Non-invocable member 'Snippet.New' cannot be used like a method.)

jorgerangel-msft: sorry, that's my fault. Can you try `Snippet.New.Instance` ?

christothes: That works, thanks!

jorgerangel-msft (approved):
visitor updates LGTM

src/Custom/Chat/Internal/GeneratorStubs.cs
```diff
@@ -120,4 +120,6 @@ internal partial class InternalUnknownChatCompletionRequestMessageContentPart {
 [CodeGenType("ChatCompletionDeletedObject")] internal readonly partial struct InternalChatCompletionDeletedObject {}
 [CodeGenType("ChatCompletionMessageListObject")] internal readonly partial struct InternalChatCompletionMessageListObject {}
 [CodeGenType("ChatCompletionList")] internal partial class InternalChatCompletionList {}
-[CodeGenType("ChatCompletionMessageList")] internal partial class InternalChatCompletionMessageList {}
\ No newline at end of file
+[CodeGenType("ChatCompletionMessageList")] internal partial class InternalChatCompletionMessageList {}
+[CodeGenType("ChatCompletionCollectionOptions")] public partial class ChatCompletionCollectionOptions {}
+[CodeGenType("ChatCompletionCollectionOrder")] public readonly partial struct ChatCompletionCollectionOrder { }
```
joseharriaga: nit: Since these are public, they shouldn't be under `Custom/Chat/Internal/GeneratorStubs.cs`. For public types, we typically give them their own file like this:
🔗 https://github.com/openai/openai-dotnet/blob/main/src/Custom/Chat/ChatInputAudioFormat.cs

src/Generated/ChatClient.cs
```diff
@@ -20,6 +23,58 @@ protected ChatClient()
 
         public ClientPipeline Pipeline { get; }
 
+        [Experimental("OPENAI001")]
+        public virtual CollectionResult GetChatCompletions(string after, int? limit, string order, IDictionary<string, string> metadata, string model, RequestOptions options)
```
joseharriaga, line 27: Could you re-run the Export-Api.ps1 script to re-generate the API listing?

specification/client/models/common.models.tsp
```diff
@@ -124,3 +126,22 @@ union DotNetChatVoiceIds {
...
 }
+
+// CollectionQueryParameters
+
+alias CollectionLimitQueryParameter = {
+  /**
+   * A limit on the number of objects to be returned. Limit can range between 1 and 100, and the
+   * default is 20.
+   */
+  @query pageSizeLimit?: int32 = 20;
```
joseharriaga, line 137: Should this name match the REST API: `pageSizeLimit` vs. `limit` ?

specification/client/models/common.models.tsp
```diff
@@ -124,3 +126,22 @@ union DotNetChatVoiceIds {
...
+};
+
+alias CollectionAfterQueryParameter = {
+  /**
+   * A cursor for use in pagination. `after` is an object ID that defines your place in the list.
+   * For instance, if you make a list request and receive 100 objects, ending with obj_foo, your
+   * subsequent call can include after=obj_foo in order to fetch the next page of the list.
+   */
+  @query @continuationToken afterId?: string;
```
joseharriaga, line 146: Should this name match the REST API: `afterId` vs. `after` ?

specification/main.tsp
```diff
@@ -16,5 +16,6 @@ import "./client/models/audio.models.tsp";
 import "./client/models/common.models.tsp";
 import "./client/models/responses.models.tsp";
 import "./client/models/vector-stores.models.tsp";
+import "./client/models/chat.models.tsp";
```
joseharriaga: nit: Could you sort it alphabetically?

src/Generated/Models/Chat/ChatCompletionCollectionOrder.cs
```diff
@@ -0,0 +1,46 @@
...
+        public ChatCompletionCollectionOrder(string value)
+        {
+            Argument.AssertNotNull(value, nameof(value));
+
+            _value = value;
+        }
+
+        public static ChatCompletionCollectionOrder Asc { get; } = new ChatCompletionCollectionOrder(AscValue);
+
+        public static ChatCompletionCollectionOrder Desc { get; } = new ChatCompletionCollectionOrder(DescValue);
```
joseharriaga: nit: Could you rename these to `Ascending` and `Descending` ?

src/Generated/ChatClientGetChatCompletionsAsyncCollectionResult.cs
```diff
@@ -0,0 +1,67 @@
...
+#nullable disable
+
+using System;
+using System.ClientModel;
+using System.ClientModel.Primitives;
+using System.Collections.Generic;
+
+namespace OpenAI.Chat
+{
+    internal partial class ChatClientGetChatCompletionsAsyncCollectionResult : AsyncCollectionResult
```
joseharriaga, line 12: @jorgerangel-msft @JoshLove-msft: What do you think about generating these internal pagination classes under `Generated/Internal` instead of `Generated` ?

JoshLove-msft: Internal is meant for internal helper types that are consistent across libraries. They aren't meant for library specific types that happen to be internal. We can consider renaming this folder, but I wouldn't put these collection types in the same folder as the Internal helper types.

codegen/generator/src/Visitors/MetadataQueryParamVisitor.cs
JoshLove-msft: We could add a new enum that represents the method kind, Protocol, Convenience, CreateRequest

JoshLove-msft: Basically we can make this public - https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs#L777-L782

jorgerangel-msft: Opened https://github.com/microsoft/typespec/issues/8037 for tracking. @christothes - I think this can just be considered a nice to have for now,  and is non-blocking for this pr, given that it would require another generator version bump.

specification/client/models/common.models.tsp
christothes: This custom model doesn't drive generation of the REST client, so I've named the property the way we want it for the client model. Does that address your concern?

specification/client/models/common.models.tsp
christothes: same as above

src/Generated/ChatClient.cs
christothes: done!

codegen/generator/src/Visitors/MetadataQueryParamVisitor.cs
```diff
@@ -0,0 +1,130 @@
...
+using Microsoft.TypeSpec.Generator.Snippets;
+using Microsoft.TypeSpec.Generator.Statements;
+using static OpenAILibraryPlugin.Visitors.VisitorHelpers;
+
+namespace OpenAILibraryPlugin.Visitors;
+
+/// <summary>
+/// This visitor modifies GetRawPagesAsync methods to consider HasMore in addition to LastId when deciding whether to continue pagination.
+/// It also replaces specific parameters with an options type for pagination methods.
+/// </summary>
```
joseharriaga, line 16: nit: This comment corresponds to a different visitor.

tests/Chat/ChatTests.cs
```diff
@@ -1061,6 +1068,392 @@ public async Task WebSearchWorks()
...
+            Model = "gpt-4o-mini-2024-07-18", // Common test model
+            PageSizeLimit = 10
+        };
+
+        int count = 0;
+        await foreach (var fetchedCompletion in client.GetChatCompletionsAsync(filterOptions))
+        {
+            count++;
+            Assert.That(fetchedCompletion.Id, Is.Not.Null.And.Not.Empty);
+            // Note: The actual model filtering behavior depends on the service implementation
```
joseharriaga: nit: What do you mean by this note? Is it not behaving as expected?

tests/Chat/ChatTests.cs
```diff
@@ -1061,6 +1068,392 @@ public async Task WebSearchWorks()
...
+        {
+            Model = "gpt-4o-mini-2024-07-18", // Common test model
+            PageSizeLimit = 10
+        };
+
+        int count = 0;
+        await foreach (var fetchedCompletion in client.GetChatCompletionsAsync(filterOptions))
+        {
+            count++;
+            Assert.That(fetchedCompletion.Id, Is.Not.Null.And.Not.Empty);
```
joseharriaga, line 1354: Could you add an assert to confirm that the `Model` property of the `fetchedCompletion` is equal to the `Model` that was specified in the `ChatCompletionCollectionOptions`? I was going to recommend the same for the metadata filtering test but realized that we don't yet expose the `Metadata` property in `ChatCompletion`.

codegen/generator/src/Visitors/MetadataQueryParamVisitor.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        {
+            "GetChatCompletions",
+            ("ChatCompletion", "ChatCompletionCollectionOptions", _chatParamsToReplace)
+        },
+        {
+            "GetChatCompletionsAsync",
+            ("ChatCompletion", "ChatCompletionCollectionOptions", _chatParamsToReplace)
+        }
+    };
```
joseharriaga: These variables appear to be unused in this class.

joseharriaga (approved):

codegen/generator/src/Visitors/MetadataQueryParamVisitor.cs
christothes: thanks - this was accidentally left in from a copy paste.

tests/Chat/ChatTests.cs
christothes: This comment wasn't supposed to be here. Removing.

tests/Chat/ChatTests.cs
christothes: done!

PR #563, opened 2025-07-25
Added Chat tool calls sample

KrzysztofCwalina:


docs/guides/text/chat/chat_tools.cs
```diff
@@ -0,0 +1,68 @@
...
+                {
+                    string result = MyTools.GetCurrentTime();
+                    messages.Add(ChatMessage.CreateToolMessage(toolCall.Id, result));
+                }
+                else
+                {
+                    Console.WriteLine($"Unknown tool call: {toolCall.FunctionName}");
+                }
+            }
+            goto complete;
```
jsquire: do we really need the `goto` here?   If this were just a standard `break`,  it'll naturally fall out of the case at this point and hit the bottom of the `while`, which bounces it right back up naturally.

KrzysztofCwalina: yeah, in this particular case it would work. But this is not how real e2e program would be structured. For example, there would be a way to get additional prompt from the user inside the loop, but then tool call results cannot go to the top of the while loop and ask the user for a new prompt. 

I noticed that in the mcp sample I did not have the loop at all, and then it kind of makes more sense. Alternatively I could turn this whole loop into a more complete program that asks the user for input (Console.ReadLine). Do you prefer any of these ways?

jsquire: If the "more complete, collect input" approach doesn't add a ton of complexity and make it harder to follow, that sounds great.   If not, then maybe eliding the loop altogether would be a better demonstration?

KrzysztofCwalina: OK. I will eliminate the loop. the interaction with the console/user kind of obscures the main points of the sample

jsquire (approved):

PR #564, opened 2025-07-25
Added .Model properties to clients

KrzysztofCwalina:


KrzysztofCwalina:
@joseharriaga, as we discussed I removed the property from the responses client. Also, I did not make the properties nullable, as we don't have nullability enabled int he repo. Once we enable it, we will mark the properties as nullable

PR #565, opened 2025-07-28
Add constructor with `AuthenticationPolicy` to `ChatClient`

joseharriaga:
System.ClientModel's new `AuthenticationPolicy` makes it possible to use custom authentication mechanisms. For example, using Azure.Identity's `DefaultAzureCredential`, we can define a `BearerTokenPolicy` and use it to communicate with Azure OpenAI like this:

```csharp
DefaultAzureCredential credential = new();
string scope = "https://cognitiveservices.azure.com/.default";
BearerTokenPolicy bearerTokenPolicy = new(credential, scope);

OpenAIClientOptions clientOptions = new()
{
    Endpoint = new Uri(Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT"))
};

ChatClient client = new(
    model: "gpt-4.1-mini",
    authenticationPolicy: bearerTokenPolicy,
    options: clientOptions);

ChatCompletion completion = client.CompleteChat("Say 'this is a test.'");
Console.WriteLine($"[ASSISTANT]: {completion.Content[0].Text}");

```

### DIFF
--- a/api/OpenAI.net8.0.cs
+++ b/api/OpenAI.net8.0.cs
@@ -1402,7 +1402,9 @@ namespace OpenAI.Chat {
         protected internal ChatClient(ClientPipeline pipeline, string model, OpenAIClientOptions options);
         public ChatClient(string model, ApiKeyCredential credential, OpenAIClientOptions options);
         public ChatClient(string model, ApiKeyCredential credential);
+        [Experimental("OPENAI001")]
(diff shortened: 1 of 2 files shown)


KrzysztofCwalina:
why are we doing this only in ChatCLient?

joseharriaga:
> why are we doing this only in ChatCLient?

@KrzysztofCwalina I wanted to start small. If this looks good, I will add it to everything else.

PR #566, opened 2025-07-29
Added a model factory to Responses namespace

KrzysztofCwalina:


src/Custom/Responses/Items/ReasoningResponseItem.cs
```diff
@@ -12,7 +11,7 @@ public partial class ReasoningResponseItem
 {
     // CUSTOM: Retain optionality of OpenAPI read-only property value
     [CodeGenMember("Status")]
-    public ReasoningStatus? Status { get; }
+    public ReasoningStatus? Status { get; internal set; }
```
joseharriaga, line 14: Interesting. This is due to a bug in the TypeSpec: The "status" property is missing there. We should fix that there. Can be done as a separate PR.

src/Custom/Responses/OpenAIResponsesModelFactory.cs
```diff
@@ -0,0 +1,146 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OpenAI.Responses;
+
+/// <summary> Model factory for models. </summary>
+public static partial class OpenAIResponsesModelFactory
```
joseharriaga, line 10: Add `[Experimental("OPENAI001")]` to the class.

joseharriaga (changes requested):

src/Custom/Responses/Items/ReasoningResponseItem.cs
KrzysztofCwalina: where do we file these issues?

src/Custom/Responses/OpenAIResponsesModelFactory.cs
KrzysztofCwalina: done

src/Custom/Responses/Items/ReasoningResponseItem.cs
joseharriaga: I'll report it to Travis.

joseharriaga (approved):

src/Custom/Responses/Items/ReasoningResponseItem.cs
joseharriaga: I have a fix here:
🔗 https://github.com/openai/openai-dotnet/pull/567

With this PR, `Status` can now be set via the internal constructor of `ReasoningResponseItem` like every other property, so this internal setter is no longer needed.

joseharriaga (approved):

PR #574, opened 2025-07-31
Add workflow to auto update generator version

JoshLove-msft:


.github/workflows/update-generator.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 8.x
+          
+      - name: Check for generator updates
+        id: check-updates
+        run: |
+          # Get current version from both package.json files
+          CURRENT_VERSION_OPENAI=$(node -p "require('./dotnet/.plugin/package.json').dependencies['@typespec/http-client-csharp']" 2>/dev/null || echo "unknown")
```
jorgerangel-msft, line 40: I think I'm a little confused with the new generator workflow here. Is `./dotnet` the correct path to use here?

JoshLove-msft: Yeah probably this needs to be reconciled with the new structure. This was an exact copy from the private repo.

PR #576, opened 2025-07-31
Add UpdateChatCompletion method in ChatClient

ShivangiReja:
This PR adds both protocol and convenience methods for `UpdateChatCompletion` in `ChatClient` to support updating the metadata of a previously stored chat completion.


specification/client/chat.client.tsp
```diff
@@ -17,3 +17,5 @@ using Azure.ClientGenerator.Core;
 @@visibility(ChatCompletionStreamResponseDelta.tool_calls, Lifecycle.Read);
 
 @@clientName(Chat.createChatCompletion, "CompleteChat");
+
+@@scope(Chat.getChatCompletionMessages, "!csharp");
```
joseharriaga: Chris just merged a PR that adds support for this operation, so I think this can be removed.

joseharriaga (changes requested):

PR #577, opened 2025-08-01
Fix DI example in README.md

evilsquirr3l:
There is a small error in README.md: `ChatClient` doesn't have a constructor with only one parameter, and you must always provide a model as the first param.


ShivangiReja (approved):
Thanks for the contribution! You're absolutely right that the model needs to be provided. Really appreciate you taking the time to fix the README!

PR #597, opened 2025-08-11
Add GitHub action to auto-update TypeSpec generator version

ShivangiReja:
This PR implements a comprehensive GitHub Actions workflow to automatically keep the TypeSpec generator version up to date.


JoshLove-msft (approved):

scripts/Submit-GeneratorUpdatePr.ps1
```diff
@@ -0,0 +1,226 @@
...
+
+    # Get current version from package.json files
+    $openAiPackageJsonPath = "package.json"
+    
+    if (-not (Test-Path $openAiPackageJsonPath)) {
+        throw "OpenAI package.json not found at: $openAiPackageJsonPath"
+    }
+    
+    # Read current versions
+    $openAiPackageJson = Get-Content $openAiPackageJsonPath -Raw | ConvertFrom-Json
```
jorgerangel-msft, line 76: What will be the path this script will be invoked from? If the intent is to get the version of the generator _for the plugin_ then I think we will need to check the plugin's [package.json](https://github.com/openai/openai-dotnet/blob/main/codegen/package.json#L33) ? Otherwise if this runs from the root of the repo, that package.json doesn't have an entry for the generator

JoshLove-msft: This should be invoked from the github/workflows directory

jorgerangel-msft: hmm, then I think we may need to update the script to reference the correct package.json when parsing the version of the generator, unless I'm missing something here

JoshLove-msft: Yeah I think you are right - it would be this file https://github.com/openai/openai-dotnet/blob/main/codegen/package.json

ShivangiReja: Updated the path. Thanks!

scripts/Submit-GeneratorUpdatePr.ps1
```diff
@@ -0,0 +1,226 @@
...
+        $openAiCsproj = $openAiCsproj -replace '(<PackageReference Include="Microsoft\.TypeSpec\.Generator\.ClientModel" Version=")[^"]*(")', "`${1}$PackageVersion`${2}"
+        Set-Content -Path $openAiCsprojPath -Value $openAiCsproj
+        Write-Log "Updated OpenAI csproj: $openAiCsprojPath"
+    } else {
+        Write-Warning-Log "OpenAI csproj not found at: $openAiCsprojPath"
+    }
+    
+    # Install dependencies from root (using workspaces)
+    Write-Log "Installing dependencies from root (using npm workspaces)"
+    npm install
```
jorgerangel-msft, line 120: We may need to Push-Location to the codegen directory then run `npm install` so the plugin installs the updated dependencies prior to the `npm run build` call. If there are issues with this current implementation, we can see if updating the script to do that helps. Just something to keep in mind

scripts/Submit-GeneratorUpdatePr.ps1
```diff
@@ -0,0 +1,226 @@
...
+    if ($LASTEXITCODE -ne 0) {
+        Write-Warning-Log "OpenAI plugin build failed, but continuing..."
+    }
+    Pop-Location
+    
+    # Regenerate OpenAI SDK code
+    Write-Log "Regenerating OpenAI SDK code"
+    Push-Location "."
+    try {
+        pwsh .scripts/Invoke-CodeGen.ps1
```
jorgerangel-msft:
```suggestion
        pwsh scripts/Invoke-CodeGen.ps1
```

jorgerangel-msft (approved):

jorgerangel-msft (approved):

PR #601, opened 2025-08-14
[Tools] Adding snippet generator and update script

jsquire:
# Summary

The focus of these changes is to add the snippet generator tool from the Azure development feed and a PowerShell script to update snippets.

### DIFF
--- a/scripts/Update-Snippets.ps1
+++ b/scripts/Update-Snippets.ps1
@@ -1,0 +1,24 @@
+[CmdletBinding()]
+param(
+    [string]$Path
+)
+
+# If this is not a real path, assume snippets should be refreshed in a relative
+# path to the OpenAI.csproj file from the git root.
+if (-not $Path) {
+    $_gitRoot = git rev-parse --show-toplevel 2>$null
+
+    if ((-not $_gitRoot) -or ($LASTEXITCODE -ne 0)) {
+        Write-Error "Could not determine the git root directory. Please specify the path to the project file."
+        exit 1
+    }
+
+    $Path = Join-Path -Path $_gitRoot -ChildPath '/src/OpenAI.csproj'
+    $Path = (Resolve-Path -LiteralPath $Path).Path
+}
+
+# Update snippets
+Write-Host "Updating snippets in $Path"
+
+dotnet tool restore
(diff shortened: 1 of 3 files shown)


scripts/Update-Snippets.ps1
```diff
@@ -0,0 +1,24 @@
...
+
+    $Path = Join-Path -Path $_gitRoot -ChildPath '/src/OpenAI.csproj'
+    $Path = (Resolve-Path -LiteralPath $Path).Path
+}
+
+# Update snippets
+Write-Host "Updating snippets in $Path"
+
+dotnet tool restore
+dotnet tool run snippet-generator -b $Path $additionalArgs
```
jsquire:
```suggestion
dotnet tool run snippet-generator -b $Path
```

Missed scrubbing this.   I'll either do it local on the next push or take this suggestion before merge.

scripts/Update-Snippets.ps1
```diff
@@ -0,0 +1,24 @@
+[CmdletBinding()]
```
christothes, line 1: Should we add something to the readme or some other contributing details mentioning this script?

jsquire: We absolutely need a Contributing Guide update now that everything is moved over to public and we can accept contributions.    (The current guide basically says "don't.")  
I'm going to leave that as a separate task.

scripts/Update-Snippets.ps1
joseharriaga: If possible, it would be nice to convert one of the examples to a snippet as part of this PR to confirm that it all works. A good candidate might be the very first one, under the "Using the client library" section of the README. That code snippet comes from here:
🔗 https://github.com/openai/openai-dotnet/blob/main/examples/Chat/Example01_SimpleChat.cs

joseharriaga (approved):

scripts/Update-Snippets.ps1
jsquire: I'm going to leave moving samples over to snippets for a new scope and PR.   We're picking up an existing tool that has test coverage here, so we can safely assume that it works.   The script is new, but straight forward and based on prior art.    I've got an issue tracking moving docs over to snippets that will cover this one.

PR #604, opened 2025-08-14
Generate ListInputItems for Responses

christothes:


api/OpenAI.net8.0.cs
```diff
@@ -4704,6 +4704,10 @@ public class OpenAIResponseClient {
         public virtual ClientResult<ResponseDeletionResult> DeleteResponse(string responseId, CancellationToken cancellationToken = default);
         public virtual Task<ClientResult> DeleteResponseAsync(string responseId, RequestOptions options);
         public virtual Task<ClientResult<ResponseDeletionResult>> DeleteResponseAsync(string responseId, CancellationToken cancellationToken = default);
+        public virtual CollectionResult<ResponseItem> GetInputItems(string responseId, ResponseItemCollectionOptions options = null, CancellationToken cancellationToken = default);
+        public virtual CollectionResult GetInputItems(string responseId, int? limit, string order, string after, string before, RequestOptions options);
```
joseharriaga, line 4708: I think it would be nice to normalize the parameter order. For example, in `ChatClient`'s `GetChatCompletionMessages` we have 1) after, 2) limit, and 3) order, but here we have 1) limit, 2) order, 3) after, and 4) before. I don't have a strong preference but I'm thinking the former might be slightly better (i.e., starting with after and before) since those are necessary to be able to paginate. The current order comes from the TypeSpec. In general, I think we should avoid modifying the TypeSpec, but I'm thinking this change would be benign since one could argue that the TypeSpec also benefits from the consistency.

This method comes from here:
🔗 https://github.com/openai/openai-dotnet/blob/main/specification/base/typespec/responses/operations.tsp#L85

It is using the `CommonPageQueryParameters` defined here:
🔗 https://github.com/openai/openai-dotnet/blob/2237506a1aff0ac2c7ca339e8d6b498bf31f8e3a/specification/base/typespec/common/custom.tsp#L65-L70

So if we rearrange those `CommonPageQueryParameters` and re-generate, this issue should be fixed.

joseharriaga (approved):

api/OpenAI.net8.0.cs
christothes: I'm going to defer this for now since making this change affected more generated code than I anticipated. I opened https://github.com/openai/openai-dotnet/issues/627 to track.

PR #606, opened 2025-08-14
Implement generated ListBatches operations

christothes:


src/Generated/BatchClient.cs
```diff
@@ -22,6 +23,26 @@ protected BatchClient()
 
         public ClientPipeline Pipeline { get; }
 
+        public virtual CollectionResult GetBatches(string after, int? limit, RequestOptions options)
```
christothes, line 26: Shouldn't we be generating the doc comments?

jorgerangel-msft: It looks like we have xml-docs disabled https://github.com/openai/openai-dotnet/blob/main/specification/tspconfig.yaml#L14

src/Custom/Batch/BatchCollectionOptions.cs
```diff
@@ -0,0 +1,4 @@
+namespace OpenAI.Batch;
+
+// CUSTOM: Use the correct namespace.
+[CodeGenType("BatchCollectionOptions")] public partial class BatchCollectionOptions { }
```
joseharriaga, line 4: nit: The name of this file doesn't match the class.

jorgerangel-msft (approved):

src/Custom/Batch/BatchCollectionOptions.cs
christothes: fixed!

Issue #520, opened 2025-07-07
[Protocol Methods] Pagination support: Generated methods should support pagination patterns where applicable.


ShivangiReja:
**Action Item**: Check TypeSpec tooling for any gaps in pagination generation.

ShivangiReja:
CC: @christothes

christothes:
related: https://github.com/openai/openai-dotnet/pull/549

PR #612, opened 2025-08-18
Convert GetVectorStores from custom to generated code

christothes:
related #520


src/Custom/VectorStores/VectorStoreClient.cs
```diff
@@ -285,20 +265,6 @@ public virtual AsyncCollectionResult<VectorStore> GetVectorStoresAsync(
             as AsyncCollectionResult<VectorStore>;
     }
```
joseharriaga: Should this overload of `GetVectorStoresAsync` be deleted too?

src/Custom/VectorStores/VectorStoreClient.cs
```diff
@@ -285,20 +265,6 @@ public virtual AsyncCollectionResult<VectorStore> GetVectorStoresAsync(
...
-        VectorStoreCollectionOptions options = default,
-        CancellationToken cancellationToken = default)
-    {
-        return GetVectorStores(options?.PageSizeLimit, options?.Order?.ToString(), options?.AfterId, options?.BeforeId, cancellationToken.ToRequestOptions())
-            as CollectionResult<VectorStore>;
-    }
-
     /// <summary>
     /// Rehydrates a page collection holding <see cref="VectorStore"/> instances from a page token.
     /// </summary>
```
joseharriaga: Should this overload of `GetVectorStores` be deleted too?

src/Custom/VectorStores/VectorStoreClient.cs
```diff
@@ -594,157 +484,6 @@ public virtual CreateBatchFileJobOperation CreateBatchFileJob(
         return CreateBatchFileJob(vectorStoreId, content, waitUntilCompleted, options);
     }
```
joseharriaga, line 452: I think this overload of `GetBatchFileJobAsync` was deleted by accident.

api/OpenAI.net8.0.cs
```diff
@@ -5578,23 +5578,18 @@ public class VectorStoreClient {
...
-        public virtual AsyncCollectionResult<VectorStoreFileAssociation> GetFileAssociationsAsync(string vectorStoreId, string batchJobId, VectorStoreFileAssociationCollectionOptions options = null, CancellationToken cancellationToken = default);
-        public virtual AsyncCollectionResult<VectorStoreFileAssociation> GetFileAssociationsAsync(string vectorStoreId, string batchJobId, ContinuationToken firstPageToken, CancellationToken cancellationToken = default);
-        public virtual AsyncCollectionResult GetFileAssociationsAsync(string vectorStoreId, string batchId, int? limit, string order, string after, string before, string filter, RequestOptions options);
+        public virtual CollectionResult<VectorStoreFileAssociation> GetFileAssociationsInBatch(string vectorStoreId, string batchId, VectorStoreFileAssociationCollectionOptions options = null, CancellationToken cancellationToken = default);
+        public virtual CollectionResult GetFileAssociationsInBatch(string vectorStoreId, string batchId, int? limit, string order, string after, string before, string filter, RequestOptions options);
```
joseharriaga, line 5768: Good call with the rename to add "InBatch" to make it clearer!

api/OpenAI.net8.0.cs
```diff
@@ -5693,27 +5692,16 @@ public class VectorStoreFileAssociation : IJsonModel<VectorStoreFileAssociation>
...
+    public class VectorStoreFileAssociationCollectionOptions : IJsonModel<VectorStoreFileAssociationCollectionOptions>, IPersistableModel<VectorStoreFileAssociationCollectionOptions> {
         public string AfterId { get; set; }
         public string BeforeId { get; set; }
         public VectorStoreFileStatusFilter? Filter { get; set; }
-        public VectorStoreFileAssociationCollectionOrder? Order { get; set; }
+        public VectorStoreCollectionOrder? Order { get; set; }
         public int? PageSizeLimit { get; set; }
-    }
-    [Experimental("OPENAI001")]
-    public readonly partial struct VectorStoreFileAssociationCollectionOrder : IEquatable<VectorStoreFileAssociationCollectionOrder> {
```
joseharriaga, line 5884: For consistency, would it make sense to keep the "files collection" order enum separate from the "vector stores collection" order enum?

src/Custom/VectorStores/VectorStoreClient.cs
christothes: You are correct, thanks for the catch!

src/Custom/VectorStores/VectorStoreClient.cs
christothes: Yes!

src/Custom/VectorStores/VectorStoreClient.cs
christothes: Yes!

api/OpenAI.net8.0.cs
christothes: This type actually doesn't exist anymore (need to re-export). 

I debated this.. While it is more consistent, it would create two new types without any difference in functionality. Perhaps a single order type is enough per client. Currently they all use `VectorStoreCollectionOrder`. Thoughts?

joseharriaga: You bring up a good point, and I see pros and cons for the different possible approaches. On one hand, I do think that having different types that are functionally identical for every operation is definitely too much. On the other hand, I wonder if users would be confused by the name if we re-use it across different operations. For example: The name of `VectorStoreCollectionOrder` implies a collection of Vector Stores, but in this case we're reusing it to also refer to a collection of Vector Store Files, which might be a little surprising. Because of this, I'm leaning towards having separate types based on the return value. In other words:
* `GetVectorStores` (which returns a collection of `VectorStore`) uses `VectorStoreCollectionOptions` which in turn uses `VectorStoreCollectionOrder`.
* But `GetFileAssociations` and `GetFileAssociationsInBatch` (both of which return a collection of `VectorStoreFileAssociation`) both use `VectorStoreFileAssociationCollectionOptions` which in turn uses `VectorStoreFileAssociationOrder`.

This increases the number of `*Order` types from 1 to 2, but it also decreases the number of `*Options` types from 3 to 2. This feels like a good balance to me.

What do you think?

codegen/generator/src/Visitors/PaginationVisitor.cs
```diff
@@ -18,57 +18,82 @@ namespace OpenAILibraryPlugin.Visitors;
...
         },
         {
             "GetChatCompletionMessages",
-            ("ChatCompletionMessageListDatum", "ChatCompletionMessageCollectionOptions", _chatParamsToReplace)
+            ("ChatCompletionMessageListDatum", "ChatCompletionCollectionOptions", _paginationParamsToReplace)
         },
         {
             "GetChatCompletionMessagesAsync",
-            ("ChatCompletionMessageListDatum", "ChatCompletionMessageCollectionOptions", _chatParamsToReplace)
+            ("ChatCompletionMessageListDatum", "ChatCompletionMessageCollectionOptions", _paginationParamsToReplace)
```
joseharriaga, line 48: Adding a note here because this is related to the other thread regarding `*Options` and `*Order` types. Depending on what we decide, we should just ensure that Chat aligns with it for consistency. 🙂 

I just noticed that `GetChatCompletionMessages` was changed but its `*Async` counterpart wasn't.

api/OpenAI.net8.0.cs
christothes: I don't feel strongly either way, so I think adding 1 extra type is valid for consistency's sake.

codegen/generator/src/Visitors/PaginationVisitor.cs
christothes: thanks - fixed!

src/Custom/VectorStores/GeneratorStubs.cs
```diff
@@ -0,0 +1,6 @@
+namespace OpenAI.VectorStores;
+
+/// <summary> The options to configure how <see cref="VectorStore"/> objects are retrieved and paginated. </summary>
+// Customization: fix namespace
+[CodeGenType("VectorStoreFileAssociationCollectionOptions")] public partial class VectorStoreFileAssociationCollectionOptions{ }
+[CodeGenType("VectorStoreFileAssociationOrder")] public partial struct VectorStoreFileAssociationOrder{ }
```
joseharriaga: This one should be called `VectorStoreFileAssociationCollectionOrder` for consistency. 

(It's a little bit of a mouthful for now, but I also have a proposal in the works to rename `VectorStoreFileAssociation*` to simply `VectorStoreFile*` to match the REST API, but this rename will come later in the future. 🙂)

joseharriaga (approved):

src/Custom/VectorStores/GeneratorStubs.cs
christothes: fixed!

joseharriaga (approved):

PR #615, opened 2025-08-19
build: mark stage as succeeded with issues when there are errors

jorgerangel-msft:


scripts/Submit-GeneratorUpdatePr.ps1
```diff
@@ -50,6 +52,8 @@ function Write-Log {
 function Write-Warning-Log {
     param([string]$Message)
     Write-Host "$(Get-Date -Format 'yyyy-MM-dd HH:mm:ss'): WARNING: $Message" -ForegroundColor Yellow
+    # Set the global warning flag to track that warnings occurred
+    $script:WarningsEncountered = $true
```
ShivangiReja, line 56: Is this only handling warnings? What about the errors? Like:
image
Ref: https://github.com/openai/openai-dotnet/actions/runs/17036113061/job/48288884625

jorgerangel-msft: The "warnings" should be getting logged as part of the try / catch block for these commands. ie https://github.com/openai/openai-dotnet/blob/main/scripts/Submit-GeneratorUpdatePr.ps1#L197. From my understanding, this means if any part of the command invocation returns a non-success exit code (warning + errors) then the script should be flagged as successful with issues.

PR #620, opened 2025-08-21
Update @typespec/http-client-csharp to 1.0.0-alpha.20250722.1

ShivangiReja:
This PR updates the TypeSpec HTTP client C# generator version to 1.0.0-alpha.20250722.1 and regenerates the SDK code.


api/OpenAI.net8.0.cs
```diff
@@ -42,7 +42,6 @@ public class OpenAIClientOptions : ClientPipelineOptions {
         public string ProjectId { get; set; }
         public string UserAgentApplicationId { get; set; }
     }
-    [Experimental("OPENAI001")]
```
jorgerangel-msft, line 45: was this expected? Do we need to also account for the context in our experimental visitor?

ShivangiReja: Yes, I was adding the Experimental attribute through the visitor, but @JoshLove-msft suggested we don’t want this attribute for `OpenAIContext`, so I didn’t include it there.

jorgerangel-msft (approved):

PR #621, opened 2025-08-21
Add `Quality` and `Background` properties to `ImageEditOptions`

joseharriaga:


src/Custom/OpenAIContext.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace OpenAI;
+
+[Experimental("OPENAI001")]
+public partial class OpenAIContext
```
christothes, line 6: why is this needed? If it's for the `Experimental` attribute, shouldn't the visitor do this automatically?

joseharriaga: It wasn't caught by the visitor. This is a "special" type, so I suspect it might have failed this check:
`type is ClientProvider or ModelProvider or ClientOptionsProvider or EnumProvider`

PR #624, opened 2025-08-25
Generate container convenience APIs including paginated APIs

christothes:
This PR also removes the `ExplicitConversionFromClientResultVisitor` so that all explicit cast operators for models get generated.


src/Custom/Containers/ContainerFileCollectionOptions.cs
```diff
@@ -0,0 +1,4 @@
+namespace OpenAI.Containers;
+
+// CUSTOM: Use the correct namespace.
+[CodeGenType("ContainerFileCollectionOptions")] public partial class ContainerFileCollectionOptions { }
```
ShivangiReja: Can we move `ContainerFileCollectionOptions`, `ContainerCollectionOrder`, and `ContainerCollectionOptions` custom classes/structs to `GeneratorStubs`, since we’re only creating these to use the correct namespace?

christothes: done!

joseharriaga:
Looks great! 🙂

PR #625, opened 2025-08-25
Add `ResponseItem` constructors

joseharriaga:
All the classes derived from `ResponseItem` now have public constructors, and these constructors are generated.

A side effect of making these classes behave as both, inputs and outputs, was that their required properties got setters. Setters make sense when the values of these properties can be updated, but in this particular case they should only be settable at construction time. This is a limitation in the code generator being tracked by the following issue:
🔗 https://github.com/microsoft/typespec/issues/5175#issuecomment-3222039076

### DIFF
--- a/api/OpenAI.netstandard2.0.cs
+++ b/api/OpenAI.netstandard2.0.cs
@@ -4314,7 +4314,7 @@ namespace OpenAI.Responses {
         Incomplete = 2
     }
     public class ReasoningSummaryPart : IJsonModel<ReasoningSummaryPart>, IPersistableModel<ReasoningSummaryPart> {
-        public static ReasoningSummaryPart CreateTextPart(string text);
+        public static ReasoningSummaryTextPart CreateTextPart(string text);
         protected virtual ReasoningSummaryPart JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
         protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
         protected virtual ReasoningSummaryPart PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
--- a/api/OpenAI.net8.0.cs
+++ b/api/OpenAI.net8.0.cs
@@ -4888,7 +4888,7 @@ namespace OpenAI.Responses {
     }
     [Experimental("OPENAI001")]
     public class ReasoningSummaryPart : IJsonModel<ReasoningSummaryPart>, IPersistableModel<ReasoningSummaryPart> {
-        public static ReasoningSummaryPart CreateTextPart(string text);
+        public static ReasoningSummaryTextPart CreateTextPart(string text);
         protected virtual ReasoningSummaryPart JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
         protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
         protected virtual ReasoningSummaryPart PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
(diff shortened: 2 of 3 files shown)


api/OpenAI.netstandard2.0.cs
```diff
@@ -4038,16 +4039,17 @@ public class ComputerOutput : IJsonModel<ComputerOutput>, IPersistableModel<Comp
         public override readonly string ToString();
     }
     public class FileSearchCallResponseItem : ResponseItem, IJsonModel<FileSearchCallResponseItem>, IPersistableModel<FileSearchCallResponseItem> {
+        public FileSearchCallResponseItem(IEnumerable<string> queries);
         public IList<string> Queries { get; }
-        public IList<FileSearchCallResult> Results { get; }
+        public IList<FileSearchCallResult> Results { get; set; }
```
ShivangiReja, line 4119: I believe we don't provide setter for collection properties? is it not the case here?

src/Custom/Responses/Items/Internal/InternalResponsesAssistantMessage.cs
```diff
@@ -9,8 +9,8 @@ internal partial class InternalResponsesAssistantMessage
...
     public IList<ResponseContentPart> InternalContent { get; }
 
-    // CUSTOM: For reuse as an input model
-    internal InternalResponsesAssistantMessage(IEnumerable<ResponseContentPart> internalContent)
-        : this(id: null, status: null, internalContent)
-    { }
+    //// CUSTOM: For reuse as an input model
+    //internal InternalResponsesAssistantMessage(IEnumerable<ResponseContentPart> internalContent)
+    //    : this(status: null, internalContent)
+    //{ }
```
ShivangiReja: nit: can we remove these comments?

src/Custom/Responses/Items/Internal/InternalResponsesDeveloperMessage.cs
```diff
@@ -9,8 +9,8 @@ internal partial class InternalResponsesDeveloperMessage
...
     public IList<ResponseContentPart> InternalContent { get; }
 
-    // CUSTOM: For reuse as an input model
-    internal InternalResponsesDeveloperMessage(IEnumerable<ResponseContentPart> internalContent)
-        : this(id: null, status: null, internalContent)
-    { }
+    //// CUSTOM: For reuse as an input model
+    //internal InternalResponsesDeveloperMessage(IEnumerable<ResponseContentPart> internalContent)
+    //    : this(status: null, internalContent)
+    //{ }
```
ShivangiReja: nit: remove comments

src/Custom/Responses/Items/Internal/InternalResponsesSystemMessage.cs
```diff
@@ -9,8 +9,8 @@ internal partial class InternalResponsesSystemMessage
...
     public IList<ResponseContentPart> InternalContent { get; }
 
-    // CUSTOM: For reuse as an input model
-    internal InternalResponsesSystemMessage(IEnumerable<ResponseContentPart> internalContent)
-        : this(id: null, status: null, internalContent)
-    { }
+    //// CUSTOM: For reuse as an input model
+    //internal InternalResponsesSystemMessage(IEnumerable<ResponseContentPart> internalContent)
+    //    : this(status: null, internalContent)
+    //{ }
```
ShivangiReja: nit: remove comments

src/Custom/Responses/Items/Internal/InternalResponsesUserMessage.cs
```diff
@@ -9,8 +9,8 @@ internal partial class InternalResponsesUserMessage
...
     public IList<ResponseContentPart> InternalContent { get; }
 
-    // CUSTOM: For reuse as an input model
-    internal InternalResponsesUserMessage(IEnumerable<ResponseContentPart> internalContent)
-        : this(id: null, status: null, internalContent)
-    { }
+    //// CUSTOM: For reuse as an input model
+    //internal InternalResponsesUserMessage(IEnumerable<ResponseContentPart> internalContent)
+    //    : this(status: null, internalContent)
+    //{ }
```
ShivangiReja: nit: remove comments

api/OpenAI.netstandard2.0.cs
joseharriaga: Yes, this is correct. See my comment in PR description. This is a limitation in our codegen, but since this is experimental, I don't think it should be a blocker. 🙂

ShivangiReja (approved):

PR #626, opened 2025-08-25
[wip] build: bump to latest generator version

jorgerangel-msft:


src/Custom/Chat/Messages/AssistantChatMessage.cs
```diff
@@ -21,7 +21,15 @@ public partial class AssistantChatMessage : ChatMessage
     ///     The collection of content items associated with the message.
     /// </param>
     public AssistantChatMessage(IEnumerable<ChatMessageContentPart> contentParts)
-        : this(new ChatMessageContent(contentParts), ChatMessageRole.Assistant, null, null, null, null, null, null)
+        : this(
+              content: new ChatMessageContent(contentParts),
```
christothes, line 25: Is this just intended to be an improvement or is it required for some reason related to the generator bump?

christothes: Oh - is this to address the parameter re-ordering?

jorgerangel-msft: It's more of an "improvement". The main reason I added it was to avoid having to make hand edits if the parameter ordering ever changes again for the generated models' internal constructors. If we feel like this doesn't fit the established pattern, I'm fine with sticking with positional args. Just trying to reduce the burden of hand editing things 😃

christothes: Did you also re-export the public API? I wonder if any of these re-orders are reflected there, which would be a breaking change.

jorgerangel-msft: Yes I did

jorgerangel-msft: From what I saw, most if not all, the reordering happened in the internal serialization constructors for the generated models.

codegen/package.json
```diff
@@ -29,11 +29,11 @@
...
-        "@azure-tools/typespec-client-generator-core": "0.57.2",
-        "@azure-tools/typespec-azure-core": "0.57.0",
-        "@typespec/http-client-csharp": "1.0.0-alpha.20250722.1",
-        "@typespec/http": "1.1.0",
-        "@typespec/openapi": "1.1.0"
+        "@azure-tools/typespec-client-generator-core": "0.59.0",
+        "@azure-tools/typespec-azure-core": "0.59.0",
+        "@typespec/http-client-csharp": "1.0.0-alpha.20250827.2",
+        "@typespec/http": "1.3.0",
+        "@typespec/openapi": "1.3.0"
```
ShivangiReja, line 36: With the latest versions, do we still need to pin all the peer dependencies?

[Automation](https://github.com/openai/openai-dotnet/actions/runs/17036113061/job/48288884625) was able to run `npm install` successfully with just [these](https://github.com/openai/openai-dotnet/pull/611/files#diff-a8982328bdcbe070d823bf6b522131c1ca0a8e12d125d78d00d7897e59058c17) dependencies:

```json
"@azure-tools/typespec-client-generator-core": "0.59.0",
"@typespec/http-client-csharp": "1.0.0-alpha.20250815.7"
```

So I was just wondering if we still need those additional dependencies?

jorgerangel-msft: It might work today, but the generator accepts a range for it's peer dependencies which I think is what was leading to a mismatch unless the deps were pinned. With that being said, if we want to try reverting back to only having tcgc pinned, I think that could be a good follow up PR  to experiment with.

codegen/generator/src/OpenAILibraryGenerator.cs
```diff
@@ -11,7 +11,7 @@ namespace OpenAILibraryPlugin
     public class OpenAILibraryGenerator : ScmCodeModelGenerator
     {
         private static OpenAILibraryGenerator? s_instance;
-        internal static OpenAILibraryGenerator Instance => s_instance ?? throw new InvalidOperationException("OpenAILibraryGenerator was not initialized.");
+        internal static new OpenAILibraryGenerator Instance => s_instance ?? throw new InvalidOperationException("OpenAILibraryGenerator was not initialized.");
```
JoshLove-msft: nit: do we still need this property now that it is available as a public property?

JoshLove-msft (approved):

codegen/generator/src/OpenAILibraryGenerator.cs
jorgerangel-msft: Good catch. I don't think so. I went ahead and removed it and regenerated. It seems to still be working correctly.

joseharriaga (approved):

Issue #592, opened 2025-08-08
[FEATURE REQ] – Cannot access internal property ServiceTier in ResponseCreationOptions when using Create Response Client


mjmustafaev:
### Describe the feature or improvement you are requesting

The .NET SDK’s ResponseCreationOptions includes a ServiceTier field in its serialized output, but it’s typed as InternalServiceTier? and is therefore inaccessible from application code, preventing setting the service_tier parameter supported by the REST API ("flex", "standard", "premium"). This forces to specify a service tier to bypass the strongly typed client and send raw HTTP requests instead, breaking consistency. Making ServiceTier public with a documented enum or string setter would be very helpful.

### Additional context

It is to set the service tier parameter mentioned here: https://platform.openai.com/docs/guides/priority-processing

jsquire:
Hi @mjmustafaev.  Thanks for reaching out and we regret that you're experiencing difficulties.  This field was internalized for an earlier version of the REST API and should be public at this point.   We'll look at making the adjustment for a future release.

In the meantime, you can work around this and continue to use the strongly typed models by influencing serialization, which would look something like:

```csharp
var client = new OpenAIResponseClient("gpt-4.1", "<< YOUR API KEY >>");

// Force an additional member into the options properties bag.
var options = ((IJsonModel<ResponseCreationOptions>)new ResponseCreationOptions())
    .Create(BinaryData.FromObjectAsJson(new
    {
        service_tier = "auto"
    }), 
    ModelReaderWriterOptions.Json);
    
// Set the desired options with the strongly-typed interface.
options.Instructions = "Always talk like a pirate when you answer.";
    
// The service call can be made as normal.
var response = await client.CreateResponseAsync(
    "What is Krzsztof Cwalina's shoe size?",
    options);

Console.WriteLine(response.Value.GetOutputText());
```

mjmustafaev:
Understood, thanks very much. Super helpful info and example 👍 🙏 @jsquire

ShivangiReja:
Hi @mjmustafaev, thanks for raising this and for the detailed context!

We’ve fixed this in [PR #628](https://github.com/openai/openai-dotnet/pull/628). The change will be included in the next release of the SDK.

Really appreciate your patience and feedback!

PR #628, opened 2025-08-26
Add service tier support to Chat and Responses APIs

ShivangiReja:
Introduces service-tier support for Chat and Responses.

**Chat:**
* Added extensible enum `ChatServiceTier`.
* Added `ServiceTier` property to ChatCompletionOptions (request).
* Added `ServiceTier` to ChatCompletion and StreamingChatCompletionUpdate (response).

**Responses:**
* Added extensible  enum `ResponseServiceTier`.
* Added `ServiceTier` property to ResponseCreationOptions(request).
* Added `ServiceTier` to OpenAIResponse(response) .

Issue: https://github.com/openai/openai-dotnet/issues/592


specification/client/chat.client.tsp
```diff
@@ -18,3 +18,7 @@ using Azure.ClientGenerator.Core;
 @@visibility(ChatCompletionStreamResponseDelta.tool_calls, Lifecycle.Read);
 
 @@clientName(Chat.createChatCompletion, "CompleteChat");
+
+@@alternateType(CreateChatCompletionRequest.service_tier, DotNetChatServiceTier);
+@@alternateType(CreateChatCompletionResponse.service_tier, DotNetChatServiceTier);
+@@alternateType(CreateChatCompletionStreamResponse.service_tier, DotNetChatServiceTier);
```
christothes, line 24: Just curious - why did we have to do this versus only renaming `InternalServiceTier` and making it public?

christothes: Thinking about it some more, this is so that we can have the differentiated type name for Chat and Responses. 😄

christothes (approved):

specification/client/chat.client.tsp
ShivangiReja: In the spec, `ServiceTier` is defined as a common type shared across multiple subclients (like Chat and Responses) but in the .NET SDK, we don’t share types between subclients. Instead, for common types we generate custom wrappers per subclient. That’s why we introduce subclient-specific types, see [spec example here](https://github.com/openai/openai-dotnet/blob/caa88171854017930acfb4b0ea060f96db53ee7f/specification/client/models/common.models.tsp#L158-L164):

```tsp
union DotNetChatServiceTier {
  ServiceTier
}

union DotNetResponseServiceTier {
  ServiceTier
}
```

This way each subclient gets its own strongly-typed version while still mapping back to the same spec definition.

api/OpenAI.net8.0.cs
```diff
@@ -1511,6 +1511,8 @@ public class ChatCompletion : IJsonModel<ChatCompletion>, IPersistableModel<Chat
         public string Refusal { get; }
         public IReadOnlyList<ChatTokenLogProbabilityDetails> RefusalTokenLogProbabilities { get; }
         public ChatMessageRole Role { get; }
+        [Experimental("OPENAI001")]
+        public ChatServiceTier? ServiceTier { get; }
```
joseharriaga, line 1515: We should add this to the `ChatCompletion` method of the `OpenAIChatModelFactory`:
🔗 https://github.com/openai/openai-dotnet/blob/75ec79084f6ca85e289df894a2da7f97a2461249/src/Custom/Chat/OpenAIChatModelFactory.cs#L49

api/OpenAI.net8.0.cs
```diff
@@ -2143,6 +2165,8 @@ public class StreamingChatCompletionUpdate : IJsonModel<StreamingChatCompletionU
         public IReadOnlyList<ChatTokenLogProbabilityDetails> RefusalTokenLogProbabilities { get; }
         public string RefusalUpdate { get; }
         public ChatMessageRole? Role { get; }
+        [Experimental("OPENAI001")]
+        public ChatServiceTier? ServiceTier { get; }
```
joseharriaga, line 2169: This one also should be updated in the `OpenAIChatModelFactory`:
🔗 https://github.com/openai/openai-dotnet/blob/75ec79084f6ca85e289df894a2da7f97a2461249/src/Custom/Chat/OpenAIChatModelFactory.cs#L268

api/OpenAI.net8.0.cs
```diff
@@ -4772,6 +4796,7 @@ public class OpenAIResponse : IJsonModel<OpenAIResponse>, IPersistableModel<Open
         public bool ParallelToolCallsEnabled { get; }
         public string PreviousResponseId { get; }
         public ResponseReasoningOptions ReasoningOptions { get; }
+        public ResponseServiceTier? ServiceTier { get; }
```
joseharriaga, line 4799: This one should be updated in the `OpenAIResponseModelFactory`:
🔗 https://github.com/openai/openai-dotnet/blob/main/src/Custom/Responses/OpenAIResponsesModelFactory.cs#L14

joseharriaga (changes requested):

joseharriaga (approved):

PR #646, opened 2025-09-02
Add derived types for `ResponseTool`

joseharriaga:
Added derived for `ResponseTool` to make it easier to work with when used as an output.


src/Custom/Responses/Tools/ComputerTool.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace OpenAI.Responses;
+
+[CodeGenType("ComputerUsePreviewTool")]
```
christothes, line 4: nit: for the added CodeGenType attributes can we add the typical comment about what was customized?

ShivangiReja (approved):

src/Custom/Responses/Tools/ComputerTool.cs
joseharriaga: I added the missing ones. Thanks!

Issue #306, opened 2024-12-10
Consider making a number of properties on `ChatCompletionOptions` public


eiriktsarpalis:
### Confirm this is a feature request for the .NET library and not the underlying OpenAI API

- [X] This is a feature request for the .NET library

### Describe the feature or improvement you are requesting

I'm currently implementing [a feature in Microsoft.Extensions.AI](https://github.com/dotnet/extensions/pull/5697/) that maps OpenAI chat completion requests (as represented in the `ChatCompletionOptions` type) to the Microsoft.Extensions.AI model. This is to accelerate development of Copilot agents that must be able to accept completion requests in the OpenAI wire format.

One of the issues I've encountered is that a number of important properties in the [`ChatCompletionOptions` class](https://github.com/openai/openai-dotnet/blob/main/src/Custom/Chat/ChatCompletionOptions.cs) are currently marked internal. This includes [the `Messages` property](https://github.com/openai/openai-dotnet/blob/c046c34ad06f7ca050b2f8044cfc8372d8cf0e76/src/Custom/Chat/ChatCompletionOptions.cs#L26), [the `Stream` property](https://github.com/openai/openai-dotnet/blob/c046c34ad06f7ca050b2f8044cfc8372d8cf0e76/src/Custom/Chat/ChatCompletionOptions.cs#L45), and [the `Model` property](https://github.com/openai/openai-dotnet/blob/c046c34ad06f7ca050b2f8044cfc8372d8cf0e76/src/Custom/Chat/ChatCompletionOptions.cs#L35).

While I've been able to [work around the issue temporarily using reflection](https://github.com/dotnet/extensions/pull/5697#discussion_r1878230017), this isn't a viable solution longer term. Please consider making these (and other properties in the model) public so that your types can be used as DTOs in the context of agents implementing the OpenAI wire format.

KrzysztofCwalina:
I think the right way to implement the abstractions is to use the protocol methods. They will not only give you access to all, otherwise internal, properties, but you will be able to avoid the cost of instantiating OpenAI model types and then copying their properties into the abstraction layer model types. You can read about the protocol methods here: https://learn.microsoft.com/en-us/dotnet/azure/sdk/protocol-convenience-methods?tabs=convenience-methods%2Csystem-clientmodel

eiriktsarpalis:
Correct me if I'm wrong, but it appears protocol methods necessitate working with a client. In the case of agents though we require processing the wire format from the perspective of a server.

[...]

Issue #644, opened 2025-09-02
[FEATURE REQ] Allow per-request model


verdie-g:
### Describe the feature or improvement you are requesting

The OpenAI protocol allows select a different model for each request, though this is not exposed as `ChatCompletionOptions.Model` is internal.

### Additional context

Fixing https://github.com/openai/openai-dotnet/issues/306 would also fix that issue.

jsquire:
Hi @verdie-g.   Thank you for reaching out and for your suggestion.   Today, it is possible to specify models at the operation level when dropping down to the protocol overloads, but they are otherwise intentionally governed by the client.   We have been discussing whether or not there should be an override on each operation for the model, but no conclusion has been reached yet.

jsquire:
@KrzysztofCwalina, @christothes:  Your thoughts would be appreciated.

KrzysztofCwalina:
I am not totally opposed to exposing Model, but there are negatives of exposing it, and so I would like to understand better why it's a blocker for M.E.AI.A?  M.E.AI.A adapter should probably be calling protocol APIs to avoid converting various small datatypes. 

As to the negatives, I worry a bit that users will see the Model property and will unnecessarily worry that they have to set it. 

In general, I wonder if it's not the time to expose "protocol models", i.e. types that are similar in shape to options/results but ones that correspond 100% to the request/response content.

verdie-g:
Invoking MEAI's developer @eiriktsarpalis that could give a better explanation than I.

verdie-g:
This is how MEAI create the options

https://github.com/dotnet/extensions/blob/0b4c4d53f728a0d4ea12c099bba3f3c1218dc15a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs#L534-L614

They can't set the model here because it's internal.

You said there was another way to set it?

KrzysztofCwalina:
>You said there was another way to set it?

Yeah, the adapter should not be using the options class. It should call so called "protocol" overloads that take RequestContext, e.g. 
public virtual ClientResult CompleteChat(BinaryContent content, RequestOptions options = null); 

This way, the adapter would not be translating from MEAI types to OpenAI Options to JSON, and instead go directly from MEAI types to JSON (the BinaryContent parameter is the JSON)

stephentoub:
[...]

PR #647, opened 2025-09-02
Expose more properties from ChatCompletionOptions

verdie-g:
Fixes https://github.com/openai/openai-dotnet/issues/306, https://github.com/openai/openai-dotnet/issues/644

- Added a public getter to `Messages`, `Model`, `Stream`. That allows creating an OpenAI-compatible endpoint without having to use reflection.
- Added a public setter to `Model`. That allows choosing a different model per request with a fallback to the client's model.
- `StreamOptions`, `ChatResponseFormat` and `ToolChoice` are not addressed in this PR because they have a more complex type that you might want to rework before exposing.

### DIFF
--- a/src/Custom/Chat/ChatClient.cs
+++ b/src/Custom/Chat/ChatClient.cs
@@ -357,7 +357,7 @@ public partial class ChatClient
     private void CreateChatCompletionOptions(IEnumerable<ChatMessage> messages, ref ChatCompletionOptions options, bool stream = false)
     {
         options.Messages = messages.ToList();
-        options.Model = _model;
+        options.Model ??= _model;
         if (stream)
         {
             options.Stream = true;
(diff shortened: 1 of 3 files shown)


jsquire (changes requested):
Hi @verdie-g.   Thank you for your contribution and interest in improving the OpenAI developer experience.   Unfortunately, this is not a submission that will be moving forward at present.   

Changes to the public API are something that we are very intentional about, and which require wider discussion and architect approval.   Please open an issue to discuss your suggestions.

verdie-g:
Let's move back the discussion to the 2 open issues then but https://github.com/openai/openai-dotnet/issues/306 has been open for a while and not being able to set a model is a hard blocker in Microsoft.Extensions.AI :/

PR #648, opened 2025-09-02
Add `ContentParts` to `ChatCompletionMessageListDatum`

christothes:


jsquire (approved):
LGTM, based on compliance with the platform docs:  https://platform.openai.com/docs/api-reference/chat/message-list

api/OpenAI.net8.0.cs
```diff
@@ -2132,7 +2133,7 @@ public class FunctionChatMessage : ChatMessage, IJsonModel<FunctionChatMessage>,
...
+        public static ChatCompletion ChatCompletion(string id = null, ChatFinishReason finishReason = ChatFinishReason.Stop, ChatMessageContent content = null, string refusal = null, IEnumerable<ChatToolCall> toolCalls = null, ChatMessageRole role = ChatMessageRole.System, ChatFunctionCall functionCall = null, IEnumerable<ChatTokenLogProbabilityDetails> contentTokenLogProbabilities = null, IEnumerable<ChatTokenLogProbabilityDetails> refusalTokenLogProbabilities = null, DateTimeOffset createdAt = default, string model = null, ChatServiceTier? serviceTier = null, string systemFingerprint = null, ChatTokenUsage usage = null, ChatOutputAudio outputAudio = null, IEnumerable<ChatMessageAnnotation> messageAnnotations = null, IEnumerable<ChatMessageContentPart> contentParts = null);
```
joseharriaga: Since we have a separate model called `ChatCompletionMessageListDatum`, we should add a factory method for it in this `OpenAIChatModelFactory`, and we should add the `contentParts` parameter to that method instead.

christothes: Adding a factory for `ChatCompletionMessageListDatum` makes sense, but don't we still need the contentParts param here also to properly construct the ChatCompletion?

ShivangiReja: Yes, we still need this parameter to create `InternalChatCompletionResponseMessage` and properly construct the `ChatCompletion`.

joseharriaga: Looking at the API reference, `ChatCompletion` shouldn't have a "content_parts" property; it should only have a "content". Here's the link:
🔗 https://platform.openai.com/docs/api-reference/chat/object
If you check the "chat completion object", under "choices.message" you will only find "content" and no "content_parts". This is contrary to the "chat completion message list object", where you can find "data" and under it you will see both: "content" and also "content_parts".

I think I see where the issue is: In the TypeSpec, there is a model called `ChatCompletionResponseMessage` that represents the message in a `ChatCompletion`. There is also another model called `ChatCompletionMessageListData` that represents the `ChatCompletionMessageListDatum`; it spreads `ChatCompletionResponseMessage` and adds an "id" property to it. In this PR, we're adding the "content_parts" property to the former (so it appears in both models). Instead, the property should be added to the latter so that it only appears there (similar to "id").

christothes: Thanks @joseharriaga - I've fixed this.

joseharriaga (approved):

Issue #447, opened 2025-06-02
Complete conversation state remapping


lmmartinez97:
### Confirm this is a feature request for the .NET library and not the underlying OpenAI API

- [x] This is a feature request for the .NET library

### Describe the feature or improvement you are requesting

Hi,

I have seen another issue regarding this topic that was closed without giving a closed answer.

I'm working in a system that is able to reproduce the entire state of a conversation to give my users the chance to recover chats that were last updated more than 30 days ago. To do this, I would like to be able to build the input item list with things like annotations and such.

As it is right now, the constructors for the ResponseMessageAnnotation class are internal, and the public partial class does not include a constructor, which effectively means we cannot add annotations to the output parts of the input item list, even though the ResponseContentPart.CreateOutputTextPart static method accepts an enumerable of ResponseMessageAnnotations. 

My proposal is to add factory methods to the ResponseMessageAnnotation class, in the same way that the ResponseContentPart class has them. 

Thanks!

jsquire:
Hi @lmmartinez97.   Thanks for reaching out and we regret that you're experiencing difficulties.  It looks as if this is an oversight with `ResponseMessageAnnotation` being generated as an output only model.   This results in it lacking a public constructor, which is needed to be able to pass it as a parameter to `CreateOutputTextPart`.   This should have been identified as an input/output model.     We'll look at getting this fixed up.

ShivangiReja:
Hi @lmmartinez97, thanks again for raising this! We’ve merged a [PR with the fix](https://github.com/openai/openai-dotnet/pull/649).

The derived types of `ResponseMessageAnnotation` are now public, so you can create instances and pass them into `CreateOutputTextPart`. This update will be included in the next SDK release.

Really appreciate your patience and feedback!

PR #649, opened 2025-09-02
Add derive types for ResponseMessageAnnotation

ShivangiReja:
This change makes `ResponseMessageAnnotation` derived classes public so they can be used as input when constructing items such as `ResponseContentPart`. This resolves the limitation where annotations could only be treated as output and could not be passed into APIs requiring input annotations.

Issue: https://github.com/openai/openai-dotnet/issues/447


api/OpenAI.net8.0.cs
```diff
@@ -5622,6 +5634,18 @@ public class StreamingResponseWebSearchCallSearchingUpdate : StreamingResponseUp
         protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
     }
     [Experimental("OPENAI001")]
+    public class UrlCitationAnnotation : ResponseMessageAnnotation, IJsonModel<UrlCitationAnnotation>, IPersistableModel<UrlCitationAnnotation> {
```
joseharriaga: Rename this to `UriCitationAnnotation`

api/OpenAI.net8.0.cs
```diff
@@ -5622,6 +5634,18 @@ public class StreamingResponseWebSearchCallSearchingUpdate : StreamingResponseUp
...
+        public UrlCitationAnnotation(Uri url, int startIndex, int endIndex, string title);
+        public int EndIndex { get; set; }
+        public int StartIndex { get; set; }
+        public string Title { get; set; }
+        public Uri Url { get; set; }
```
joseharriaga: Rename this property to `Uri`

joseharriaga (approved):

src/Custom/Responses/FileCitationAnnotation.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace OpenAI.Responses;
+
+// CUSTOM:
+// - Renamed.
+[CodeGenType("AnnotationFileCitation")]
+public partial class FileCitationAnnotation
```
joseharriaga: nit: Could you put the following files under a Custom/Responses/Annotations folder:
- FileCitationAnnotation.cs
- FilePathAnnotation.cs
- UriCitationAnnotation.cs
- ResponseMessageAnnotation.cs

Also, now that I think more about it, I think we should name them `FileCitationMessageAnnotation`, `FilePathMessageAnnotation` and `UriCitationMessageAnnotation` for consistency with how we name derived types. I apologize for the churn because I originally gave you the other names.

joseharriaga (approved):

PR #656, opened 2025-09-04
Add initial implementation of `McpTool` to Responses

joseharriaga:
* Added initial support for integrating with remote MCP servers via the Responses API.
* Users can add the `McpTool` to the `Tools` property of their `ResponseCreationOptions` and configure it.
* Use the `AllowedTools` property to limit which of the server tools can be called by the model.
* Use the `ToolCallApprovalPolicy` property to specify which tools require an explicit approval before being called by the model.
* Support for selecting the `McpTool` via the `ToolChoice` property is coming soon.
* Support for configuring the `McpTool` with an access token that can be used to authenticate with the remote MCP server is coming soon.
* Support for connectors is coming soon.


joseharriaga:
> LGTM - should we also create a few samples?

I added an example for Responses with MCP. I also added one for Responses with function calling because I noticed we didn't have one. I would argue that we're missing more examples for Responses. I will add a few more, but I will do it in a separate PR. 🙂

PR #666, opened 2025-09-08
Onboard to test framework

m-redding:
### General changes
- Adds the test proxy to the local set of tools
- Adds a reference to the dev feed version of Microsoft.ClientModel.TestFramework
- Added an [Ignore] attribute to all tests that were failing consistently in main, after retrying enough times to be certain it wasn't just flaky
- Migrated from NUnit 3 to NUnit 4, see https://docs.nunit.org/articles/nunit/release-notes/Nunit4.0-MigrationGuide.html for more information about the breaking changes

### Test utility changes
- Migrate from bespoke System.ClientModel mock types to use the mocks provided by Microsoft.ClientModel.TestFramework
- Added an OpenAI test environment
- Added an Open AI recorded test base:
    - Automatically handles proxying the client and options. Proxying the client is required for sync/async auto-conversion, that is how the automatic interception of async calls happens. Instrumenting the options is required for recorded tests, it injects the test proxy transport to allow for recording.
    - Adds additional sanitization

### Mock test changes
- Added auto async/sync conversion for tests and removed all logic for synchronous testing. This is done automatically by the test framework now. It will intercept any calls to asynchronous methods, and in sync mode, forwards them to the sync counterpart instead. It works for pageable APIs as well.
- Moved some additional tests into mock test files since the recordings were empty (i.e. they didn't actually make any network calls), there's a bit more work to do on this front though

### Recorded test changes
- added auto async/sync same as with mock tests
 - Added recordings for live tests
- Removed [Parallelizable] attributes - playback/record cannot be run in parallel due to constraints in NUnit, replaced with LiveParallelizable so that live test runs are still run in parallel. Since playback is significantly faster than live, this will really only slow down re-recording, which hopefully won't have to happen in bulk very often

### Areas for future improvement
- Realtime tests cannot be recorded yet due to issues with recording web sockets, need to fix these issues and then record these tests


tests/Chat/ChatSmokeTests.cs
```diff
@@ -35,14 +37,18 @@ public async Task SmokeTest()
         string mockResponseId = Guid.NewGuid().ToString();
         long mockCreated = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
 
-        BinaryData mockRequest = BinaryData.FromString($$"""
```
m-redding, line 38: I don't understand the value in injecting content through create message in the mock transport instead of just calling the client method with the data. I'm not planning to add this functionality to the test framework and the test still passes without doing this, so removing.

examples/Responses/Example02_SimpleResponseStreaming.cs
```diff
@@ -11,6 +11,7 @@ namespace OpenAI.Examples;
 
 public partial class ResponseExamples
 {
+    [Ignore("Test org cannot use streaming")]
```
m-redding, line 14: The first thing I did was run every test in main and add Ignore attributes to all failing tests so that I could effectively validate the migration did not affect existing test behavior.

tests/OpenAI.Tests.csproj
```diff
@@ -15,15 +15,23 @@
...
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
-    <PackageReference Include="NUnit" Version="3.13.2" />
-    <PackageReference Include="NUnit3TestAdapter" Version="4.4.2" />
+    <PackageReference Include="NUnit" Version="4.4.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="5.1.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
     <PackageReference Include="Moq" Version="[4.18.2]" />
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
-  </ItemGroup>
+		<PackageReference Include="Microsoft.ClientModel.TestFramework" Version="1.0.0-alpha.20250915.1" />
```
m-redding: One last to do - I need to bump this one more time to pull in a few fixes I merged yesterday once the dev feed is updated

examples/Responses/Example02_SimpleResponseStreaming.cs
```diff
@@ -11,6 +11,7 @@ namespace OpenAI.Examples;
 
 public partial class ResponseExamples
 {
+    [Ignore("Awaiting org verification.")]
```
jsquire: Wat dis?

tests/Audio/TranslationTests.cs
```diff
@@ -1,22 +1,18 @@
-﻿using NUnit.Framework;
+﻿using Microsoft.ClientModel.TestFramework;
+using NUnit.Framework;
 using OpenAI.Audio;
 using OpenAI.Tests.Utility;
 using System;
```
jsquire: Can we please sort with System namespaces first?   Someone messed these up.

tests/Utility/GenerationWithBytesResponseWorks.json
jsquire, line 1: Should this be in a session records location?

jsquire (approved):
I didn't scrutinize each test class, but poking through randoms look like the expected patterns.   We'll want to get some additional eyes on, but I think it's pretty low risk to advance this, as the tests aren't super-reliable today.

tests/Utility/GenerationWithBytesResponseWorks.json
m-redding: Yep pulled this out for a side by side to sort out an issue with image sanitization (for size reduction). Will remove

examples/Responses/Example02_SimpleResponseStreaming.cs
m-redding: This fails with "Your organization must be verified to stream this model". After spending way too much time debugging tests that were failing even before my changes (I had assumed tests were all green) I marked everything failing in main as ignore.

I can take this out though since it's an example, customers will be confused by this.

jsquire: Ah.   Yes, this is a continuing problem for testing.   @achandmsft is working on getting us permissions to the "must be validated APIs".    We should leave the `[Ignore]`, I think and just be a bit more explicit in the reason.   "Test fails because organization is still pending verification"  - or something like that.

jsquire (approved):

m-redding:
Just pushed commits that:
- re-recorded multipart tests to use the new functionality from the test proxy
- added recordings for tests that were previously ignored due to org verification
- added an "api-key" placeholder in the variables to fix some issues in playback.
- added recordings for recently merged tests

Tests are all green - lmk if there's any additional feedback!!

Issue #673, opened 2025-09-10
[FEATURE REQ] `ChatReasoningEffortLevel` is missing `Minimal` option


kurnakovv:
### Describe the feature or improvement you are requesting

According to the [official OpenAI API documentation](https://platform.openai.com/docs/api-reference/chat/create#chat-create-reasoning_effort), the `reasoning_effort` parameter currently supports four values:
* minimal
* low
* medium
* high

However, in the generated SDK (ChatReasoningEffortLevel class), only three values are available:
* Low
* Medium
* High

The Minimal option is missing, which causes inconsistencies between the documentation and SDK.

jsquire:
Hi @kurnakovv.   Thanks for reaching out and we regret that you're experiencing difficulties.   The `minimal` value was recently added to the service and has not yet been released as part of the REST API specification that the library is generated from.   Once the spec has been updated with the value, it will be added to the client.

In the meantime, you can make use of the value by dropping to the protocol level as demonstrated in [Example07_ChatProtocol.](https://github.com/openai/openai-dotnet/blob/main/examples/Chat/Example07_ChatProtocol.cs).

tomw-twinthread:
how long does it typically take to update the api spec?  

It's been over a month and a half since gpt 5 was released

jsquire:
@tomw-twinthread:  The spec updates happen upstream and then flow down to the libraries, so we do not have a great deal of insight nor influence over them.  Generally speaking, the updates lag behind the service release by a couple of weeks.   My understanding is that the gpt-5 service release caused an issue in the spec metadata that broke the tooling, which has delayed the updates.    This is currently being worked through by the folks who own the spec.

JonET:
> Hi [@kurnakovv](https://github.com/kurnakovv). Thanks for reaching out and we regret that you're experiencing difficulties. The `minimal` value was recently added to the service and has not yet been released as part of the REST API specification that the library is generated from. Once the spec has been updated with the value, it will be added to the client.
> 
> In the meantime, you can make use of the value by dropping to the protocol level as demonstrated in [Example07_ChatProtocol.](https://github.com/openai/openai-dotnet/blob/main/examples/Chat/Example07_ChatProtocol.cs).

It kind of looks like you can just use the string constructor for the `ResoningEffortLevel` instead of dropping to the protocol level while we wait for a revved library version.

```new ChatReasoningEffortLevel("minimal")```

jsquire:
[...]

PR #675, opened 2025-09-10
#673 Add ChatReasoningEffortLevel.Minimal value

kurnakovv:
#673

### DIFF
--- a/src/Custom/Chat/ChatCompletionOptions.cs
+++ b/src/Custom/Chat/ChatCompletionOptions.cs
@@ -153,7 +153,7 @@ public partial class ChatCompletionOptions
     // - Renamed.
     /// <summary>
     /// (o1 and newer reasoning models only) Constrains effort on reasoning for reasoning models.
-    /// Currently supported values are <see cref="ChatReasoningEffortLevel.Low"/>, <see cref="ChatReasoningEffortLevel.Medium"/>, and <see cref="ChatReasoningEffortLevel.High"/>.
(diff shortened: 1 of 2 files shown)


christothes (changes requested):
Thank you for your contribution @kurnakovv! 

This PR modifies generated code. Instead, it should modify the [TypeSpec spec that defines the `ReasoningEffort` type](https://github.com/openai/openai-dotnet/blob/e4963f0617581017d28ae2766621ecfa6bdb4392/specification/base/typespec/common/models.tsp#L231). 

That said, there is an updated spec in the works that already includes this value, so this change will show up as soon as we incorporate the new spec and regenerate.

jsquire:
Hi @kurnakovv.   As Chris mentioned here and I detailed in the associated issue, the value must be added to the REST API specification in order for it to appear in the client.   While we maintain a local copy of the spec in the repository, the official version is authored upstream and then flows to us.   Anything that is changed in the local spec or the C# code generated from it will be overwritten when the official spec is next released.  

Since we're unable to move forward with this contribution, I'm going to close out the PR.   Thank you, again, for your contribution and interested in improving the OpenAI developer experience.

kurnakovv:
@christothes @jsquire 
> That said, there is an updated spec in the works that already includes this value, so this change will show up as soon as we incorporate the new spec and regenerate.

Thank you very much for your answer, and could you tell me where I can look at this? I just don't quite understand where it changes (I didn't find a branch/PR with changes)
Or is it that it changes in a different way?

If anything, then I am ready to change the code that you indicated in the link (but most likely not, because you said that the code has already changed somewhere)

christothes:
The place I linked is correct, however we are in the process updating a private version of the spec which should be published soon. In the meantime, if you are curious about the process for updating the spec, you could add the new minimal value to the type I linked and then run the [Invoke-CodeGen](https://github.com/openai/openai-dotnet/blob/main/scripts/Invoke-CodeGen.ps1) script.

PR #676, opened 2025-09-10
Fix README link to Observability doc (match filename case)

1w2w3y:
Update README.md to point to ./docs/Observability.md (capitalized) so the link matches the actual filename and avoids broken links on case-sensitive filesystems.


joseharriaga:
Thank you for your contribution, @1w2w3y ! Approved and merged. 🙂

PR #681, opened 2025-09-11
Update file-based apps to be aligned with .NET 10 RC1

justinyoo:
This PR is mainly for fixing two things:

- Updated the `#:property` directive to be working with .NET 10 RC1.
- Updated the OpenAI SDK package to use `2.*` instead of `2.2.*-*`.
- Added the `OPENAI001` warning suppressions for some methods.

### DIFF
--- a/docs/guides/global.json
+++ b/docs/guides/global.json
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "allowPrerelease": true
+    "version": "10.*-rc*"
   }
 }


docs/guides/global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "10.*-preview"
+    "allowPrerelease": true
```
KrzysztofCwalina: What's the reason for this change? I am not sure, but I think this only works if 10 is already installed on the machine and the setting simply allows dotnet.exe to pick a prerelease version of .net (10 in this case). The before setting was specifically asking for a preview version of 10 and so running these samples would fail if 10 is not installed.

justinyoo: Now, we're in the RC phase, 10.0.100-rc.1.25451.107. There's no `-preview` any longer. But you're right. I'll update it to look for `10.*-rc*`.

joseharriaga:
Thank you for your contribution, @justinyoo ! Approved and merged.

PR #685, opened 2025-09-12
Add code samples for the overview page

justinyoo:
This PR is intended to add code samples for the [overview page](https://platform.openai.com/docs/overview) including:

- developer quickstart


docs/overview/developer_quickstart.cs
```diff
@@ -0,0 +1,14 @@
+// SAMPLE: Generate text from a simple prompt
```
KrzysztofCwalina: should this mention that it's the Responses API?

docs/overview/developer_quickstart.cs
```diff
@@ -0,0 +1,14 @@
...
+// GUIDANCE: Instructions to run this code: https://aka.ms/oai/net/start
+#pragma warning disable OPENAI001
+
+#:package OpenAI@2.*
```
KrzysztofCwalina: should this use 2.4?

docs/overview/developer_quickstart.cs
justinyoo: That's good idea. I'll update it.

docs/overview/developer_quickstart.cs
justinyoo: It currently uses 2.4 - the latest version. Should we exclusively pin to `2.4` or loosely point to `2.*` so that it can be more flexible for future version update?

docs/overview/developer_quickstart.cs
justinyoo: Updated

PR #686, opened 2025-09-12
Expose CodeInterpreterTool for responses

### LINKED ISSUE #448: Support for code interpreter tool in Resonses API
### Confirm this is a feature request for the .NET library and not the underlying OpenAI API

- [x] This is a feature request for the .NET library

Are there any plans to support the **code interpreter** tool in the Responses API?
Hopefully it's already supported and I simply overlooked it — that would be great!

### Additional context

_No response_

christothes:
closes #448


src/Custom/Responses/Tools/ResponseTool.cs
```diff
@@ -74,4 +74,42 @@ public static McpTool CreateMcpTool(string serverLabel, Uri serverUri, IDictiona
...
+
+    // CUSTOM: Added factory method a a convenience.
+    /// <summary>
+    /// Creates a new instance of the <see cref="CodeInterpreterTool"/> class with a specified container ID.
+    /// </summary>
+    /// <param name="containerId">The Id of a previously created container</param>
+    /// <returns></returns>
+    public static CodeInterpreterTool CreateCodeInterpreterTool(string containerId)
+    {
+        Argument.AssertNotNull(containerId, nameof(containerId));
```
jsquire: Should we guard against empty here as well?  That doesn't seem a reasonable thing to use.

api/OpenAI.net8.0.cs
```diff
@@ -4590,6 +4590,15 @@ public class TurnDetectionOptions : IJsonModel<TurnDetectionOptions>, IPersistab
...
 namespace OpenAI.Responses {
+    [Experimental("OPENAI001")]
+    public class CodeInterpreterTool : ResponseTool, IJsonModel<CodeInterpreterTool>, IPersistableModel<CodeInterpreterTool> {
+        public CodeInterpreterTool(BinaryData container);
+        public BinaryData Container { get; set; }
+        protected override ResponseTool JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected override ResponseTool PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
```
joseharriaga, line 4661: Support for a tool in the Responses API is divided in four parts:
1. There is the tool itself: A class that derives from `ResponseTool`, which is what you have here.
2. The tool-related items: Classes that derive from `ResponseItem`. In the case of Code Interpreter, there is only one, which we can call `CodeInterpreterToolCallResponseItem`.
3. The tool-related streaming events: Classes that derive from `StreamingResponseUpdate`. In the case of Code Interpreter, there are five:
    1. "response.code_interpreter_call.in_progress" which we can call `StreamingResponseCodeInterpreterCallInProgressUpdate`
    2. "response.code_interpreter_call.interpreting" which we can call `StreamingResponseCodeInterpreterCallInterpretingUpdate`
    3. "response.code_interpreter_call.completed" which we can call `StreamingResponseCodeIntepreterCallCompletedUpdate`
    4. "response.code_interpreter_call_code.delta" which we can call `StreamingResponseCodeInterpreterCallCodeDeltaUpdate`
    5. "response.code_interpreter_call_code.done" which we can call `StreamingResponseCodeInterpreterCallCodeDoneUpdate`
4. The tool-related tool choice. Note that I have a plan to re-design tool choices a little bit. This part is not strictly necessary for Code Interpreter to be usable, so I think it's okay to skip it for now.

My recommendation is that we include parts 1, 2 and 3 in the same release. As such, I suggest that we include them all together as part of this PR if possible. What do you think?

api/OpenAI.net8.0.cs
```diff
@@ -4590,6 +4590,15 @@ public class TurnDetectionOptions : IJsonModel<TurnDetectionOptions>, IPersistab
     }
 }
 namespace OpenAI.Responses {
+    [Experimental("OPENAI001")]
+    public class CodeInterpreterTool : ResponseTool, IJsonModel<CodeInterpreterTool>, IPersistableModel<CodeInterpreterTool> {
+        public CodeInterpreterTool(BinaryData container);
+        public BinaryData Container { get; set; }
```
joseharriaga: The value of this property is a non-discriminated union between a plain string and object that looks like this:

```json
{
    "type": "auto",
    "file_ids: []
}
```

Because .NET cannot represent unions, this was generated as `BinaryData`, but we should customize it to offer a more appropriate experience. With this in mind, I recommend the following design:

```csharp
public class CodeInterpreterTool : ResponseTool
{
    public CodeInterpreterTool(CodeInterpreterContainer container);
    public CodeIntepreterContainer Container { get; }
}

// This class represents the union.
public class CodeInterpreterContainer
{
    public CodeInterpreterContainer(string containerId);
    public CodeInterpreterContainer(CodeInterpreterContainerConfiguration containerConfiguration);

    // Only one of these two properties will be non-null at a time.
    public string ContainerId { get; }
    public CodeInterpreterContainerConfiguration ContainerConfiguration { get; }
}

// Base class
public class CodeInterpreterContainerConfiguration
{
    public static AutomaticCodeInterpreterContainerConfiguration CreateAutomaticConfiguration(IEnumerable<string> fileIds = null);
}

// Per the REST API, there is only one derived type for now
public class AutomaticCodeInterpreterContainerConfiguration : CodeInterpreterContainerConfiguration
{
    public AutomaticCodeInterpreterContainerConfiguration();
    public IList<string> FileIds { get; }
}

// With this design, ResponseTool has a single factory method.
public class ResponseTool
{
    ...
    public CodeInterpreterTool CreateCodeInterpreterTool(CodeInterpreterContainer container);
    ...
}

```

For reference, check out how I handled `MCpToolCallApprovalPolicy`, which is a non-discriminated union between an enum and an object:
🔗 https://github.com/openai/openai-dotnet/blob/b60d65838954396ed8edecb847c8172b014c756a/src/Custom/Responses/Tools/McpTool/McpToolCallApprovalPolicy.cs

Here's the corresponding custom serialization code that forwards the serialization logic to the individual components of the union:
🔗 https://github.com/openai/openai-dotnet/blob/b60d65838954396ed8edecb847c8172b014c756a/src/Custom/Responses/Tools/McpTool/McpToolCallApprovalPolicy.Serialization.cs

You will need to define a custom model in TypeSpec to represent the union like I did here:
🔗 https://github.com/openai/openai-dotnet/blob/b60d65838954396ed8edecb847c8172b014c756a/specification/client/models/responses.models.tsp#L64-L67

joseharriaga (changes requested):

api/OpenAI.net8.0.cs
jsquire: The distinction between `string` and `object` are represented in Chris' changes by the different method overloads and should produce the correct payload in each case.   I don't know that I agree we need to add more classes to the hierarchy for this.   I'd be very interested in the rationale behind the feedback in this case.   My assumption is that I must be overlooking something, but I don't see it.

trrwilson: Can we do both?

I agree with @jsquire that the creation semantics are unnecessarily complicated if we require the intermediate layer; this feels right:

```csharp
class CodeInterpreterTool : ResposeTool {
  public CodeInterpreterTool(string containerId) {}
  public CodeInterpreterTool(IEnumerable<string> fileIds) {}
  // Leaves plenty of room for this without it seeming awkward
  public CodeInterpreterTool(CodeInterpreterContainer containerDefinition);
}
```

But I agree with @joseharriaga that `BinaryData` on the access side feels wrong. In general, we should try not to use `BinaryData` at the convenience layer unless it really is representing arbitrary data -- which, in this case, it isn't.

In some earlier surfaces, we've used "squishing" to hide this -- which has yielded mixed results for maintainability vs. usability tradeoffs. Having `ContainerId` and `ContainerFileIds` directly exposed as mutually exclusive properties would be "OK" right now, but if two or three more container types are added with even more possibilities, it'd get messy in a hurry.

If we do both, we can have strongly typed access (with the truer-to-JSON hierarchy) without forcing it into the (frankly more common) creation path.

```csharp
class CodeInterpreterTool : ResponseTool {
  // This is always populated when a CodeInterprerTool is created
  public CodeInterpreterContainer Container { get; }

  // These are how we expect people to make them for the foreseeable future
  public CodeInterpreterTool(string containerId) {}
  public CodeInterpreterTool(IEnumerable<string> fileIds) {}

  // This is the future-proofed extensible ctor (could even be hidden for now)
  public CodeInterpreterTool(CodeInterpreterContainer containerDefinition);
}
```

This makes it look a lot more consistent with our reuse of a single input/output model for the tools, too -- I need to double-check, but I think the specification misses that the "automatic container definition" is really just an input-only concept that then creates a container for you, like vector store creation helpers in Assistants. If that's the case (i.e., if trying to get a list of file IDs from an `OpenAIResponse.Tools` coming back from the model would never make sense) then just having a `.Container` property when you're not defining the container yourself seems right.

That is, I'm pretty sure that if I provide

```json
{
  "tools": [{
      "type": "code_interpreter",
      "container": {"type": "auto", "file_ids": ["file_1", "file_2"]}
    }]
}
```

...then the response will just tell me what new container (that it just created for me) it's using:

```json
{
  "tools": [{
    {"type": "code_interpreter", "container": "container_id_just_created"}
  }]
}
```

There's a general challenge with the usability of these polymorphic hierarchies ("how do I know that I need to cast `CodeInterpreterContainer` to `AutomaticCodeInterpreterContainer`?") but that's a much broader problem with a more systemic (and still additive) piece of help needed.

joseharriaga: > I don't know that I agree we need to add more classes to the hierarchy for this.  I'd be very interested in the rationale behind the feedback in this case. 

@jsquire : Yes, let me provide some more context about why I think it is important to have a class.

1. First, `CodeInterpreterTool` is both an input and an output. While the factory methods are helpful for writing the `BinaryData` in the input scenario, having to manually parse the `BinaryData` in the output scenario is not a good experience. By contrast, both scenarios can be solved simultaneously by adding a proper class with its constructors and properties.
2. Additionally, I would like to note that the factory methods in the base `ResponseTool` class are meant specifically for discoverability. The idea is that a new user would see that they can pass an array of `ResponseTool` objects as an input, but they won't yet know what the possible `ResponseTool` types are. The factory methods help bridge this a little bit; it's a way for the base class to reveal the derived classes that are available. Following that rationale, these factory methods should be thought of as an additional convenience (a "nice to have"), and not a substitute for the "normal" way of doing things. By this, I'm referring to the fact that the typical .NET developer expects to be able to simply `new` objects up.
3. We should aim to reserve `BinaryData` for handling raw data (i.e., raw bytes, arbitrary JSON, etc.). Consistency here will help train users on how to deal with it when they encounter it.
4. Finally, the design that I illustrated above is the result of applying the lessons that we have learned while dealing with other non-discriminated unions, and it is something that I think we should aim to apply consistently every time we see one of these unions going forward. I think it's versatile enough to handle different scenarios (inputs and outputs, unions with two components and unions with twenty components, etc.), and I think the resulting experience is "good enough". Having said that, I'm always interested in hearing alternatives. Ultimately, my goal is a standard design that works everywhere, so that we can someday generate it using the generator plugin.

> Can we do both?

@trrwilson : Yes, maybe. I'm not strictly opposed to including a constructor in `CodeInterpreterTool` that takes the file IDs directly, but let me mention a couple of things:
1. I think it's okay to add this constructor as a "nice to have" convenience as long as the other parts are still there, which I think is what you're saying too, so we're good 🙂. I just want to emphasize this because, like I mentioned above, my ultimate goal is to arrive at a standard design, so that we can stop having to design every union separately and therefore boost our productivity.
2. One small issue with this convenience constructor in particular is that, while it is certainly less verbose to use, it also doesn't make it very clear what it is doing. By this, I mean that it's not very obvious that a new container is automatically being created and that the files are being added to it.

```csharp
ResponseCreationOptions options = new()
{
    Tools = new CodeInterpreterTool([ "file_1", "file_2" ]); // What is happening to these files?
}
```

The approach that I'm proposing for this scenario is certainly more verbose, but I think it's a little clearer what is going on.

```csharp
AutomaticCodeInterpreterContainerConfiguration automaticContainerConfig = new()
{
    FileIds = { "file_1", "file_2" }
};

CodeInterpreterContainer container = new(automaticContainerConfig);

ResponseCreationOptions options = new()
{
    Tools = new CodeInterpreterTool(container);
}
```

If it helps make this even clearer, I'm open to considering different names too. For example, we could rename `AutomaticCodeInterpreterContainerConfiguration` to simply be `AutomaticContainerConfiguration`.

We could also add the convenience that you're proposing to `CodeInterpreterContainer` instead of `CodeInterpreterTool`:

```csharp
public static CodeInterpreterContainer CreateAutomaticContainer(IEnumerable<string fileIds>);
```

So the end result would look like this:

```csharp
CodeInterpreterContainer container = CodeInterpreterContainer.CreateAutomaticContainer(["file_1", "file_2"]);

ResponseCreationOptions options = new()
{
    Tools = new CodeInterpreterTool(container);
}
```

api/OpenAI.net8.0.cs
christothes: Thanks for pointing this out. I'll add 1-3.

specification/base/typespec/responses/models.tsp
```diff
@@ -2593,10 +2599,10 @@ model ResponseCodeInterpreterCallInProgressEvent extends ResponseStreamEvent {
   type: ResponseStreamEventType.response_code_interpreter_call_in_progress;
 
   /** The index of the output item that the code interpreter call is in progress. */
-  output_index: int32;
+  output_index: integer;
```
joseharriaga: Should we keep this as an `int32` (at least for now) for consistency with other streaming classes and also based on it being conceptually an "index"?

api/OpenAI.net8.0.cs
```diff
@@ -4602,6 +4602,56 @@ public class TurnDetectionOptions : IJsonModel<TurnDetectionOptions>, IPersistab
...
+        public string ContainerId { get; set; }
+        public IList<BinaryData> Results { get; }
+        public CodeInterpreterCallStatus Status { get; set; }
+        protected override ResponseItem JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected override ResponseItem PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public readonly partial struct CodeInterpreterCallStatus : IEquatable<CodeInterpreterCallStatus> {
```
joseharriaga: For consistency with other tool-related items, could you make this enum a CLR enum instead of an extensible enum? Here is an example of what I mean:
🔗 https://github.com/openai/openai-dotnet/blob/main/src/Custom/Responses/Items/FunctionTool/FunctionCallStatus.cs

We do have an ongoing discussion about if these should be CLR enums or extensible enums, but in the meantime, while we arrive at a consensus, I think it makes sense to keep them consistent.

api/OpenAI.net8.0.cs
```diff
@@ -4602,6 +4602,56 @@ public class TurnDetectionOptions : IJsonModel<TurnDetectionOptions>, IPersistab
...
+    public class AutomaticCodeInterpreterContainerConfiguration : CodeInterpreterContainerConfiguration {
+        public AutomaticCodeInterpreterContainerConfiguration(IEnumerable<string> fileIds = null);
+        public IList<string> FileIds { get; }
+    }
+    [Experimental("OPENAI001")]
+    public class CodeInterpreterCallResponseItem : ResponseItem, IJsonModel<CodeInterpreterCallResponseItem>, IPersistableModel<CodeInterpreterCallResponseItem> {
+        public CodeInterpreterCallResponseItem(CodeInterpreterCallStatus status, string code, IEnumerable<BinaryData> results);
+        public string Code { get; set; }
+        public string ContainerId { get; set; }
+        public IList<BinaryData> Results { get; }
```
joseharriaga: It looks like our spec is outdated here too. Looking at the API reference docs:
1. There is no "results" property; instead, it is called "outputs". 
2. Similarly, I also see that this was generated as `BinaryData` because it is defined as a union. However, it should be specifically defined as a discriminated union so that we can generate the appropriate class hierarchy. 

With this in mind, I recommend amending the TypeSpec with something like this:

1. In `specification/ base/typespec/responses/models.tsp`, rename the "results" property of the `CodeInterpreterToolCallItemBase` alias to "outputs":
🔗 https://github.com/openai/openai-dotnet/blob/f68f1628ae6dfebc4ac02a3f44dbfb707334cc43/specification/base/typespec/responses/models.tsp#L2479

2. In the same file, delete the union called `CodeInterpreterToolOutput`.

2. Then, in `specification/base/typespec/responses/custom.tsp`, add the following union and discriminated base model:
```typespec
union CodeInterpreterToolOutputType {
    string,
    logs: "logs",
    image: "image"
}

@discriminator("type")
model CodeInterpreterToolOutput {
  @doc("The type of the output.")
  type: CodeInterpreterToolOutputType;
}
```

3. Then, going back to `specification/base/typespec/responses/models.tsp`, delete the model called `CodeInterpreterTextOutput` and add a new model called `CodeInterpreterToolLogsOutput` that extends the new discriminated base model like this: 

```typespec
model CodeInterpreterToolLogsOutput extends CodeInterpreterToolOutput {
  @doc("The type of the output. Always 'logs'.")
  type: CodeInterpreterToolOutputType.logs;

  @doc("The logs output from the code interpreter.")
  logs: string;
}
```

4. In the same file, also delete the model called `CodeInterpreterFileOutput` and add a new model called `CodeInterpreterToolImageOutput` that extends the new discriminated base model like this:

```typespec
model CodeInterpreterToolImageOutput extends CodeInterpreterToolOutput {
  @doc("The type of the output. Always 'image'.")
  type: CodeInterpreterToolOutputType.image;

  @doc("The URL of the image output from the code interpreter.")
  url: url;
}
```

5. Regenerate the code.

specification/base/typespec/responses/models.tsp
christothes: We can - I just thought this would eventually change anyway.

api/OpenAI.net8.0.cs
christothes: I made the changes. It appears that the value of 
```
outputs: CodeInterpreterToolOutput[];
```
is always null in the response. Is that expected?

api/OpenAI.net8.0.cs
```diff
@@ -4602,6 +4602,98 @@ public class TurnDetectionOptions : IJsonModel<TurnDetectionOptions>, IPersistab
     }
 }
 namespace OpenAI.Responses {
+    public class AutomaticCodeInterpreterContainerConfiguration : CodeInterpreterContainerConfiguration {
+        public AutomaticCodeInterpreterContainerConfiguration(IEnumerable<string> fileIds = null);
+        public IList<string> FileIds { get; }
+    }
+    [Experimental("OPENAI001")]
+    public class CodeInterpreterCallResponseItem : ResponseItem, IJsonModel<CodeInterpreterCallResponseItem>, IPersistableModel<CodeInterpreterCallResponseItem> {
+        public CodeInterpreterCallResponseItem(CodeInterpreterCallStatus status, string code);
```
joseharriaga: Contrary to the other properties, the `Status` is not required (except when this class is used as an output) and is read-only. To achieve this, could you add the following customization in the TypeSpec client.tsp?

```typespec
@@visibility(CodeInterpreterToolCallItemResource.status, Lifecycle.Read);
```

You will notice other tools have the same customization:
🔗 https://github.com/openai/openai-dotnet/blob/main/specification/client/responses.client.tsp#L10

Making the `status` property read-only like this should have the effect of removing the `status` parameter from this constructor.

api/OpenAI.net8.0.cs
```diff
@@ -4602,6 +4602,98 @@ public class TurnDetectionOptions : IJsonModel<TurnDetectionOptions>, IPersistab
...
+        public string Code { get; set; }
+        public string ContainerId { get; set; }
+        public IList<CodeInterpreterToolOutput> Outputs { get; }
+        public CodeInterpreterCallStatus Status { get; set; }
```
joseharriaga: Because the `Status` property is only required when used as an output, it should be nullable when used as an input. To reflect this, could you add the following customization in the C# code:

```csharp
// CUSTOM: Made nullable since this is a read-only property.
[CodeGenMember("Status")]
public CodeInterpreterCallStatus? Status { get; }
```

You will notice that we do the same for the other tools:
🔗 https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/src/Custom/Responses/Items/FunctionTool/FunctionCallResponseItem.cs#L5

If you're curious about why the `Status` property is not generated correctly based on this comment and my previous, this is an issue in the TypeSpec that comes from the original swagger spec. I need to confirm if this issue is still present in the new version that Travis and his team are working on, but for now we can unblock ourselves with these simple customizations.

specification/base/typespec/responses/models.tsp
joseharriaga: That's true. Although based on what Krzysztof was saying, we will still want certain properties to be generated as `int` instead of `long` for usability reasons, and he mentioned indeces as an example. So we can make it `integer` here, but then we need to customize it in client.tsp using the `@alternateType` decorator to be an `int32`.

api/OpenAI.net8.0.cs
```diff
@@ -4602,6 +4602,98 @@ public class TurnDetectionOptions : IJsonModel<TurnDetectionOptions>, IPersistab
...
+        protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected override ResponseItem PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public readonly partial struct CodeInterpreterCallStatus : IEquatable<CodeInterpreterCallStatus> {
+        public CodeInterpreterCallStatus(string value);
+        public static CodeInterpreterCallStatus Completed { get; }
+        public static CodeInterpreterCallStatus InProgress { get; }
+        public static CodeInterpreterCallStatus Interpreting { get; }
```
joseharriaga: I'm reading the API reference docs, and it appears that there two possibles values missing: "incomplete" and "failed". Could you manually add them here?
🔗 https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/specification/base/typespec/responses/custom/items.tsp#L175

I have confirmed that the new TypeSpec that Travis's team is working on already has this fix, so we won't an issue integrating that part.

EDIT: Having said that, it's weird that neither the docs nor the spec have corresponding streaming events for these other two possible values, while they do have events for "completed", "in_progress", and "interpreting". Could these events be missing? Or could it be that the "incomplete" and "failed" status don't actually exist? 🤔

api/OpenAI.net8.0.cs
```diff
@@ -4602,6 +4602,98 @@ public class TurnDetectionOptions : IJsonModel<TurnDetectionOptions>, IPersistab
...
+        public CodeInterpreterContainer Container { get; }
+        protected override ResponseTool JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected override ResponseTool PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class CodeInterpreterToolImageOutput : CodeInterpreterToolOutput, IJsonModel<CodeInterpreterToolImageOutput>, IPersistableModel<CodeInterpreterToolImageOutput> {
+        public CodeInterpreterToolImageOutput(Uri url);
+        public Uri Url { get; set; }
```
joseharriaga: nit: Could you rename this to `ImageUri` in responses.client.tsp?

```typespec
@@clientName(CodeInterpreterToolImageOutput.url, "ImageUri");
```

src/Custom/Responses/GeneratorStubs.cs
```diff
@@ -0,0 +1,7 @@
+namespace OpenAI.Responses;
+
+[CodeGenType("CodeInterpreterToolImageOutput")] public partial class CodeInterpreterToolImageOutput {}
+[CodeGenType("CodeInterpreterToolLogsOutput")] public partial class CodeInterpreterToolLogsOutput {}
+[CodeGenType("UnknownCodeInterpreterToolOutput")] public partial class UnknownCodeInterpreterToolOutput {}
+[CodeGenType("CodeInterpreterToolOutput")] public partial class CodeInterpreterToolOutput {}
+[CodeGenType("CodeInterpreterToolOutputType")] public readonly partial struct CodeInterpreterToolOutputType {}
```
joseharriaga: `UnknownCodeInterpreterToolOutput` and `CodeInterpreterToolOutputType` should remain `internal`, their names prefixed with "Internal*", and in the GeneratorStubs.cs under the Custom/Responses/Internal folder.

src/Custom/Responses/Tools/AutomaticCodeInterpreterContainerConfiguration.cs
```diff
@@ -0,0 +1,31 @@
+using System.Collections.Generic;
+
+namespace OpenAI.Responses;
+
+/// <summary>
+/// Represents a configuration for an automatic Code Interpreter container.
+/// </summary>
+public class AutomaticCodeInterpreterContainerConfiguration : CodeInterpreterContainerConfiguration
```
joseharriaga: This one should be a partial linked to the generated class here:
🔗 https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/src/Generated/Models/Responses/InternalCodeInterpreterToolAuto.cs

You will need to do the same thing you did with the `Outputs` of the `CodeInterpreterCallResponseItem` to generate the hierarchy (although there is only one derived class for now).

1. In `specification/base/typespec/responses/custom.tsp`, add the following union and discriminated base model:

```typespec
union CodeInterpreterContainerConfigurationType {
    string,
    auto: "auto"
}

@discriminator("type")
model CodeInterpreterContainerConfiguration {
  @doc("The type of the output.")
  type: CodeInterpreterContainerConfigurationType;
}
```

2. In `specification/base/typespec/responses/models.tsp`, edit the `CodeInterpreterToolAuto` model like this:

```typespec
model CodeInterpreterToolAuto extends CodeInterpreterContainerConfiguration {
  @doc("Always `auto`.")
  type: CodeInterpreterContainerConfigurationType.auto;

  /** An optional list of uploaded files to make available to your code. */
  file_ids?: string[];
}
```

3. Optionally, you can rename `CodeInterpreterToolAuto` to something like `CodeInterpreterContainerConfigurationAuto` to indicate the new hierarchy more explicitly.

src/Custom/Responses/Tools/AutomaticCodeInterpreterContainerConfiguration.cs
```diff
@@ -0,0 +1,31 @@
+using System.Collections.Generic;
+
+namespace OpenAI.Responses;
+
+/// <summary>
+/// Represents a configuration for an automatic Code Interpreter container.
+/// </summary>
+public partial class CodeInterpreterToolAuto : CodeInterpreterContainerConfiguration
```
joseharriaga: nit: Add the `[CodeGenType("CodeInterpreterToolAuto")]` attribute and keep the rename: `AutomaticCodeIntepreterContainerConfiguration`.

src/Custom/Responses/Tools/AutomaticCodeInterpreterContainerConfiguration.cs
```diff
@@ -0,0 +1,31 @@
...
+        else
+        {
+            FileIds = new List<string>();
+        }
+    }
+
+    /// <summary>
+    /// Gets the list of file IDs associated with the container.
+    /// </summary>
+    public IList<string> FileIds { get; }
```
joseharriaga: nit: I think this property can be removed from custom code and let it be generated.

src/Custom/Responses/Tools/CodeInterpreterTool.cs
```diff
@@ -0,0 +1,14 @@
...
+
+// CUSTOM: correct namespace / customize Container property.
+[CodeGenType("CodeInterpreterTool")]
+[CodeGenVisibility(nameof(CodeInterpreterTool), CodeGenVisibility.Internal, typeof(BinaryData))]
+public partial class CodeInterpreterTool
+{
+    [CodeGenMember("Container")]
+    internal BinaryData InternalContainer { get; }
+
+    public CodeInterpreterContainer Container { get; }
```
joseharriaga, line 10: I believe you can replace these two properties with the following:

```csharp
// CUSTOM:
// - Renamed.
// - Changed type from BinaryData (generated from the original union) to a custom type.
[CodeGenMember("Container")]
public CodeIntepreterContainer Container { get; }
```

This will change the type of the property from `BinaryData` to `CodeInterpreterContainer`, so you won't need to have two separate `Container` properties anymore (one internal vs. one public). Furthermore, this will also make the generated constructor of this class receive a `CodeInterpreterContainer` parameter instead of a `BinaryData` parameter, which is exactly what we want. You can then remove the `CodeGenVisibility` attribute that you added here to hide that `BinaryData` constructor and let the `CodeInterpreterContainer` constructor be public.

tests/Responses/ResponsesTests.cs
```diff
@@ -229,6 +230,289 @@ in client.CreateResponseStreamingAsync(message, responseOptions))
         Assert.That(searchItemId, Is.Not.Null.And.Not.Empty);
     }
 
+    [Test]
```
joseharriaga: nit: Could you move these tests under ResponsesToolTests.cs?

src/Custom/Responses/GeneratorStubs.cs
```diff
@@ -0,0 +1,8 @@
+namespace OpenAI.Responses;
+
+[CodeGenType("CodeInterpreterToolImageOutput")] public partial class CodeInterpreterToolImageOutput {}
+[CodeGenType("CodeInterpreterToolLogsOutput")] public partial class CodeInterpreterToolLogsOutput {}
+[CodeGenType("CodeInterpreterToolOutput")] public partial class CodeInterpreterToolOutput {}
+[CodeGenType("CodeInterpreterToolAuto")] public partial class CodeInterpreterToolAuto {}
+[CodeGenType("CodeInterpreterContainerConfiguration")] public partial class CodeInterpreterContainerConfiguration {}
+[CodeGenType("CodeInterpreterContainerConfigurationType")] readonly partial struct CodeInterpreterContainerConfigurationType {}
```
joseharriaga: nit: Prefix this `CodeInterpreterContainerConfigurationType` type with "Internal*" and move to Custom/Responses/Internal/GeneratorStubs.cs

src/Custom/Responses/Tools/CodeInterpreterContainer.cs
```diff
@@ -0,0 +1,47 @@
+using System;
+using System.Linq;
+
+namespace OpenAI.Responses;
+
+/// <summary>
+/// Represents a container for the code interpreter tool.
+/// </summary>
+public class CodeInterpreterContainer
```
joseharriaga: This class (which represents a non-discriminated union) should go through the generator like any other of our models.

Consider the example of `McpToolCallApprovalPolicy`:

1. First, define a custom model in the TypeSpec so that the class gets generated:
🔗 https://github.com/openai/openai-dotnet/blob/main/specification/client/models/responses.models.tsp#L64
Note that models that we add to the TypeSpec for .NET-specific reasons (such as this one) are prefixed with "DotNet*".

2. Then add the custom constructors and ensure that the properties don't have setters (since this is a union, they can only be set via the constructors):
🔗 https://github.com/openai/openai-dotnet/blob/main/src/Custom/Responses/Tools/McpTool/McpToolCallApprovalPolicy.cs
This is basically what you have here already. 🙂 Note that all our IJsonModels need a parameterless constructor. I believe one will be generated once this class is hooked to codegen, but just make sure to make it `internal` if it is not already

3. Finally, create a `*.Serialization.cs` file and customize the serialization and deserialization code to do the right thing depending on what component of the union we're dealing with:
🔗 https://github.com/openai/openai-dotnet/blob/main/src/Custom/Responses/Tools/McpTool/McpToolCallApprovalPolicy.Serialization.cs
In the example with `McpToolCallApprovalPolicy`, notice how we check if we're dealing with a string or an object and act based on that. 

Bonus: Check out the smoke tests that we have to validate this custom serialization and deserialization logic:
🔗 https://github.com/openai/openai-dotnet/blob/main/tests/Responses/ResponsesSmokeTests.cs#L194
The test methods are called `SerializeMCPToolCallPolicyApprovalAsString` and `SerializeMCPToolCallApprovalPolicyAsObject`. They validate that we can serialize and deserialize this type regardless of if it is a plain string and or an object.

PR #690, opened 2025-09-17
Update the getting started document with .NET 10

justinyoo:
There are more sample codes coming under the `docs` directory. Therefore, it makes more sense to move the guide doc from `docs/guides` to `docs`.

This PR is:

- To move the getting started doc from `docs/guides` to `docs`.
- To add Mac OS support.
- To make the existing doc more consistent in headings.


docs/guides/global.json
```diff
@@ -1,5 +0,0 @@
-{
```
KrzysztofCwalina, line 1: why is this file being removed? I dont think we can remove it till .NET 10 GAs. It's in rc now

justinyoo: It's actually moved to the `docs` directory because the `docs` directory covers more samples.

jsquire: @KrzysztofCwalina:  Confirming.    
https://github.com/openai/openai-dotnet/blob/main/docs/global.json

Issue #41, opened 2024-06-11
Remove AllowUnsafeBlocks once no longer required


stephentoub:
https://github.com/openai/openai-dotnet/pull/33 adds AllowUnsafeBlocks into the netstandard2.0 build in support of the polyfill for System.Net.ServerSentEvents. Once that polyfill is removed and a package reference added to System.Net.ServerSentEvents, the use of AllowUnsafeBlocks can be removed.

PR #693, opened 2025-09-18
Replace polyfill for System.Net.ServerSentEvents with real package

stephentoub:
Closes https://github.com/openai/openai-dotnet/issues/41


src/OpenAI.csproj
```diff
@@ -82,6 +75,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
     <PackageReference Include="System.ClientModel" Version="1.6.1" />
+    <PackageReference Include="System.Net.ServerSentEvents" Version="9.0.0" />
```
jsquire:
```suggestion
    <PackageReference Include="System.Net.ServerSentEvents" Version="9.0.9" />
```
I'm assuming that we should be using the latest?

jsquire:
Thanks, @stephentoub!

PR #695, opened 2025-09-18
Add code samples for web search

justinyoo:
This PR is intended to add code samples for the [Web search](https://platform.openai.com/docs/guides/tools-web-search?api-mode=responses) page including:

- Quickstart using web search
- Web search by specifying user location

This PR doesn't include the sample codes for:

- Web search by specifying sources

### DIFF
--- a/docs/guides/tools-web-search/responses/quickstart.cs
+++ b/docs/guides/tools-web-search/responses/quickstart.cs
@@ -1,0 +1,25 @@
+// SAMPLE: Get information from web search through Responses API
+// PAGE: https://platform.openai.com/docs/guides/tools-web-search?api-mode=responses
+// GUIDANCE: Instructions to run this code: https://aka.ms/oai/net/start
+#pragma warning disable OPENAI001
+
+#:package OpenAI@2.*
+#:property PublishAot=false
+
+using OpenAI.Responses;
+
+string key = Environment.GetEnvironmentVariable("OPENAI_API_KEY")!;
+OpenAIResponseClient client = new(model: "gpt-5", apiKey: key);
+
+ResponseCreationOptions options = new();
+options.Tools.Add(ResponseTool.CreateWebSearchTool());
+
+OpenAIResponse response = (OpenAIResponse)client.CreateResponse([
+    ResponseItem.CreateUserMessageItem([
+        ResponseContentPart.CreateInputTextPart(
+            "What was a positive news story from today?"
+        )
+    ])
+], options);
+
+Console.WriteLine(response.GetOutputText());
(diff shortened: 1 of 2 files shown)


docs/guides/tools-web-search/responses/quickstart.cs
christothes, line 1: Is this an intentional duplicate of [this](https://github.com/openai/openai-dotnet/pull/698/files#diff-a43606b9a60268f56fcd3672e8545c2d0c5bb41194ba5d2b0f3a1a3b24bdbb1dR15-R24)?

docs/guides/tools-web-search/responses/user-location.cs
```diff
@@ -0,0 +1,31 @@
...
+        country: "GB",
+        city: "London",
+        region: "London"
+    )
+));
+
+OpenAIResponse response = (OpenAIResponse)client.CreateResponse([
+    ResponseItem.CreateUserMessageItem([
+        ResponseContentPart.CreateInputTextPart(
+            "What are the best restaurants around Granary Square?"
```
christothes: It seems like this would be a better example if the `region` above was "Granary Square" and this prompt was changed to "What are the best restaurants near me?"

achandmsft: Yes @christothes, this is intentional duplication in order to mimick the structure of openai.com API documentation to make it easy for developers to discover samples from multiple entry points.

docs/guides/tools-web-search/responses/quickstart.cs
achandmsft: Yes @christothes, this is intentional duplication in order to mimick the structure of openai.com API documentation to make it easy for developers to discover samples from multiple entry points.

docs/guides/tools-web-search/responses/user-location.cs
achandmsft: Agree, but again we are trying to mimick the OpenAI code in other languages identically, so developers can see 1:1 mapping of the code across languages: https://platform.openai.com/docs/guides/tools-web-search?api-mode=responses#user-location

christothes: Let's change it everywhere then 😄

christothes:
```suggestion
        region: "Granary Square"
    )
));

OpenAIResponse response = (OpenAIResponse)client.CreateResponse([
    ResponseItem.CreateUserMessageItem([
        ResponseContentPart.CreateInputTextPart(
            "What are the best restaurants near me?"
```

achandmsft (commented):
Looks good, were you able to see if this actually gives you localized prompt response within this region @christothes ?

christothes:
> Looks good, were you able to see if this actually gives you localized prompt response within this region @christothes ?

I just tried this locally and the agent asked for my location as a follow up question 😮 - Looking into it now.

christothes:
> > Looks good, were you able to see if this actually gives you localized prompt response within this region @christothes ?
> 
> I just tried this locally and the agent asked for my location as a follow up question 😮 - Looking into it now.

OK - I confirmed that it works. The model must be gpt-5 though, which is consistent with this sample.

PR #697, opened 2025-09-19
Add support for authorization token and connectors to `McpTool`

joseharriaga:


joseharriaga:
> Can we add a test or sample that uses an auth token?

Yes, I added an example that uses auth based on the documentation here:
🔗 https://platform.openai.com/docs/guides/tools-connectors-mcp#authentication

PR #701, opened 2025-09-22
Add code samples for the "images and vision" page

justinyoo:
This PR is intended to add code samples for the [Images and vision](https://platform.openai.com/docs/guides/images-vision?api-mode=responses) page including:

- Generate or edit images
- Analyze images by passing an image URL
- Analyze images by passing base64 encoded string
- Analyze images by passing a file

### DIFF
--- a/docs/guides/images-vision/responses/analyze_images_passing_file.cs
+++ b/docs/guides/images-vision/responses/analyze_images_passing_file.cs
@@ -1,0 +1,30 @@
+// SAMPLE: Analyzes file from a file upload through Responses API
+// PAGE: https://platform.openai.com/docs/guides/images-vision?api-mode=responses&format=file#analyze-images
+// GUIDANCE: Instructions to run this code: https://aka.ms/oai/net/start
+#pragma warning disable OPENAI001
+
+#:package OpenAI@2.*
+#:property PublishAot=false
+
+using OpenAI.Files;
+using OpenAI.Responses;
+
+string key = Environment.GetEnvironmentVariable("OPENAI_API_KEY")!;
+OpenAIResponseClient client = new(model: "gpt-5", apiKey: key);
+
+using var http = new HttpClient();
+
+// Download an image as stream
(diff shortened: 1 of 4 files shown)


justinyoo:
It's a draft PR because not all the features are yet to released.

docs/guides/images-vision/responses/analyze_images_passing_file.cs
```diff
@@ -0,0 +1,32 @@
...
+// Download an image as stream
+using var stream = await http.GetStreamAsync(imageUrl);
+
+OpenAIFileClient files = new(key);
+OpenAIFile file = await files.UploadFileAsync(BinaryData.FromStream(stream), filename, FileUploadPurpose.Vision);
+
+OpenAIResponse response = (OpenAIResponse)client.CreateResponse([
+    ResponseItem.CreateUserMessageItem([
+        ResponseContentPart.CreateInputImagePart(file.Id),
+        ResponseContentPart.CreateInputTextPart("what's in this image?")
```
christothes: nit: the previous samples order the two input parts like this - we should be consistent. 

```suggestion
        ResponseContentPart.CreateInputTextPart("what's in this image?"),
        ResponseContentPart.CreateInputImagePart(file.Id)
```

christothes (approved):

achandmsft:
@christothes cc @ShivangiReja I am assuming this is not yet ok to merge is that correct? I thought the current (1.5) release does not support this? cc @jsquire

docs/guides/images-vision/responses/analyze_images_passing_file.cs
justinyoo: Updated!

justinyoo:
> @christothes cc @ShivangiReja I am assuming this is not yet ok to merge is that correct? I thought the current (1.5) release does not support this? cc @jsquire

@achandmsft I don't think the current 2.5.0 has yet supported this sample: https://platform.openai.com/docs/guides/images-vision?api-mode=responses#generate-or-edit-images This PR doesn't include it yet. Once a newer version is released, I'll create another PR for it.

christothes (approved):

Issue #259, opened 2024-10-14
Additional properties from OpenAI-compatible response are needed but not exposed


stephentoub:
### Service

OpenAI

### Describe the bug

https://github.com/openai/openai-dotnet/blob/ef201cd6016bc1111be4c0098fd643554d6e9a0b/src/OpenAI.csproj#L54-L57

The OpenAI library is tracking all received properties and storing them into a SerializedAdditionalRawData dictionary. It doesn't publicly expose this data, except via InternalsVisibleTo to Azure.AI.OpenAI.

Why is Azure.AI.OpenAI special here, such that this data is only available to that specific library?

### Steps to reproduce

Try to access additional properties sent down as part of an OpenAI-compatible response. You can't.

### OS

Any

### .NET version

Any

### Library version

2.1.0-beta

KrzysztofCwalina:
We will be exposing this publicly. We just did not have time to design it properly in v1.

JadynWong:
Expect `extra_body`, `extra_headers` to be supported like the Python version.

fartwhif:
can't wait

I understand everything is beta and preview and "happening".  But I'm just saying that the "you've been rate limited" information is extra important for those using realtime with **only 100 RPD**  Not your fault but testing is becoming a drag and knowing instead of guessing the actual block duration ( Please try again in blah blah ) would be _outstanding_.

```
ConversationResponseFinishedUpdate.StatusDetails.SerializedAdditionalRawData

{{"type":"requests","code":"rate_limit_exceeded","message":"Rate limit reached for 
gpt-4o-realtime-preview in organization org-WMbMovmWiKQ0jxiZKkHP5hSx on requests 
per day (RPD): Limit 100, Used 100, Requested 1. Please try again in 14m24s. 
Visit https://platform.openai.com/account/rate-limits to learn more."}}


```

jichulu:
When reading the response from DeepSeek-R1, I need to get the `reasoning_content` property from SerializedAdditionalRawData.
Currently, I'm reading the value of SerializedAdditionalRawData via reflection, which may have a performance issue.
SerializedAdditionalRawData needs to be exposed as public.
```
var chatUpdate = client.CompleteChatStreamingAsync(Messages, options, cancellationToken);
await foreach (var item in chatUpdate)
{
    try
    {
		var additionalRawData = item.GetAdditionalRawData();
		if (additionalRawData?.TryGetValue("reasoning_content", out var val) == true)
		{
[...]

PR #702, opened 2025-09-22
Expose ImageGenerationTool for responses

ShivangiReja:
This PR exposes ImageGenerationTool publicly for Responses.


src/Generated/Models/Responses/ImageGenerationCallResponseItem.Serialization.cs
```diff
@@ -88,7 +88,7 @@ internal static InternalImageGenToolCallItemResource DeserializeInternalImageGen
                 }
                 if (prop.NameEquals("status"u8))
```
ericstj, line 90: There are additional properties on this item that are important to expose.  At a bare minimum `output_format`, but maybe others.  What drives the decision around what to expose vs what to leave  in the `additionalBinaryDataProperties` (which isn't public)?

ShivangiReja: These properties come directly from the OpenAI spec we use for codegen. For .NET, the definitions are in [ImageGenToolCallItemResource](https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/specification/base/typespec/responses/custom/items.tsp#L162) and [ImageGenToolCallItemBase](https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/specification/base/typespec/responses/models.tsp#L1940). The decision on what to expose vs. what ends up in `additionalBinaryDataProperties` follows the spec, only explicitly defined fields are exposed.

src/Custom/Responses/Streaming/StreamingResponseImageGenerationCallCompletedUpdate.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace OpenAI.Responses;
+
+// CUSTOM: Renamed.
+[CodeGenType("ResponseImageGenCallCompletedEvent")]
+public partial class StreamingResponseImageGenerationCallCompletedUpdate
```
christothes, line 5: nit: Can we group these and other simple rename attributes like this into a single GeneratorStubs.cs file under the Responses folder?

christothes (approved):

src/Generated/Models/Responses/ImageGenerationCallResponseItem.Serialization.cs
stephentoub: > follows the spec

Which then begs the question, what drives inclusion in that typespec? :-)

src/Custom/Responses/Streaming/StreamingResponseImageGenerationCallCompletedUpdate.cs
ShivangiReja: I’m following the existing format for now. I do plan to create a separate PR to reorganize the response files more cleanly. In that PR, I’ll add a `GeneratorStubs.cs` and properly organize everything like this:

```
Tools/
│
├── FunctionTool
│   ├── ResponseTool/
│   │   └── (all types for FunctionTool that inherit from ResponseTool + supporting classes)
│   └── ResponseItem/
│       └── (all types for FunctionTool that inherit from ResponseItem + supporting classes)
│
├── ImageGenTool
│   ├── ResponseTool/
│   │   └── (same pattern as FunctionTool)
│   └── ResponseItem/
│       └── (same pattern as FunctionTool)
```

That way we keep the current changes scoped, and the reorganization can be handled cleanly in its own PR.

src/Generated/Models/Responses/ImageGenerationCallResponseItem.Serialization.cs
ShivangiReja: We consume the [OpenAI specification](https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/specification/base/external-specs/latest.yaml) and convert it into TypeSpec. Whatever fields are explicitly defined there (in this case under [ImageGenToolCall](https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/specification/base/external-specs/latest.yaml#L28119-L28155)) end up in the TypeSpec models, and those are what get exposed in .NET.

ShivangiReja: @ericstj The API docs show the same response properties as the spec. See the **Image generation call object** here: [https://platform.openai.com/docs/api-reference/responses/object](https://platform.openai.com/docs/api-reference/responses/object).

image

`OutputFormat` is present on the **request** (via `ImageGenerationTool`), but it’s not part of the **response** payload , hence why it isn’t exposed on the response model.

stephentoub: > We consume the [OpenAI specification](https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/specification/base/external-specs/latest.yaml) and convert it into TypeSpec. Whatever fields are explicitly defined there (in this case under [ImageGenToolCall](https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/specification/base/external-specs/latest.yaml#L28119-L28155)) end up in the TypeSpec models, and those are what get exposed in .NET.

Got it. Thanks.

ericstj: When experimenting with this I noticed I was getting back quite a few other properties like `revised_prompt`, `output_format`, and a few others.  It is helpful to get these along side the image data, so that the data can be interpreted correctly.

Here's sample output, with the data `<excluded>` for brevity:
```
ImageGenerationCall(
  id='ig_68d1dc30632c81a1a2db372f0c9f37b20f65e6e151ee115a', 
  result='<excluded>', 
  status='completed', 
  type='image_generation_call', 
  background='opaque', 
  output_format='png', 
  quality='high', 
  revised_prompt='High-resolution, heartwarming illustration of a gray tabby cat hugging an otter that is wearing an orange scarf. The gray tabby has soft silver fur with darker tabby stripes and bright green eyes; the otter has warm brown fur, small whiskers, and friendly eyes. The otter’s knitted orange scarf is wrapped once around its neck with a little fringe visible. They are seated and embracing, the cat’s paws around the otter, both smiling gently. Soft, cozy lighting; pastel bokeh background; clean composition; no text; cute storybook watercolor style; 3:4 portrait orientation.', 
  size='1024x1536'
)
```
Here's a python sample -- https://gist.github.com/ericstj/e6705d79ce423c009c9a88805c9b311a

I take it you're saying that because these properties aren't included in spec for the response we shouldn't expect them, and instead try to read any configuration we might need from the tool's inputs?

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
         public override readonly string ToString();
     }
     [Experimental("OPENAI001")]
+    public class ImageGenerationCallResponseItem : ResponseItem, IJsonModel<ImageGenerationCallResponseItem>, IPersistableModel<ImageGenerationCallResponseItem> {
+        public ImageGenerationCallResponseItem(ImageGenerationCallStatus status, string result);
```
joseharriaga: Contrary to the other properties, the `Status` is not required (except when this class is used as an output) and is read-only. To achieve this, could you add the following customization in the TypeSpec client.tsp?

```typespec
@@visibility(ImageGenToolCallItemResource.status, Lifecycle.Read);
```

You will notice other tools have the same customization:
🔗 https://github.com/openai/openai-dotnet/blob/main/specification/client/responses.client.tsp#L10

Making the `status` property read-only like this should have the effect of removing the `status` parameter from this constructor.

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
...
+        public string Result { get; set; }
+        public ImageGenerationCallStatus Status { get; set; }
```
joseharriaga: Because the `Status` property is only required when used as an output, it should be nullable when used as an input. To reflect this, could you add the following customization in the C# code:

```csharp
// CUSTOM: Made nullable since this is a read-only property.
[CodeGenMember("Status")]
public ImageGenerationCallStatus? Status { get; }
```

You will notice that we do the same for the other tools:
🔗 https://github.com/openai/openai-dotnet/blob/529cfa054227358ff8f8df2a8628f3edc4406d29/src/Custom/Responses/Items/FunctionTool/FunctionCallResponseItem.cs#L5

If you're curious about why the `Status` property is not generated correctly based on this comment and my previous, this is an issue in the TypeSpec that comes from the original swagger spec. I need to confirm if this issue is still present in the new version that Travis and his team are working on, but for now we can unblock ourselves with these simple customizations.

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
         public override readonly string ToString();
     }
     [Experimental("OPENAI001")]
+    public class ImageGenerationCallResponseItem : ResponseItem, IJsonModel<ImageGenerationCallResponseItem>, IPersistableModel<ImageGenerationCallResponseItem> {
+        public ImageGenerationCallResponseItem(ImageGenerationCallStatus status, string result);
+        public string Result { get; set; }
```
joseharriaga: This result represents the image data as a base64-encoded string. Could you redefine this property in the TypeSpec as follows:

```typespec
@encode("base64", string)
result: bytes | null;
```

This change will make this property be generated as `BinaryData`, which is what we want because it represents raw bytes. To that end, I also wonder if we should rename this property to `GeneratedImageBytes` for clarity. 🤔 (Normally I would like the original name "result" to be somewhere in there, but I can't come up with a good name that has "result" in it).

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
...
+        Generating = 2,
+        Failed = 3
+    }
+    [Experimental("OPENAI001")]
+    public class ImageGenerationTool : ResponseTool, IJsonModel<ImageGenerationTool>, IPersistableModel<ImageGenerationTool> {
+        public ImageGenerationTool();
+        public ImageGenerationToolBackground? Background { get; set; }
+        public ImageGenerationToolInputImageMask InputImageMask { get; set; }
+        public string Model { get; set; }
+        public ImageGenerationToolModeration? Moderation { get; set; }
```
joseharriaga: nit: For consistency with `ImageGenerationOptions`, rename this type to `ImageGenerationToolModerationLevel` and this property to `ModerationLevel`.

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
...
+        public int? OutputCompression { get; set; }
+        public ImageGenerationToolOutputFormat? OutputFormat { get; set; }
```
joseharriaga: nit: For consistency with `ImageGenerationOptions`, rename this class to `ImageGenerationToolOutputFileFormat` and this property to `OutputFileFormat`.

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
...
+        public int? PartialImages { get; set; }
```
joseharriaga: nit: Rename this property to `PartialImageCount`

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
...
+        Failed = 3
+    }
+    [Experimental("OPENAI001")]
+    public class ImageGenerationTool : ResponseTool, IJsonModel<ImageGenerationTool>, IPersistableModel<ImageGenerationTool> {
+        public ImageGenerationTool();
+        public ImageGenerationToolBackground? Background { get; set; }
+        public ImageGenerationToolInputImageMask InputImageMask { get; set; }
+        public string Model { get; set; }
+        public ImageGenerationToolModeration? Moderation { get; set; }
+        public int? OutputCompression { get; set; }
```
joseharriaga: nit: For consistency with `ImageGenerationOptions`, rename this property to `OutputCompressionFactor`.

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
...
+    }
+    [Experimental("OPENAI001")]
+    public enum ImageGenerationCallStatus {
+        InProgress = 0,
+        Completed = 1,
+        Generating = 2,
+        Failed = 3
+    }
+    [Experimental("OPENAI001")]
+    public class ImageGenerationTool : ResponseTool, IJsonModel<ImageGenerationTool>, IPersistableModel<ImageGenerationTool> {
```
joseharriaga, line 4999: nit: Consider manually adding the missing property "input_fidelity" in `ImageGenerationTool` to the TypeSpec by copying it from Travis's TypeSpec.

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override readonly int GetHashCode();
+        public static bool operator ==(ImageGenerationToolBackground left, ImageGenerationToolBackground right);
+        public static implicit operator ImageGenerationToolBackground(string value);
+        public static implicit operator ImageGenerationToolBackground?(string value);
+        public static bool operator !=(ImageGenerationToolBackground left, ImageGenerationToolBackground right);
+        public override readonly string ToString();
+    }
+    [Experimental("OPENAI001")]
+    public class ImageGenerationToolInputImageMask : IJsonModel<ImageGenerationToolInputImageMask>, IPersistableModel<ImageGenerationToolInputImageMask> {
```
joseharriaga, line 5050: If I understand correctly, this class represents a mask that is provided either via a file ID or an image URL. In other words, it is one or the other. Could you confirm this is correct? If so, I recommend removing the setters from the properties, and instead adding three constructors:
1. An `internal` parameterless constructor.
2. A `public` constructor that takes a `string fileId` parameter.
3. A `public` constructor that takes a `Uri imageUri` parameter.

EDIT: Reading the documentation, it appears that the "image URL" is the "base64-encoded image"? I interpret this as saying that this is actually a "data URL" that contains the base64-encoded data of the image instead of pointing to an Internet location. If this is actually the case, then this property should be neither a `string` nor a `Uri`, but `BinaryData`. Here's an example of how we deal with data URLs in other places:
🔗 https://github.com/openai/openai-dotnet/blob/346256d889a7df83301d66dd7f5c6ad3d1d60e54/src/Custom/Responses/ResponseContentPart.cs#L49

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
...
+        public string FileId { get; set; }
+        public string ImageUrl { get; set; }
```
joseharriaga: nit: If this is a "data URL" this should be `BinaryData` and the property should be called `ImageBytes`. Or if this is a normal URL pointing to an Internet location, then it should be a `Uri` and be called `ImageUri`.

api/OpenAI.net8.0.cs
```diff
@@ -4897,6 +4897,135 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
...
+        public static bool operator ==(ImageGenerationToolQuality left, ImageGenerationToolQuality right);
+        public static implicit operator ImageGenerationToolQuality(string value);
+        public static implicit operator ImageGenerationToolQuality?(string value);
+        public static bool operator !=(ImageGenerationToolQuality left, ImageGenerationToolQuality right);
+        public override readonly string ToString();
+    }
+    [Experimental("OPENAI001")]
+    public readonly partial struct ImageGenerationToolSize : IEquatable<ImageGenerationToolSize> {
+        public ImageGenerationToolSize(string value);
+        public static ImageGenerationToolSize _1024x1024 { get; }
```
joseharriaga: For consistency with `GeneratedImageSize`, these should be called in the format "W{width}xH{height}", e.g. "W1024xH1024".

It should also hide the constructor that takes a string in favor of one that takes `int width` and `int height`.

Check out `GeneratedImageSize` for reference:
🔗 https://github.com/openai/openai-dotnet/blob/346256d889a7df83301d66dd7f5c6ad3d1d60e54/src/Custom/Images/GeneratedImageSize.cs

api/OpenAI.net8.0.cs
```diff
@@ -5642,6 +5772,44 @@ public class StreamingResponseFunctionCallArgumentsDoneUpdate : StreamingRespons
...
+        protected override StreamingResponseUpdate JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected override StreamingResponseUpdate PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class StreamingResponseImageGenerationCallPartialImageUpdate : StreamingResponseUpdate, IJsonModel<StreamingResponseImageGenerationCallPartialImageUpdate>, IPersistableModel<StreamingResponseImageGenerationCallPartialImageUpdate> {
+        public string ItemId { get; }
+        public int OutputIndex { get; }
+        public string PartialImageB64 { get; }
```
joseharriaga: This one also needs to be defined in the TypeSpec as `bytes` instead of `string`, and it should have the following decorator: `@encode("base64", string)`. This would generate as `BinaryData`, which is what we want given that this represents raw bytes. To that end, we should also rename this to `PartialImageBytes`.

api/OpenAI.net8.0.cs
ShivangiReja: `ImageUrl` accepts both a data URL and a web URL.

**Examples:**

```csharp
// Data URL (base64-embedded bytes)
BinaryData imageData = BinaryData.FromStream(File.OpenRead("Assets/images_dog_and_cat.png"));
string base64 = Convert.ToBase64String(imageData.ToArray());
string dataUrl = $"data:image/png;base64,{base64}";
inputImageMask.ImageUrl = dataUrl;

// Internet URL (https)
inputImageMask.ImageUrl = "https://upload.wikimedia.org/wikipedia/commons/4/47/PNG_transparency_demonstration_1.png";
```

Based on the above observations, we can do something like this:
```csharp
public partial class ImageGenerationToolInputImageMask
{
    internal ImageGenerationToolInputImageMask() { }

    public ImageGenerationToolInputImageMask(string fileId)
    {
        Argument.AssertNotNullOrEmpty(fileId, nameof(fileId));
        
        FileId = fileId;
    }

    public ImageGenerationToolInputImageMask(BinaryData imageBytes, string imageBytesMediaType)
    {
        Argument.AssertNotNull(imageBytes, nameof(imageBytes));
        Argument.AssertNotNullOrEmpty(imageBytesMediaType, nameof(imageBytesMediaType));

        string base64EncodedData = Convert.ToBase64String(imageBytes.ToArray());
        ImageUrl = $"data:{imageBytesMediaType};base64,{base64EncodedData}";
    }

    public ImageGenerationToolInputImageMask(Uri imageUri)
    {
        Argument.AssertNotNull(imageBytes, nameof(imageBytes));

        ImageUrl = imageUri?.AbsoluteUri;
    }

    public string ImageUrl { get; }

    public string FileId { get; }
}
```

src/Generated/Models/Responses/ImageGenerationCallResponseItem.Serialization.cs
ShivangiReja: In the .NET SDK we only expose properties that are in the Open AI spec. Undocumented properties aren’t considered stable and can change, so we don’t expose them as first-class properties.

That said, you can still read any extra properties from the raw payload. They should be in our internal `SerializedAdditionalRawData` dictionary and we’re tracking making that public here: https://github.com/openai/openai-dotnet/issues/259.

api/OpenAI.net8.0.cs
ShivangiReja: Yes, the mask should be provided by exactly one source: either FileId or ImageUrl. I tested sending both and the service currently accepts it without error, but the docs don’t define which one wins. That’s effectively undefined behavior, but I think the SDK shouldn’t allow both.

api/OpenAI.net8.0.cs
```diff
@@ -4980,9 +4980,9 @@ public class FunctionTool : ResponseTool, IJsonModel<FunctionTool>, IPersistable
     }
     [Experimental("OPENAI001")]
     public class ImageGenerationCallResponseItem : ResponseItem, IJsonModel<ImageGenerationCallResponseItem>, IPersistableModel<ImageGenerationCallResponseItem> {
-        public ImageGenerationCallResponseItem(ImageGenerationCallStatus status, string result);
-        public string Result { get; set; }
-        public ImageGenerationCallStatus Status { get; set; }
+        public ImageGenerationCallResponseItem(BinaryData generatedImageBytes);
+        public BinaryData GeneratedImageBytes { get; set; }
```
christothes, line 4984: Does this need a setter?

ShivangiReja: Yes, this is an input output model, similar to [CodeInterpreterCallResponseItem](https://github.com/openai/openai-dotnet/blob/main/src/Generated/Models/Responses/CodeInterpreterCallResponseItem.cs).

src/Generated/Models/Responses/ImageGenerationCallResponseItem.Serialization.cs
ericstj: I'm not sure if you have a way to relay the ask to get these documented, but `output_format` would be nice to have.

ShivangiReja: As mentioned, we can’t expose properties that aren’t in the OpenAI API spec. The .NET SDK is generated from that spec, so we can’t add anything that isn’t defined there.

ShivangiReja: As mentioned, we can’t expose properties that aren’t in the OpenAI spec. The .NET SDK is generated from that spec, so we can’t add anything that isn’t defined there.

stephentoub: I believe @ericstj is asking if there's a channel to ask OpenAI for those properties to be included in the spec / be documented.

ShivangiReja: @ericstj OpenAI spec is maintained upstream in a non-public repo. Could you open an issue [here](https://github.com/openai/openai-dotnet/issues) with the properties you’d like added to the spec? We’ll relay it upstream and track it here.

ericstj: I opened https://github.com/openai/openai-dotnet/issues/722

ericstj:
@ShivangiReja  is there a preview package available with this?

ShivangiReja:
@ericstj - Not yet. It will be included in the next release.

ericstj:
I see, it would help us be ready for that release in MEAI if a preview package was available.  For now I've built it myself to test locally.

PR #717, opened 2025-09-24
Custom Serializer for Message Creation - Resolves Issue 552

BenjaminDavidPinter:
# Fixes Issue 552 

I'm actually surprised this hasn't come up more often, the FileSearch/CodeGen combo allows for really simple data analysis.

Anyway; I discovered the root cause,  and here what I changed.

## Root cause 
The file_search tool definition in the spec ([models.tsp](https://github.com/openai/openai-dotnet/blob/08a1193f616124e9c93ff991ab8dcb376ef52d28/specification/base/typespec/assistants/models.tsp#L182)) is shared between Assistant creation and Message tool serialization. 

The generated serializer ([FileSearchToolDefinition.Serialization.cs](https://github.com/openai/openai-dotnet/blob/08a1193f616124e9c93ff991ab8dcb376ef52d28/src/Generated/Models/Assistants/FileSearchToolDefinition.Serialization.cs#L30)) force emits a "file_search" object when serializing the tool, which ends up injected into the message attachment payload (an unsupported property for that request).

## Solution
 I did not edit the generated FileSearchToolDefinition.Serialization.cs. Instead I intercepted message-level tool serialization in [MessageCreationAttachment.cs](https://github.com/openai/openai-dotnet/blob/08a1193f616124e9c93ff991ab8dcb376ef52d28/src/Custom/Assistants/MessageCreationAttachment.cs#L29).

I buffer the generated JSON for each tool, parse it, remove any "file_search" property, then write the cleaned JSON into the message attachment payload. This prevents the unwanted property from being sent.

**Why this approach**: generated code can’t be changed directly, so post-processing the generated output in the custom serializer is the safest fix that keeps generated behavior intact while preventing invalid request payloads.

## Testing

I was able to recreate the issue with the added test.
After implementing the fix, the test passes.

All tests with Category "Assistant" pass.

## Recommendation  
Split the spec definition so assistant-level tool properties (like file_search) are not shared with message-level tool representations. Ise separate types for tools used in assistant creation vs tools embedded in messages, or mark server-only/readOnly fields so the generator will not emit them into request models. If the Responses API already addresses this separation, consider aligning the assistants spec with that approach.

### DIFF
--- a/tests/Assistants/AssistantsTests.cs
+++ b/tests/Assistants/AssistantsTests.cs
@@ -31,7 +31,6 @@ public class AssistantsTests : SyncAsyncTestBase
     private readonly List<string> _vectorStoreIdsToDelete = [];
 
     private static readonly DateTimeOffset s_2024 = new(2024, 1, 1, 0, 0, 0, TimeSpan.Zero);
-    private static readonly string s_testAssistantName = $".NET SDK Test Assistant - Please Delete Me";
     private static readonly string s_cleanupMetadataKey = $"test_metadata_cleanup_eligible";
 
     private static AssistantClient GetTestClient() => GetTestClient<AssistantClient>(TestScenario.Assistants);
--- a/tests/Assistants/AssistantsTests.cs
+++ b/tests/Assistants/AssistantsTests.cs
@@ -1,10 +1,4 @@
-﻿using NUnit.Framework;
-using NUnit.Framework.Internal;
-using OpenAI.Assistants;
-using OpenAI.Files;
-using OpenAI.Tests.Utility;
-using OpenAI.VectorStores;
-using System;
+﻿using System;
 using System.ClientModel;
 using System.ClientModel.Primitives;
 using System.Collections.Generic;
@@ -13,6 +7,12 @@ using System.Linq;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
+using NUnit.Framework;
+using NUnit.Framework.Internal;
+using OpenAI.Assistants;
+using OpenAI.Files;
+using OpenAI.Tests.Utility;
+using OpenAI.VectorStores;
 using static OpenAI.Tests.TestHelpers;
 
 namespace OpenAI.Tests.Assistants;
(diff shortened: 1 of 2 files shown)


jsquire:
@BenjaminDavidPinter:  Thanks very much for your contribution and your interest in improving the OpenAI developer experience.   We appreciate the efforts that you put into investigating and writing this up as well.   It sounds like the long-term solution would be to fix the spec such that the models are distinct.    In the short-term, I think your approach here makes sense.    

I'm going to ask @ShivangiReja and @christothes to add some additional perspective.

jsquire:
@joseharriaga:   Would you please take a look at the investigation write-up here and then let's sync on thoughts to push upstream for spec changes?

jsquire (approved):

BenjaminDavidPinter:
@jsquire thanks! I'll fix up the conflicts when I get home tonight, and we can wrap this one up (for now), until we get those models split out in the spec.

jsquire (approved):

jsquire:
Thanks, @BenjaminDavidPinter!    I'm going to run the tests and then wait for Jose to get back for final approval.

tests/Assistants/AssistantsTests.cs
```diff
@@ -14,6 +7,19 @@
...
+using NUnit.Framework.Internal;
+using NUnit.Framework.Internal;
+using OpenAI.Assistants;
+using OpenAI.Assistants;
+using OpenAI.Files;
+using OpenAI.Files;
+using OpenAI.Tests.Utility;
+using OpenAI.Tests.Utility;
+using OpenAI.VectorStores;
+using OpenAI.VectorStores;
```
joseharriaga, line 15: Could you fix the duplicated namespaces?

tests/Assistants/AssistantsTests.cs
```diff
@@ -847,6 +852,45 @@ This file describes the favorite foods of several people.
         });
     }
 
+    [Test]
+    public async Task FileOnMessageWorks()
+    {
+        // First, we need to upload a simple test file.
+        OpenAIFileClient fileClient = GetTestClient<OpenAIFileClient>(TestScenario.Files);
+        OpenAIFile testFile = fileClient.UploadFile(
```
joseharriaga: Switch to the async version of this method: `UploadFileAsync`

tests/Assistants/AssistantsTests.cs
```diff
@@ -847,6 +852,45 @@ This file describes the favorite foods of several people.
...
+            Tekakwitha Effie: pizza
+            Filip Carola: cake
+            """).ToStream(),
+            "favorite_foods.txt",
+            FileUploadPurpose.Assistants);
+        Validate(testFile);
+
+        AssistantClient client = GetTestClient();
+
+        var thread = client.CreateThread();
```
joseharriaga: 1. Switch to the async version of this method: `CreateThreadAsync`.
2. Then call `Validate(thread);`

tests/Assistants/AssistantsTests.cs
```diff
@@ -847,6 +852,45 @@ This file describes the favorite foods of several people.
...
+        var assistant = client.CreateAssistant("gpt-4o-mini");
```
joseharriaga: 1. Switch to the async version of this method: `CreateAssistantAsync`
2. Then call `Validate(assistant);`

tests/Assistants/AssistantsTests.cs
```diff
@@ -847,6 +852,45 @@ This file describes the favorite foods of several people.
...
+
+        var message = await client.CreateMessageAsync(
+        thread.Value.Id,
```
joseharriaga: nit: Add indentation.

tests/Assistants/AssistantsTests.cs
```diff
@@ -847,6 +852,45 @@ This file describes the favorite foods of several people.
...
+        new[] {
+            MessageContent.FromText("What is this file?"),
+        },
+        new MessageCreationOptions()
+        {
+            Attachments = [
+                new MessageCreationAttachment(testFile.Id, new List<ToolDefinition>() { ToolDefinition.CreateFileSearch() }),
+                new MessageCreationAttachment(testFile.Id, new List<ToolDefinition>() { ToolDefinition.CreateCodeInterpreter() })
+                ]
+        });
```
joseharriaga: Call `Validate(message);`

joseharriaga:
I approved the PR. Just added a few comments. 🙂

PR #719, opened 2025-09-24
build: bump generator to 1.0.0-alpha.20250924.1

jorgerangel-msft:
This PR bumps the generator version to `1.0.0-alpha.20250924.1` and bumps the SCM version to `1.7.0`.


codegen/generator/src/OpenAILibraryVisitor.cs
```diff
@@ -49,10 +49,10 @@ public class OpenAILibraryVisitor : ScmLibraryVisitor
 
     protected override TypeProvider VisitType(TypeProvider type)
     {
-        if (type is ModelProvider { BaseModelProvider: null } && type.Fields.Count > 0)
+        var additionalPropertiesField = type.Fields.FirstOrDefault(f => f.Name == AdditionalPropertiesFieldName);
+        if (type is ModelProvider { BaseModelProvider: null } && additionalPropertiesField != null)
         {
             // Add an internal AdditionalProperties property to all base models
-            var additionalPropertiesField = type.Fields.Single(f => f.Name == AdditionalPropertiesFieldName);
```
jorgerangel-msft, line 55: For dynamic models, the SARD field won't be generated and this will throw.

codegen/generator/src/Visitors/ExperimentalAttributeVisitor.cs
```diff
@@ -618,6 +618,12 @@ public class ExperimentalAttributeVisitor : ScmLibraryVisitor
 
         protected override PropertyProvider? VisitProperty(PropertyProvider property)
         {
+            // Skip properties that are already marked as experimental
+            if (property.Attributes.Any(attr => attr.Type.Equals(typeof(ExperimentalAttribute))))
```
jorgerangel-msft, line 622: For dynamic models, the json patch property already contains an experimental attribute. In general, we should not apply the additional attribute if the property is generated with one by default.

src/Generated/Models/Assistants/CodeInterpreterToolDefinition.Serialization.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace OpenAI.Assistants
 {
-    public partial class CodeInterpreterToolDefinition : IJsonModel<CodeInterpreterToolDefinition>
+    public partial class CodeInterpreterToolDefinition : ToolDefinition, IJsonModel<CodeInterpreterToolDefinition>
```
ShivangiReja, line 13: Could you please run `.\scripts\Export-Api.ps1`?

ShivangiReja (approved):

src/Generated/Models/Assistants/CodeInterpreterToolDefinition.Serialization.cs
jorgerangel-msft: Sure, I ran it and there were no diffs.

ShivangiReja: This change looks like it’s updating a public class.

jorgerangel-msft: There were some other changes included in the generator. Once of them was to include any base types defined in custom code into the generated class, so it looks like this worked correctly and it is just now generating the types with the existing base types. ref: https://github.com/microsoft/typespec/pull/8480

ShivangiReja: I am just wondering If we’re updating a public class, why is there no diff when running Export-Api.ps1?

JoshLove-msft: The custom partial class already inherits from ToolDefinition. This change merely makes the generated partial class include the base type defined in the custom partial class.

ShivangiReja: Ahh, so the public api is already updated. Thanks for the clarification.

PR #734, opened 2025-10-02
Ensure explicit operators from ClientResult for AudioTranscription

### LINKED ISSUE #711: [BUG] `AudioTranscription` and `AudioTranslation` are missing the explicit conversion operator from `ClientResult`
### Describe the bug

All our top-level output types (i.e., the types that are returned directly from client methods) should have this operator, which comes from the generator. Here's an example from `ChatCompletion`:

```csharp
public static explicit operator ChatCompletion(ClientResult result);
```

For some reason, `AudioTranscription` and `AudioTranslation` didn't get one. 🤔

### Steps to reproduce

You can see that it is missing in the API listing:
🔗 https://github.com/openai/openai-dotnet/blob/OpenAI_2.4.0/api/OpenAI.net8.0.cs#L1135
🔗 https://github.com/openai/openai-dotnet/blob/OpenAI_2.4.0/api/OpenAI.net8.0.cs#L1194

### Code snippets

```C#

```

### OS

Windows

### .NET version

.NET 8

### Library version

2.4.0

christothes:
fixes #711

### DIFF
--- a/specification/base/typespec/audio/operations.tsp
+++ b/specification/base/typespec/audio/operations.tsp
@@ -3,7 +3,6 @@ import "@typespec/openapi";
 
 import "../common";
 import "./models.tsp";
-import "../../../client/models/audio.models.tsp";
 
 using TypeSpec.Http;
 using TypeSpec.OpenAPI;
--- a/specification/base/typespec/audio/operations.tsp
+++ b/specification/base/typespec/audio/operations.tsp
@@ -41,12 +41,6 @@ interface Audio {
     | CreateTranscriptionResponseVerboseJson
     | CreateTranscriptionResponseJson
     | SseResponseOf<CreateTranscriptionResponseStreamEvent>
-    | {
-        // TODO: This response is not defined in the OpenAPI spec.
-        @header contentType: "text/plain";
-
-        @body responseBody: string;
-      }
     | ErrorResponse;
 
   @route("translations")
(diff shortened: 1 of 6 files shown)


specification/base/typespec/audio/operations.tsp
```diff
@@ -40,12 +41,7 @@ interface Audio {
     | CreateTranscriptionResponseVerboseJson
     | CreateTranscriptionResponseJson
     | SseResponseOf<CreateTranscriptionResponseStreamEvent>
-    | {
-        // TODO: This response is not defined in the OpenAPI spec.
-        @header contentType: "text/plain";
-
-        @body responseBody: string;
-      }
+    | DotNetCombinedJsonTranscriptionResponse
```
ShivangiReja: It looks like we’re defining this model in the client's audio.models.tsp and using it in the base TypeSpec. Since the client TypeSpec is language-specific, should the base TypeSpec reference client-specific types?

christothes: Ah, good catch. We need to define this customization in the client TypeSpec somehow.

christothes: I removed this and just kept the explicit operator in custom code.

specification/base/typespec/audio/operations.tsp
```diff
@@ -3,6 +3,7 @@ import "@typespec/openapi";
 
 import "../common";
 import "./models.tsp";
+import "../../../client/models/audio.models.tsp";
```
ShivangiReja: nit: can we remove this

ShivangiReja (approved):

specification/base/typespec/audio/operations.tsp
christothes: done

PR #736, opened 2025-10-02
Add JsonPatch support for Chat, Responses, and Embeddings

ShivangiReja:
This PR adds `JsonPatch` support to **Chat**, **Responses**, and **Embeddings** so callers can:

* Pass additional properties on requests 
* Read unmodeled properties from responses via the `Patch` bag.

## Motivation

* OpenAI exposes fast-moving, model-specific knobs that aren’t on the typed surface.
* Developers need a way to round-trip those fields without waiting on spec/SDK updates.

#### Examples:

**Responses**

```csharp
// You can use the Patch property to set additional properties in the request
ResponseCreationOptions options = new();
options.Patch.Set("$.reasoning.effort"u8, "high");
options.Patch.Set("$.text.verbosity"u8, "medium");

OpenAIResponse response = client.CreateResponse("What is the answer to the ultimate question of life, the universe, and everything?", options);

// You can also read additional properties back from the response
var effort = response.Patch.GetString("$.reasoning.effort"u8);
var verbosity = response.Patch.GetString("$.text.verbosity"u8);
Console.WriteLine($"effort={effort}, verbosity={verbosity}");
```

Fixes: https://github.com/openai/openai-dotnet/issues/259


examples/Chat/Example10_AdditionalProperties.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        // You can use the Patch property to set additional properties in the request
+        ChatCompletionOptions options = new();
+        options.Patch.Set("$.reasoning_effort"u8, "minimal");
+
+        ChatCompletion completion = client.CompleteChat([new UserChatMessage("Say 'this is a test.'")], options);
+
+        Console.WriteLine($"[ASSISTANT]: {completion.Content[0].Text}");
+
+        // You can also read additional properties back from the response
```
jsquire: We may want to include some additional detail here so that its clear what's happening.   Something like "the service also returns a field named `service_tier` that is not understood by `ChatCompletion`.   You can access this value by using the path with `Patch`.

It may also be helpful to link over to the AP docs and/or find something that's a bit deeper.

examples/Responses/Example07_InputAdditionalProperties.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    [Test]
+    public void Example07_InputAdditionalProperties()
+    {
+        OpenAIResponseClient client = new(model: "gpt-5", apiKey: Environment.GetEnvironmentVariable("OPENAI_API_KEY"));
+
+        // You can use the Patch property to set additional properties in the request
+        ResponseCreationOptions options = new();
+        options.Patch.Set("$.reasoning.effort"u8, "high");
+        options.Patch.Set("$.text.verbosity"u8, "medium");
```
jsquire, line 24: I like this example.   Maybe we can link to the platform docs so that folks can connect the dots and see how they would know what to send?

specification/base/typespec/responses/operations.tsp
```diff
@@ -56,11 +56,7 @@ interface Responses {
     @path
     @example("resp_677efb5139a88190b512bc3fef8e535d")
     response_id: string,
-  ): {
-    id: string;
-    object: "response";
-    deleted: true;
-  } | ResponseErrorResponse;
+  ): DeleteResponseResponse | ResponseErrorResponse;
```
jsquire, line 59: What's our plan to get this merged up with a new official spec version?

src/Custom/Chat/ChatFunctionChoice.Serialization.cs
```diff
@@ -13,6 +14,14 @@ void IJsonModel<ChatFunctionChoice>.Write(Utf8JsonWriter writer, ModelReaderWrit
 
     internal static void SerializeChatFunctionChoice(ChatFunctionChoice instance, Utf8JsonWriter writer, ModelReaderWriterOptions options)
     {
+#pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
```
jsquire, line 17: I wonder if we should just turn this warning off per type rather than try to be this focused.   Thoughts?

specification/base/typespec/responses/operations.tsp
ShivangiReja: I believe `operations.tsp` is handwritten, so when we update to the latest spec, we’ll need to re-apply those edits during the bump to keep behavior consistent.

src/Generated/Models/Responses/CodeInterpreterCallOutput.Serialization.cs
```diff
@@ -44,8 +44,6 @@ protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWrit
                 writer.WritePropertyName("type"u8);
                 writer.WriteStringValue(Kind.ToString());
             }
-
-            Patch.WriteTo(writer);
```
ShivangiReja: @jorgerangel-msft - This shouldn't be removed. Could you please take a look?

src/Generated/Models/Chat/ChatMessage.Serialization.cs
```diff
@@ -31,13 +31,12 @@ protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWrit
...
             }
-            if (Optional.IsDefined(Content) && !Patch.Contains("$.content"u8))
+            // Plugin customization: add Content.IsInnerCollectionDefined() check
+            if (Optional.IsDefined(Content) && Content.IsInnerCollectionDefined() && !Patch.Contains("$.content"u8))
             {
                 writer.WritePropertyName("content"u8);
                 SerializeContentValue(writer, options);
             }
-
-            Patch.WriteTo(writer);
```
ShivangiReja: @jorgerangel-msft - `Patch.WriteTo` method got removed from bunch of models, could you please take a look?

JoshLove-msft: Can we add a regression test that uses this model?

src/Generated/Models/Responses/CodeInterpreterCallOutput.Serialization.cs
JoshLove-msft: Can we add a regression test that uses this model?

ShivangiReja: This is a discriminator type; derived types have `Patch.WriteTo`. I added tests to confirm `Patch` is written, and all tests pass.

src/Generated/Models/Chat/ChatMessage.Serialization.cs
ShivangiReja: This is a discriminator type; derived types have Patch.WriteTo. I added tests to confirm Patch is written, and all tests pass.

jsquire (approved):

src/Generated/Models/Chat/ChatMessage.Serialization.cs
jorgerangel-msft: The discriminated types don't have the call to Patch.WriteTo since the derived types have it. That was one of the issues found initially that was causing the unit tests to fail since the patch data was getting serialized twice

src/Custom/Chat/ChatFunctionChoice.Serialization.cs
JoshLove-msft: The reason for doing it so is along the lines of principle of least privilege. If it were at the file level, users may not know what is necessitating this suppression.

examples/Chat/Example10_AdditionalProperties.cs
```diff
@@ -0,0 +1,35 @@
...
+    [Test]
+    public void Example10_AdditionalProperties()
+    {
+        ChatClient client = new(model: "gpt-5", apiKey: Environment.GetEnvironmentVariable("OPENAI_API_KEY"));
+
+        // Add extra request fields using Patch.
+        // Patch lets you set fields that aren’t modeled on ChatCompletionOptions in the request payload.
+        // See the API docs https://platform.openai.com/docs/api-reference/chat/create for supported additional fields.
+        ChatCompletionOptions options = new();
+        options.Patch.Set("$.reasoning_effort"u8, "minimal");
```
christothes, line 20: since it is something people have asked for, should we do an example where we override the model on the request?

src/Generated/Internal/ModelSerializationExtensions.cs
```diff
@@ -253,6 +256,91 @@ public static void WriteObjectValue(this Utf8JsonWriter writer, object value, Mo
             writer.WriteObjectValue<object>(value, options);
         }
 
+        public static ReadOnlySpan<byte> SliceToStartOfPropertyName(this ReadOnlySpan<byte> jsonPath)
```
christothes, line 259: It would be nice if we added some comments to these methods that describe the pattern they are evaluating and what they are returning, just to make grokking this easy in the future.

christothes (approved):

src/Generated/Internal/ModelSerializationExtensions.cs
jorgerangel-msft: If I recall, I don't think we generate xml docs for internal methods but we could possibly do an inline comment here explaining what this method does? @JoshLove-msft - any thoughts? I'm assuming we want to keep the behavior of not generating xml comments since it would cause changes in multiple places.

JoshLove-msft: I think OpenAI has turned off all XML docs (for public and internal methods). I agree that we should probably add doc comments for these methods in MTG. Perhaps we can provide a way to turn it on/off per method.

jorgerangel-msft: Sounds good. I opened https://github.com/microsoft/typespec/issues/8678 to track this. @christothes - is it fine if this is done as a follow up?

m-nash (approved):
The new visitor tests are a great start thanks for adding them

src/Generated/Internal/ModelSerializationExtensions.cs
christothes: sounds good to me

examples/Chat/Example10_AdditionalProperties.cs
ShivangiReja: Yes, I added an example for it.

examples/Responses/Example09_ModelOverridePerRequest.cs
```diff
@@ -0,0 +1,32 @@
...
+
+public partial class ResponseExamples
+{
+    [Test]
+    public void Example09_ModelOverridePerRequest()
+    {
+        OpenAIResponseClient client = new(model: "gpt-4o", apiKey: Environment.GetEnvironmentVariable("OPENAI_API_KEY"));
+
+        // Add extra request fields using Patch.
+        // Patch lets you set fields like `model` that aren't exposed on ResponseCreationOptions.
```
christothes, line 20:
```suggestion
        // Patch lets you set fields like `model` that aren't exposed on ResponseCreationOptions.
        // This overrides the model set on the client just for the request where this options instance is used.
```

sheng-jie:
Great job

sheng-jie:
When will this feature be released? I can't wait, it's been a month since the last release.

joseharriaga:
> When will this feature be released? I can't wait, it's been a month since the last release.

This will be included in this month's release. Current ETA is this Friday.

PR #778, opened 2025-10-27
Protocol models for ChatClient

christothes:


api/OpenAI.netstandard2.0.cs
```diff
@@ -1,4 +1,8 @@
 namespace OpenAI {
+    public enum CreateChatCompletionRequestModality {
```
KrzysztofCwalina: is this name from tsp? It sounds like a verb phrase. It would be better to call it ChatCompletionRequestModality

api/OpenAI.netstandard2.0.cs
```diff
@@ -1321,36 +1325,54 @@ public class ChatClient {
         public ChatClient(string model, AuthenticationPolicy authenticationPolicy);
         public ChatClient(string model, string apiKey);
         public Uri Endpoint { get; }
-        public string Model { get; }
+        public virtual string Model { get; }
         public ClientPipeline Pipeline { get; }
         public virtual ClientResult<ChatCompletion> CompleteChat(params ChatMessage[] messages);
+        public virtual ClientResult CompleteChat(CreateChatCompletionOptions options, RequestOptions requestOptions = null);
```
KrzysztofCwalina: why is this overload here? It seems like it's not model overload because it returns ClientResult, i.e. takes in a model, but does not return a model.

api/OpenAI.netstandard2.0.cs
christothes: Yes it is. Do we want to get into the business of renaming protocol model properties?

api/OpenAI.netstandard2.0.cs
christothes: I believe this is the patter we discussed for protocol methods, where they take the protocol model input but returns `ClientResult`

api/OpenAI.net8.0.cs
```diff
@@ -1475,17 +1480,23 @@ public class ChatClient {
         [Experimental("OPENAI001")]
         public Uri Endpoint { get; }
         [Experimental("OPENAI001")]
-        public string Model { get; }
+        public virtual string Model { get; }
         public ClientPipeline Pipeline { get; }
         public virtual ClientResult<ChatCompletion> CompleteChat(params ChatMessage[] messages);
+        [Experimental("OPENAI001")]
+        public virtual ClientResult<ChatCompletionResult> CompleteChat(CreateChatCompletionOptions options, CancellationToken cancellationToken = default);
```
stephentoub: Can I pass in a RequestOptions?

christothes: You could use the protocol method overload to do that like this:

```c#
ChatCompletionResult response = (ChatCompletionResult)await client.CompleteChatAsync(options, new RequestOptions());
```

The `CreateChatCompletionOptions` would be implicitly cast to `BinaryContent`.

api/OpenAI.netstandard2.0.cs
KrzysztofCwalina: we dont want to rename properties, unless only to .NET-fy them. But this is not a property name. It's a type name and there are not really type names in REST APIs. I wonder who named this model like that.

christothes: renamed

api/OpenAI.net8.0.cs
```diff
@@ -1533,34 +1540,50 @@ public class ChatClient {
         [Experimental("OPENAI001")]
         public virtual Task<ClientResult<ChatCompletionDeletionResult>> DeleteChatCompletionAsync(string completionId, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
+        public virtual ClientResult<ChatCompletionResult> GetChatCompletion(GetCompletionOptions options, CancellationToken cancellationToken = default);
```
joseharriaga, line 1543: Should the name of the options class match the method name (i.e. `GetChatCompletionOptions` instead of `GetCompletionOptions`)?

api/OpenAI.net8.0.cs
```diff
@@ -1512,17 +1513,23 @@ public class ChatClient {
         [Experimental("OPENAI001")]
         public Uri Endpoint { get; }
         [Experimental("OPENAI001")]
-        public string Model { get; }
+        public virtual string Model { get; }
```
joseharriaga, line 1516: Why is the `Model` property virtual? 🤔

api/OpenAI.net8.0.cs
```diff
@@ -1512,17 +1513,23 @@ public class ChatClient {
...
         public virtual ClientResult<ChatCompletion> CompleteChat(IEnumerable<ChatMessage> messages, ChatCompletionOptions options = null, CancellationToken cancellationToken = default);
         public virtual Task<ClientResult<ChatCompletion>> CompleteChatAsync(params ChatMessage[] messages);
+        [Experimental("OPENAI001")]
+        public virtual Task<ClientResult<ChatCompletionResult>> CompleteChatAsync(CompleteChatOptions options, CancellationToken cancellationToken = default);
         public virtual Task<ClientResult> CompleteChatAsync(BinaryContent content, RequestOptions options = null);
         public virtual Task<ClientResult<ChatCompletion>> CompleteChatAsync(IEnumerable<ChatMessage> messages, ChatCompletionOptions options = null, CancellationToken cancellationToken = default);
         public virtual CollectionResult<StreamingChatCompletionUpdate> CompleteChatStreaming(params ChatMessage[] messages);
+        public virtual CollectionResult<StreamingChatCompletionUpdate> CompleteChatStreaming(CompleteChatOptions options, CancellationToken cancellationToken = default);
```
joseharriaga, line 1529: This new method is missing the Experimental attribute.

api/OpenAI.net8.0.cs
```diff
@@ -1512,17 +1513,23 @@ public class ChatClient {
...
         public virtual CollectionResult<StreamingChatCompletionUpdate> CompleteChatStreaming(IEnumerable<ChatMessage> messages, ChatCompletionOptions options = null, CancellationToken cancellationToken = default);
         public virtual AsyncCollectionResult<StreamingChatCompletionUpdate> CompleteChatStreamingAsync(params ChatMessage[] messages);
+        public virtual AsyncCollectionResult<StreamingChatCompletionUpdate> CompleteChatStreamingAsync(CompleteChatOptions options, CancellationToken cancellationToken = default);
```
joseharriaga, line 1532: This new method is missing the Experimental attribute.

api/OpenAI.net8.0.cs
```diff
@@ -1677,7 +1726,6 @@ public class ChatCompletionMessageCollectionOptions : IJsonModel<ChatCompletionM
     public class ChatCompletionMessageListDatum : IJsonModel<ChatCompletionMessageListDatum>, IPersistableModel<ChatCompletionMessageListDatum> {
         public IReadOnlyList<ChatMessageAnnotation> Annotations { get; }
         public string Content { get; }
-        public IList<ChatMessageContentPart> ContentParts { get; }
```
joseharriaga, line 1680: Was this removal intentional?

api/OpenAI.net8.0.cs
```diff
@@ -1737,6 +1785,68 @@ public class ChatCompletionOptions : IJsonModel<ChatCompletionOptions>, IPersist
...
+        public string Refusal { get; }
+        public ChatMessageRole Role { get; }
+        public IReadOnlyList<ChatToolCall> ToolCalls { get; }
+        protected virtual ChatCompletionResponseMessage JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected virtual ChatCompletionResponseMessage PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class ChatCompletionResponseMessageFunctionCall : IJsonModel<ChatCompletionResponseMessageFunctionCall>, IPersistableModel<ChatCompletionResponseMessageFunctionCall> {
```
joseharriaga, line 1803: Can this type be removed in favor of `ChatFunctionCall` ?

api/OpenAI.net8.0.cs
```diff
@@ -1533,34 +1540,50 @@ public class ChatClient {
...
         [Experimental("OPENAI001")]
+        public virtual Task<ClientResult<ChatCompletionResult>> GetChatCompletionAsync(GetCompletionOptions options, CancellationToken cancellationToken = default);
+        [Experimental("OPENAI001")]
         public virtual Task<ClientResult> GetChatCompletionAsync(string completionId, RequestOptions options);
         [Experimental("OPENAI001")]
         public virtual Task<ClientResult<ChatCompletion>> GetChatCompletionAsync(string completionId, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
+        public virtual CollectionResult GetChatCompletionMessages(GetCompletionMessageOptions options, RequestOptions requestOptions = null);
+        [Experimental("OPENAI001")]
+        public virtual ClientResult<CompletionMessageCollection> GetChatCompletionMessages(GetCompletionMessageOptions options, CancellationToken cancellationToken = default);
```
joseharriaga, line 1557: `GetCompletionMessageOptions` --> `GetChatCompletionMessagesOptions`

api/OpenAI.net8.0.cs
```diff
@@ -1533,34 +1540,50 @@ public class ChatClient {
...
         [Experimental("OPENAI001")]
-        public virtual AsyncCollectionResult GetChatCompletionMessagesAsync(string completionId, string after, int? limit, string order, RequestOptions options);
+        public virtual AsyncCollectionResult<ChatCompletionMessageListDatum> GetChatCompletionMessagesAsync(string completionId, ChatCompletionMessageCollectionOptions options = null, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
         public virtual CollectionResult<ChatCompletion> GetChatCompletions(ChatCompletionCollectionOptions options = null, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
-        public virtual CollectionResult GetChatCompletions(string after, int? limit, string order, IDictionary<string, string> metadata, string model, RequestOptions options);
+        public virtual CollectionResult GetChatCompletions(ChatCompletionsOptions options, RequestOptions requestOptions);
+        [Experimental("OPENAI001")]
+        public virtual ClientResult<CompletionCollection> GetChatCompletions(ChatCompletionsOptions options, CancellationToken cancellationToken = default);
```
joseharriaga, line 1571: `ChatCompletionsOptions` --> `GetChatCompletionsOptions`

api/OpenAI.net8.0.cs
```diff
@@ -2081,6 +2193,11 @@ public class ChatOutputTokenUsageDetails : IJsonModel<ChatOutputTokenUsageDetail
         public static bool operator !=(ChatReasoningEffortLevel left, ChatReasoningEffortLevel right);
         public override readonly string ToString();
     }
+    [Experimental("OPENAI001")]
+    public enum ChatRequestModality {
+        Text = 0,
+        Audio = 1
+    }
```
joseharriaga, line 2200: This can probably be replaced by the existing `ChatResponseModality`. Currently, `ChatResponseModality` is implemented as flags, but I think that changing it to a normal enum like you have here might actually better. With this in mind, I'm in favor of the following:
1. Rename this `ChatRequestModality` to `ChatResponseModality`.
2. Turn this `ChatRequestModality` into an extensible enum.
3. Delete the existing flags called `ChatResponseModalities` (this is possible because it's experimental).
4. Change the type of the `ResponeModalities` property in `ChatCompletionOptions` from `ChatResponseModalities` to `IList<ChatResponseModality>` (just like in the protocol model).

api/OpenAI.net8.0.cs
```diff
@@ -2243,6 +2360,81 @@ public class ChatWebSearchOptions : IJsonModel<ChatWebSearchOptions>, IPersistab
...
+        public ChatWebSearchOptions WebSearchOptions { get; set; }
+        public static CompleteChatOptions Create(IEnumerable<ChatMessage> messages, ChatClient client, ChatCompletionOptions options = null, bool isStreaming = false);
+        protected override CompleteChatOptions CreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        [Experimental("OPENAI001")]
+        protected virtual CompleteChatOptions JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        public static implicit operator BinaryContent(CompleteChatOptions createCompletionRequest);
+        protected override void WriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class CompletionCollection : IJsonModel<CompletionCollection>, IPersistableModel<CompletionCollection> {
```
joseharriaga, line 2408: Should this be called `ChatCompletionCollection` ?

api/OpenAI.net8.0.cs
```diff
@@ -1533,34 +1540,50 @@ public class ChatClient {
...
+        public virtual ClientResult<CompletionCollection> GetChatCompletions(ChatCompletionsOptions options, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
         public virtual AsyncCollectionResult<ChatCompletion> GetChatCompletionsAsync(ChatCompletionCollectionOptions options = null, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
-        public virtual AsyncCollectionResult GetChatCompletionsAsync(string after, int? limit, string order, IDictionary<string, string> metadata, string model, RequestOptions options);
+        public virtual AsyncCollectionResult GetChatCompletionsAsync(ChatCompletionsOptions options, RequestOptions requestOptions);
+        [Experimental("OPENAI001")]
+        public virtual Task<ClientResult<CompletionCollection>> GetChatCompletionsAsync(ChatCompletionsOptions options, CancellationToken cancellationToken = default);
+        [Experimental("OPENAI001")]
+        public virtual ClientResult<ChatCompletionResult> UpdateChatCompletion(UpdateChatCompletionOptions options, CancellationToken cancellationToken = default);
```
joseharriaga, line 1579: `ChatCompletionsOptions` looks identical to `ChatCompletionCollectionOptions`. Could we remove the former?

api/OpenAI.net8.0.cs
```diff
@@ -1533,34 +1540,50 @@ public class ChatClient {
...
+        [Experimental("OPENAI001")]
+        public virtual ClientResult<CompletionCollection> GetChatCompletions(ChatCompletionsOptions options, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
         public virtual AsyncCollectionResult<ChatCompletion> GetChatCompletionsAsync(ChatCompletionCollectionOptions options = null, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
-        public virtual AsyncCollectionResult GetChatCompletionsAsync(string after, int? limit, string order, IDictionary<string, string> metadata, string model, RequestOptions options);
+        public virtual AsyncCollectionResult GetChatCompletionsAsync(ChatCompletionsOptions options, RequestOptions requestOptions);
+        [Experimental("OPENAI001")]
+        public virtual Task<ClientResult<CompletionCollection>> GetChatCompletionsAsync(ChatCompletionsOptions options, CancellationToken cancellationToken = default);
+        [Experimental("OPENAI001")]
```
joseharriaga, line 1578: I think it's a little unexpected that the output of this new method appears to be lower level than the output of the protocol method (which I have always understood as the lowest level). How do we feel about the protocol method supporting pagination while this new method doesn't?

api/OpenAI.net8.0.cs
```diff
@@ -1533,34 +1540,50 @@ public class ChatClient {
...
         [Experimental("OPENAI001")]
         public virtual ClientResult<ChatCompletion> GetChatCompletion(string completionId, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
+        public virtual Task<ClientResult<ChatCompletionResult>> GetChatCompletionAsync(GetCompletionOptions options, CancellationToken cancellationToken = default);
+        [Experimental("OPENAI001")]
         public virtual Task<ClientResult> GetChatCompletionAsync(string completionId, RequestOptions options);
         [Experimental("OPENAI001")]
         public virtual Task<ClientResult<ChatCompletion>> GetChatCompletionAsync(string completionId, CancellationToken cancellationToken = default);
         [Experimental("OPENAI001")]
+        public virtual CollectionResult GetChatCompletionMessages(GetCompletionMessageOptions options, RequestOptions requestOptions = null);
```
joseharriaga, line 1555: I think that the only difference between `GetCompletionMessageOptions` and `ChatCompletionMessageCollectionOptions` is that the former includes the `CompletionId`. I wonder if we should normalize it somehow so that we don't need two classes and two sets of methods.

PR #781, opened 2025-10-28
Fix custom code AOT compatibility

m-nash:
Changes MRW calls to use the AOT safe version in custom code.

### DIFF
--- a/src/Custom/Responses/Pagination/Internal/AsyncResponseItemCollectionResult.cs
+++ b/src/Custom/Responses/Pagination/Internal/AsyncResponseItemCollectionResult.cs
@@ -53,7 +53,7 @@ internal class AsyncResponseItemCollectionResult : AsyncCollectionResult<Respons
         Argument.AssertNotNull(page, nameof(page));
 
         PipelineResponse response = page.GetRawResponse();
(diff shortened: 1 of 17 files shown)


src/Custom/Responses/Pagination/Internal/AsyncResponseItemCollectionResult.cs
```diff
@@ -53,7 +53,7 @@ protected override IAsyncEnumerable<ResponseItem> GetValuesFromPageAsync(ClientR
         Argument.AssertNotNull(page, nameof(page));
 
         PipelineResponse response = page.GetRawResponse();
-        InternalResponseItemList list = ModelReaderWriter.Read<InternalResponseItemList>(response.Content)!;
+        InternalResponseItemList list = ModelReaderWriter.Read<InternalResponseItemList>(response.Content, ModelReaderWriterOptions.Json, OpenAIContext.Default)!;
```
joseharriaga, line 56: I suspect that the following four internal classes edited in this PR are actually not used anymore, and that they are leftovers from a previous PR where they should have been deleted instead:

* ResponseItemCollectionResult
* AsyncResponseItemCollectionResult
* VectorStoreCollectionResult
* AsyncVectorStoreCollectionResult

joseharriaga (approved):

Issue #776, opened 2025-10-25
VideoClient: use standard pagination pattern for GetVideos/GetVideosAsync


ShivangiReja:
Currently `GetVideos`/`GetVideosAsync` return a single page wrapped in `ClientResult`. Switch these methods to the standard pagination pattern (same as [ContainerClient](https://github.com/openai/openai-dotnet/blob/84fb83b2ba5825ef83b181c9541cf0c526045c4f/api/OpenAI.net8.0.cs#L2429)) so callers can enumerate pages. 

Refer [this](https://github.com/openai/openai-dotnet/pull/624) PR and [this gist](https://gist.github.com/christothes/4c58e23f063bf3ecc6f3b5d386f4a0dd) to implement the pagination pattern.

PR #789, opened 2025-10-29
Add pagination support to GetVideos and GetVideosAsync in the VideoClient

ShivangiReja:
Fixes: https://github.com/openai/openai-dotnet/issues/776


specification/client/models/videos.models.tsp
```diff
@@ -0,0 +1,30 @@
...
+
+union VideoItemCollectionOrder {
+   string,
+   Ascending: "asc",
+   Descending: "desc",
+}
+
+@access(Access.public)
+@usage(Usage.input)
+model VideoItemCollectionOptions {
```
joseharriaga: nit: Since this is a collection of "videos", these models can simply be prefixed with "Video" instead of "VideoItem":

* VideoCollectionOrderQueryParameter
* VideoCollectionOrder
* VideoCollectionOptions

And then these names should propagate to the C# layer.

For comparison, in the case of the Conversations API, the operation was specifically listing "conversation items" instead of "conversations", hence the difference in naming.

PR #792, opened 2025-10-30
Add `ContainerFileCitationMessageAnnotation`

joseharriaga:
* Added `ContainerFileCitationMessageAnnotation` as a class derived from `ResponseMessageAnnotation`
* Renamed the following methods in the `ContainerClient`:
  * `GetContainerFileContent` --> `DownloadContainerFile`
  * `GetContainerFileContentAsync` --> `DownloadContainerFileAsync`

### DIFF
--- a/examples/Responses/Example10_CodeInterpreterAsync.cs
+++ b/examples/Responses/Example10_CodeInterpreterAsync.cs
@@ -29,7 +29,7 @@ public partial class ResponseExamples
 
         List<ResponseItem> inputItems =
         [
-            ResponseItem.CreateUserMessageItem("Create an Excel spreadsheet that contains the mathematical times tables from 1-12."),
+            ResponseItem.CreateUserMessageItem("Create an Excel spreadsheet that contains the mathematical times tables from 1-12 and make it available for download."),
         ];
 
         OpenAIResponse response = await client.CreateResponseAsync(inputItems, options);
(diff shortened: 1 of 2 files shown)


examples/Responses/Example10_CodeInterpreterAsync.cs
```diff
@@ -0,0 +1,57 @@
...
+            .FirstOrDefault();
+
+        ContainerFileCitationMessageAnnotation containerFileCitation = contentPart.OutputTextAnnotations
+            .OfType<ContainerFileCitationMessageAnnotation>()
+            .FirstOrDefault();
+
+        // Download the file from the container and save it.
+        ContainerClient containerClient = new(apiKey: Environment.GetEnvironmentVariable("OPENAI_API_KEY"));
+        BinaryData fileBytes = await containerClient.DownloadContainerFileAsync(containerFileCitation.ContainerId, containerFileCitation.FileId);
+        using FileStream stream = File.OpenWrite($"{Guid.NewGuid()}.xlsx");
```
christothes: Does the response provide any information about the expected file extension of the file, or would the caller always need to know this somehow?

joseharriaga: The extension is visible in the `Filename` of the `ContainerFileCitationMessageAnnotation`. I have modified the sample to use that instead.

jsquire (approved):

examples/Responses/Example10_CodeInterpreterAsync.cs
christothes: much better!

PR #807, opened 2025-11-04
Fix generated serialization name of "partial_image_b64"

joseharriaga:


src/Generated/Models/Responses/StreamingResponseImageGenerationCallPartialImageUpdate.Serialization.cs
```diff
@@ -55,9 +55,9 @@ protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWri
                 writer.WritePropertyName("partial_image_index"u8);
                 writer.WriteNumberValue(PartialImageIndex);
             }
-            if (!Patch.Contains("$.PartialImageBytes"u8))
+            if (!Patch.Contains("$.partial_image_b64"u8))
```
christothes, line 58: Is there a test that would validate this serialiation fix?

src/Custom/Responses/Streaming/StreamingResponseImageGenerationCallPartialImageUpdate.cs
```diff
@@ -1,7 +1,12 @@
...
+﻿using System;
+
+namespace OpenAI.Responses;
 
 // CUSTOM: Renamed.
 [CodeGenType("ResponseImageGenCallPartialImageEvent")]
 public partial class StreamingResponseImageGenerationCallPartialImageUpdate
 {
+    // CUSTOM: Renamed.
+    [CodeGenMember("PartialImageB64")]
```
JoshLove-msft: We should fix this in the client.tsp by adding a JSON usage as discussed offline.

JoshLove-msft: We should probably add a JSON usage at the namespace level so it is applied everywhere. https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/decorators/#@Azure.ClientGenerator.Core.usage

joseharriaga: For now, I applied it to `ResponseStreamEvent` so that it gets applied to all the models that extend from it. These streaming models are the ones that are not referenced by the operations, and that is the reason why the "usage" needs to be specified manually.

JoshLove-msft (approved):

src/Generated/Models/Responses/StreamingResponseImageGenerationCallPartialImageUpdate.Serialization.cs
joseharriaga: Yes, I have added an assert to an existing test to validate this.

christothes (approved):

Issue #748, opened 2025-10-07
[FEATURE REQ] Support for `response.reasoning_text.delta`


anktsrkr:
### Describe the feature or improvement you are requesting

As @stephentoub mentioned in https://github.com/microsoft/agent-framework/issues/1250 currently OpenAI sdk does not support `response.reasoning_text.delta` which is emitted by **_gpt-oss_**, Ref : https://cookbook.openai.com/articles/gpt-oss/handle-raw-cot

SDK only supports `response.reasoning_summary_text.delta` for which I did not find a definitive statement that GPT-OSS emits such events. 

In such case could we add this event as well?

jsquire:
Hi @anktsrkr.  Thanks for reaching out and for your request.   The `response.reasoning_text.data` even is a newer event and is not yet represented in the set of response events in the REST API spec used by the client.  Once the spec is updated to include this area, it will be added to the client.

In the meantime, you can use following approach to get access to the raw event stream and use [SseParser](https://learn.microsoft.com/en-us/dotnet/api/system.net.serversentevents.sseparser?view=net-9.0-pp) to capture the event:

```csharp
var client = new OpenAIResponseClient("gpt-4.1", "<< YOUR API KEY >>");

// Generally, samples show iterating over this collection to process each update.
AsyncCollectionResult<StreamingResponseUpdate> responseUpdates = client.CreateResponseStreamingAsync("Say 'this is a test.'");

// Instead, unwrap into each raw page.
await foreach (var page in responseUpdates.GetRawPagesAsync())
{
    // The content stream is available on the raw response.  Note that this is unbuffered, so 
    // you cannot access via the "Content" property - you must use the stream.
    var sseContentStream =  page.GetRawResponse().ContentStream;
}
```

Writing out the raw content stream for that example results in something like:
```text
event: response.created
[...]

Issue #643, opened 2025-09-02
[FEATURE REQ] Be able to retrieve Streaming Reasoning Content


rogerbarreto:
### Describe the feature or improvement you are requesting

Currently we don't have a public and reliable way to extract the streaming thinking/reasoning  of reasoning models without resorting to reflection (not AOT friendly) or serialization (extra resource) of a specific chunk.

### Additional context

Ideally we should be able to check for the `StreamingResponseUpdate` type and get the `Delta` property back if it is a suggested public `ResponseReasoningSummaryTextDeltaUpdate` type unfortunately we don't have a way to do it today.

rliberoff:
We are facing the same issue. Actually, our users are requesting to see the chain of thought while the model "thinks or reasons" the proper actions.

rogerbarreto:
FYI, while the feature isn't public or available yet, it is possible thru breaking glass to get this information, not ideal though.

### Workaround (Unblocking)

I have a proposed change to the `Microsoft.Extensions.AI.OpenAI` with the breaking glass approach below, to be updated as soon this issue is resolved.

```csharp
// Breaking-Glass capture of streaming reasoning contents
if (streamingUpdate.GetType().Name == "InternalResponseReasoningSummaryTextDeltaEvent")
{
    var responseDeltaElement = JsonSerializer.Deserialize<JsonElement>(JsonSerializer.Serialize(streamingUpdate));

    if (responseDeltaElement.TryGetProperty("delta", out var deltaProperty))
    {
        // Your code with the thinking content in 
        var deltaContent = deltaProperty.GetString()
    }
}
```

- https://github.com/dotnet/extensions/pull/6761

jsquire:
Hi @rogerbarreto.   Thanks for reaching out and we regret that you're experiencing difficulties.   Would you please provide additional context around your end-to-end scenario and where you're running into difficulties?   Seeing a code snippet of what you're attempting to do would be helpful for us to better understand and potentially offer guidance.

gaspardpetit:
If I may comment on this, here's a full sample that provides reasoning summaries and encrypted content. It does not compile without altering the client code, but it shows you the path we would need for a clean usage.

```cs
#pragma warning disable OPENAI001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
#pragma warning disable OPENAICUA001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.


[...]

PR #808, opened 2025-11-05
Expose reasoning streaming events in Responses

ShivangiReja:
This PR exposes the following response streaming events in the SDK:

* StreamingResponseReasoningSummaryPartAddedUpdate
(`response.reasoning_summary_part.added`)

* StreamingResponseReasoningSummaryPartDoneUpdate
(`response.reasoning_summary_part.done`)

* StreamingResponseReasoningSummaryTextDeltaUpdate
(`response.reasoning_summary_text.delta`)

* StreamingResponseReasoningSummaryTextDoneUpdate
(`response.reasoning_summary_text.done`)

* StreamingResponseReasoningTextDeltaUpdate
(`response.reasoning_text.delta`)

* StreamingResponseReasoningTextDoneUpdate
(`response.reasoning_text.done`)

Ref issues:
* https://github.com/openai/openai-dotnet/issues/643
* https://github.com/openai/openai-dotnet/issues/748

### DIFF
--- a/specification/base/typespec/responses/models.tsp
+++ b/specification/base/typespec/responses/models.tsp
@@ -1762,10 +1762,10 @@ model ResponseReasoningTextDeltaEvent extends ResponseStreamEvent {
   item_id: string;
 
   /** The index of the output item this reasoning text delta is associated with. */
-  output_index: integer;
+  output_index: int32;
 
   /** The index of the reasoning content part this delta is associated with. */
-  content_index: integer;
+  content_index: int32;
 
   /** The text delta that was added to the reasoning content. */
   delta: string;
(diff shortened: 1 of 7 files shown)


specification/base/typespec/responses/models.tsp
```diff
@@ -1750,6 +1752,44 @@ model ResponseReasoningSummaryTextDoneEvent extends ResponseStreamEvent {
...
+  type: ResponseStreamEventType.response_reasoning_text_delta;
+
+  /** The ID of the item this reasoning text delta is associated with. */
+  item_id: string;
+
+  /** The index of the output item this reasoning text delta is associated with. */
+  output_index: integer;
+
+  /** The index of the reasoning content part this delta is associated with. */
+  content_index: integer;
```
joseharriaga: Could we temporarily replace these two `integer` for `int32` and re-generate, so that the properties are generated as `int` instead of `long` in C#? Same in the `ResponseReasoningTextDoneEvent` model. We have an open conversation about how we want handle these unspecified numeric types going forward, but it is probably simpler to keep everything all the same for now, and then change it all at the same time if needed. 🙂

Issue #811, opened 2025-11-06
[FEATURE REQ] CreateWebSearchTool() support "web_search"


The-Mojoo:
### Describe the feature or improvement you are requesting

When I use the model **gpt-5-chat-latest**, the following error is prompted
<img width="469" height="65" alt="Image" src="https://github.com/user-attachments/assets/046f3cd9-6b11-4673-8aa0-f7ee3b5dcecb" />
I tried to modify the source code by adding the following code
```
private const string WebSearchValue = "web_search";
private const string WebSearchPreviewValue = "web_search_preview";
internal static InternalToolType WebSearch { get; } = new InternalToolType(WebSearchValue);
internal static InternalToolType WebSearchPreview { get; } = new InternalToolType(WebSearchPreviewValue);
```
I found that the number of classes to be added kept increasing, which made it not concise and elegant enough. I hope the official implementation allows, when using **ResponseTool.CreateWebSearchTool()**, the option to choose **ResponseTool.CreateWebSearchPreviewTool()**, or to add a parameter to control whether it is a _preview_.

### Additional context

The reason for using the **gpt-5-chat-latest** model is that there were organizational verification issues when using gpt-5.
<img width="784" height="81" alt="Image" src="https://github.com/user-attachments/assets/9dd15b83-dd3c-4fe2-999a-ac170a8caa53" />
However, even after verification, there is still such a prompt. Refer to [Gpt5 400: Your organization must be verified to stream this model](https://www.reddit.com/r/OpenWebUI/comments/1ml9egl/gpt5_400_your_organization_must_be_verified_to/). I used **gpt-5-chat-latest**, so the above error was displayed.I also cannot guarantee that after passing the organization verification, there won't still be an error with **ResponseTool.CreateWebSearchTool()**.

jsquire:
Hi @The-Mojoo.   Thanks for reaching out and we regret that you're experiencing difficulties.  The non-preview version of the web search tool is a recent addition to the OpenAI service and is not yet represented in the REST API spec used for client generation.   Once added to the spec, it will be added to the client library.   

In the meantime, you can override the default type using the new JSON Patch feature introduced in v2.6.0:

```csharp
// The default creation is the preview tool.
var webSearchTool = ResponseTool.CreateWebSearchTool();

// Override the type property of the tool to the stable version.
webSearchTool.Patch.Set("$.type"u8, "web_search");
```

The-Mojoo:
[...]

Issue #721, opened 2025-09-25
Update base TypeSpec to generate clients in their corresponding nested namespaces


ShivangiReja:
Currently, there is only one namespace, `namespace OpenAI;`, in the base TypeSpec. We need to be able to generate clients in their corresponding nested namespaces (e.g., `ChatClient` → `OpenAI.Chat`, `OpenAIResponseClient` → `OpenAI.Responses`, and any new feature → `OpenAI.<Feature>`).

My understanding is that TypeSpec has a feature that could help with this: 🔗 [https://typespec.io/docs/language-basics/namespaces/](https://typespec.io/docs/language-basics/namespaces/)

The solution might be to write our base TypeSpec files (`models.tsp`, `operations.tsp`..etc) in a way that leverages this feature so that each sub-client goes into its own namespace.

By doing this, 
* We don’t have to write custom code for each type, just to update the namespace.
* This will also enable us to apply decorators at the namespace level in `client.tsp`. Currently, we can’t scope a decorator to a specific subclient(like Response or Chat) via namespace, so the only option is annotating each target model individually.

baywet:
Update from internal discussion with @glecaros: this might not be desirable in the source specification as it might impact component names and other things. Also service consumers probably won't want to have multiple imports. Maybe this would be better addressed through a plugin instead.

jsquire:
@baywet:   I'd prefer that we discuss before that decision is made.   There's a non-trivial impact here that I would like to be sure that we're accounting for as well as understanding your scenario better.

//cc: @achandmsft, @trrwilson

baywet:
@jsquire would you mind please describing some of those impacts here for everyone's context?

ShivangiReja:
@baywet A plugin/visitor can customize emitted C# namespaces, but it doesn’t address issues caused by having a single `OpenAI` namespace in the spec. TypeSpec (and our C# codegen) support namespace-level decorators, but with only one namespace we can’t use them without applying the policy globally. That means we can’t scope a decorator to just Chat or just Responses, , any namespace-level decorator becomes global, so we’re forced into noisy, error-prone per-type annotations. Modeling sub-clients as nested namespaces in the TypeSpec (e.g., `OpenAI.Chat`, `OpenAI.Responses`) lets us apply policy once at the right boundary and keeps the spec and generated code aligned.

jsquire:
[...]

PR #815, opened 2025-11-06
Add stable web_search support and separate factory for web_search_preview

ShivangiReja:
This PR adds support for the stable web search tool and separate factory for preview web search tool.

  ```csharp
public static WebSearchPreviewTool CreateWebSearchPreviewTool(
WebSearchToolLocation userLocation = null, WebSearchToolContextSize? searchContextSize = null);

public static WebSearchTool CreateWebSearchTool(
WebSearchToolLocation userLocation = null, WebSearchToolContextSize? searchContextSize = null, WebSearchToolFilters filters = null);
  ```

Ref issue: https://github.com/openai/openai-dotnet/issues/811

### DIFF
--- a/api/OpenAI.net8.0.cs
+++ b/api/OpenAI.net8.0.cs
@@ -6650,6 +6650,9 @@ namespace OpenAI.Responses {
     [Experimental("OPENAI001")]
     public class WebSearchToolFilters : IJsonModel<WebSearchToolFilters>, IPersistableModel<WebSearchToolFilters> {
         public IList<string> AllowedDomains { get; set; }
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Experimental("SCME0001")]
+        public ref JsonPatch Patch { get; }
         protected virtual WebSearchToolFilters JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
         protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
         protected virtual WebSearchToolFilters PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
--- a/api/OpenAI.netstandard2.0.cs
+++ b/api/OpenAI.netstandard2.0.cs
@@ -5851,6 +5851,8 @@ namespace OpenAI.Responses {
     }
     public class WebSearchToolFilters : IJsonModel<WebSearchToolFilters>, IPersistableModel<WebSearchToolFilters> {
         public IList<string> AllowedDomains { get; set; }
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public ref JsonPatch Patch { get; }
         protected virtual WebSearchToolFilters JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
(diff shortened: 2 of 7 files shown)


jsquire (approved):
Nice!    The only thing that we may want to do is ask Chris and KC their opinions on the name for `usePreview`.    It makes perfect sense to me.

christothes:
> Nice! The only thing that we may want to do is ask Chris and KC their opinions on the name for `usePreview`. It makes perfect sense to me.

The only feedback I have is that we typical name bool flags something like `IsPreviewEnabled`

joseharriaga:
I talked to Shivangi offline. I'm realizing now that one issue with this approach is that it doesn't fully work for outputs (or more specifically, for deserialization). That is because the following code would return an "unknown tool" instead of the `WebSearchTool` based on value of the discriminator:
🔗 https://github.com/openai/openai-dotnet/blob/ba45b1f4a01b0f0ca6e0e75d6382d7a1662ffa90/src/Generated/Models/Responses/ResponseTool.Serialization.cs#L79-L80

One possible fix is to customize the deserialization code to make it understand both "web_search" and "web_search_preview". However, I think it would be a little inconvenient having to maintain this code manually, especially given how common it is for new tools to be added. Based on this, I'm tempted to suggest that we should add a new `WebSearchPreviewTool` to support this use case side by side with the existing `WebSearchTool`. We will keep `WebSearchPreviewTool` as "experimental" forever (it is preview after all), and then, one day, we can probably remove it fully.

ShivangiReja:
Yes, agreed. Rather than maintaining custom deserialization, I think we should generate two types `WebSearchTool` and `WebSearchPreviewTool` and add separate factory methods:

```csharp
public static WebSearchTool CreateWebSearchTool(
    WebSearchToolLocation userLocation = null,
    WebSearchToolContextSize? searchContextSize = null);

public static WebSearchPreviewTool CreateWebSearchPreviewTool(
    WebSearchToolLocation userLocation = null,
    WebSearchToolContextSize? searchContextSize = null);
```

jsquire (approved):
I'd wait for @joseharriaga to be the authoritative reviewer, but it looks reasonable to me.

christothes (approved):

api/OpenAI.net8.0.cs
```diff
@@ -6636,6 +6648,14 @@ public class WebSearchToolApproximateLocation : WebSearchToolLocation, IJsonMode
         public override readonly string ToString();
     }
     [Experimental("OPENAI001")]
+    public class WebSearchToolFilters : IJsonModel<WebSearchToolFilters>, IPersistableModel<WebSearchToolFilters> {
```
joseharriaga, line 6651: Should this new class have JsonPatch too?

joseharriaga (approved):

api/OpenAI.net8.0.cs
jsquire: oh, I missed that.  Yes.   JsonPatch all the things!

ShivangiReja: I hope https://github.com/openai/openai-dotnet/issues/721 is fixed soon so we don’t miss adding JsonPatch per model when they can be enabled once at the namespace level.

Issue #779, opened 2025-10-28
[BUG] Responses: Can't propagate a reasoning item from previous interaction's output into next interaction's input (serialization of 'id')


trrwilson:
### Describe the bug

This is borderline "feature request: don't serialize effective OAS-readonly `id` (/`status`) when using `ResponseItem` as an input model."

## First, REST API description

Items provided by the `output` of a service response include a service-assigned `id` and `status`; these are typically just ignored if they're round-tripped back to the service as input, but there's at least *one* painful exception in the form of reasoning items:

```curl
curl "https://api.openai.com/v1/responses" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "gpt-5-mini",
    "input":[
      {"type":"message","role":"user","content":"hello, world!"},
      {
        "id": "rs_089e8b550e4db875016900178c0b04819c98aed44cccd5465d",
        "type": "reasoning",
        "summary": []
      },
      {
        "id": "msg_089e8b550e4db875016900178d4ad0819cbb6b8b3287b6d9e7",
        "type": "message",
        "status": "completed",
        "content": [
          {
            "type": "output_text",
            "annotations": [],
            "logprobs": [],
            "text": "Hello! How can I help you today?"
          }
        ],
        "role": "assistant"
      },
      {"type":"message","role":"user","content":"try that again, but dramatically"}
    ],
    "store":false
  }'
```

The above, in which the `output` from the first turn is provided as `input` to the second, yields an error:

```bash
{
  "error": {
    "message": "Item with id 'rs_089e8b550e4db875016900178c0b04819c98aed44cccd5465d' not found. Items are not persisted when `store` is set to false. Try again with `store` set to true, or remove this item from your input.",
    "type": "invalid_request_error",
    "param": "input",
    "code": null
  }
}
```

This error disappears if `id` is removed from the reasoning item on the input (even leaving the `id` on the message, which is different from `"store":true`...), indicating that there are at least some circumstances where we can't safely round-trip the ID value and assume it'll be ignored.

## Now the .NET part

This is intensely problematic for multi-turn Responses use impacted by the applied validation, in this case with reasoning items when store is false -- it effectively blocks convenient round-tripping of items. Here's a complete, standalone repro aligned with the above REST:

```csharp
// GUIDANCE: Instructions to run this code: https://aka.ms/oai/net/start
#:package OpenAI@2.*
#:property PublishAot=false
[...]

PR #820, opened 2025-11-07
Add support for included properties (a.k.a. `include`) in Responses

joseharriaga:
In addition to offering the expected functionality, this change also fixes https://github.com/openai/openai-dotnet/issues/779. That is because the failure reported in this issue is by design, and to fix it, users must request the encrypted reasoning content via the included properties like this:

```csharp
ResponseCreationOptions options = new()
{
    StoredOutputEnabled = false,
    IncludedProperties = { IncludedResponseProperty.ReasoningEncryptedContent }
};
```

For more context, see the following docs:

> When using the Responses API in a stateless mode (either with `store` set to `false`, or when an organization is enrolled in zero data retention), you must still retain reasoning items across conversation turns using the techniques described above. But in order to have reasoning items that can be sent with subsequent API requests, each of your API requests must have `reasoning.encrypted_content` in the `include` parameter of API requests ... Any reasoning items in the `output` array will now have an `encrypted_content` property, which will contain encrypted reasoning tokens that can be passed along with future conversation turns.

From: 🔗 https://platform.openai.com/docs/guides/reasoning#encrypted-reasoning-items

### DIFF
--- a/api/OpenAI.net8.0.cs
+++ b/api/OpenAI.net8.0.cs
@@ -5657,9 +5657,9 @@ namespace OpenAI.Responses {
         public virtual Task<ClientResult> DeleteResponseAsync(string responseId, RequestOptions options);
         public virtual Task<ClientResult<ResponseDeletionResult>> DeleteResponseAsync(string responseId, CancellationToken cancellationToken = default);
         public virtual ClientResult GetResponse(string responseId, IEnumerable<IncludedResponseProperty> include, bool? stream, int? startingAfter, bool? includeObfuscation, RequestOptions options);
-        public virtual ClientResult<OpenAIResponse> GetResponse(string responseId, IEnumerable<IncludedResponseProperty> include = null, int? startingAfter = null, bool? includeObfuscation = null, CancellationToken cancellationToken = default);
+        public virtual ClientResult<OpenAIResponse> GetResponse(string responseId, IEnumerable<IncludedResponseProperty> include = null, CancellationToken cancellationToken = default);
         public virtual Task<ClientResult> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> include, bool? stream, int? startingAfter, bool? includeObfuscation, RequestOptions options);
-        public virtual Task<ClientResult<OpenAIResponse>> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> include = null, int? startingAfter = null, bool? includeObfuscation = null, CancellationToken cancellationToken = default);
+        public virtual Task<ClientResult<OpenAIResponse>> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> include = null, CancellationToken cancellationToken = default);
         public virtual CollectionResult<ResponseItem> GetResponseInputItems(string responseId, ResponseItemCollectionOptions options = null, CancellationToken cancellationToken = default);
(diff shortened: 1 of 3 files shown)


src/Generated/OpenAIModelFactory.cs
```diff
@@ -574,6 +574,37 @@ public static McpToolFilter McpToolFilter(IEnumerable<string> toolNames = defaul
             return new McpToolFilter(toolNames.ToList(), isReadOnly, default);
         }
 
+        public static ResponseCreationOptions ResponseCreationOptions(IDictionary<string, string> metadata = default, float? temperature = default, float? topP = default, string endUserId = default, ResponseServiceTier? serviceTier = default, string previousResponseId = default, string model = default, ResponseReasoningOptions reasoningOptions = default, bool? backgroundModeEnabled = default, int? maxOutputTokenCount = default, string instructions = default, ResponseTextOptions textOptions = default, IEnumerable<ResponseTool> tools = default, ResponseToolChoice toolChoice = default, ResponseTruncationMode? truncationMode = default, IEnumerable<ResponseItem> input = default, IEnumerable<IncludedResponseProperty> includedProperties = default, bool? parallelToolCallsEnabled = default, bool? storedOutputEnabled = default, bool? stream = default)
```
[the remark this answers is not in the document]
joseharriaga: Since this code is generated by a tool, readability is not a problem.

src/Custom/Responses/IncludedResponseProperty.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace OpenAI.Responses;
+
+[CodeGenType("Includable")]
```
jsquire, line 3: Man, Krzysztof would be twitching if he saw a member named "Includable"

specification/base/typespec/responses/operations.tsp
```diff
@@ -36,14 +36,19 @@ interface Responses {
   @route("{response_id}")
   getResponse(
     @path
-    @example("resp_677efb5139a88190b512bc3fef8e535d")
     response_id: string,
 
     @query(#{ name: "include[]", explode: true })
-    includables?: Includable[] = #[],
```
jsquire, line 43: WTF even is `#[]` ?

src/Custom/Responses/IncludedResponseProperty.cs
```diff
@@ -0,0 +1,11 @@
...
+public readonly partial struct IncludedResponseProperty
+{
+    [CodeGenMember("MessageInputImageImageUrl")]
+    public static IncludedResponseProperty MessageInputImageUri { get; } = new IncludedResponseProperty(MessageInputImageImageUrlValue);
```
jsquire, line 7: `@@clientName` can't do this?   That seems like something it should cover...

specification/base/typespec/responses/operations.tsp
joseharriaga: It represents a default empty array in TypeSpec.

src/Custom/Responses/IncludedResponseProperty.cs
joseharriaga: I think you're right. I'll do a rename cleanup in a separate PR. We have a few instances of this, and it would be good to migrate it all to client.tsp.

PR #839, opened 2025-11-17
Rename include parameter to includedProperties in the convenience methods

joseharriaga:


src/Custom/Responses/OpenAIResponseClient.cs
```diff
@@ -240,16 +240,16 @@ internal async Task<ClientResult<OpenAIResponse>> GetResponseAsync(string respon
         return ClientResult.FromValue(convenienceResult, protocolResult.GetRawResponse());
     }
 
-    public virtual Task<ClientResult<OpenAIResponse>> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> include = default, CancellationToken cancellationToken = default)
+    public virtual Task<ClientResult<OpenAIResponse>> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> includedProperties = default, CancellationToken cancellationToken = default)
```
ShivangiReja, line 243: Should we rename it in client.tsp as well so the parameter name is updated in the generated protocol methods?

joseharriaga: We want to keep the names of the protocol method parameters identical to the REST API to preserve the protocol as is.

PR #845, opened 2025-11-19
Add support for new Response properties

ShivangiReja:
This PR adds support for the latest Response properties:

* `max_tool_calls`
* `safety_identifier`
* `top_logprobs`
* `conversation`

### DIFF
--- a/api/OpenAI.net8.0.cs
+++ b/api/OpenAI.net8.0.cs
@@ -5663,15 +5663,15 @@ namespace OpenAI.Responses {
         public virtual Task<ClientResult> DeleteResponseAsync(string responseId, RequestOptions options);
         public virtual Task<ClientResult<ResponseDeletionResult>> DeleteResponseAsync(string responseId, CancellationToken cancellationToken = default);
         public virtual ClientResult GetResponse(string responseId, IEnumerable<IncludedResponseProperty> include, bool? stream, int? startingAfter, bool? includeObfuscation, RequestOptions options);
-        public virtual ClientResult<OpenAIResponse> GetResponse(string responseId, IEnumerable<IncludedResponseProperty> include = null, CancellationToken cancellationToken = default);
+        public virtual ClientResult<OpenAIResponse> GetResponse(string responseId, IEnumerable<IncludedResponseProperty> includedProperties = null, CancellationToken cancellationToken = default);
         public virtual Task<ClientResult> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> include, bool? stream, int? startingAfter, bool? includeObfuscation, RequestOptions options);
-        public virtual Task<ClientResult<OpenAIResponse>> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> include = null, CancellationToken cancellationToken = default);
+        public virtual Task<ClientResult<OpenAIResponse>> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> includedProperties = null, CancellationToken cancellationToken = default);
         public virtual CollectionResult<ResponseItem> GetResponseInputItems(string responseId, ResponseItemCollectionOptions options = null, CancellationToken cancellationToken = default);
         public virtual CollectionResult GetResponseInputItems(string responseId, int? limit, string order, string after, string before, RequestOptions options);
         public virtual AsyncCollectionResult<ResponseItem> GetResponseInputItemsAsync(string responseId, ResponseItemCollectionOptions options = null, CancellationToken cancellationToken = default);
         public virtual AsyncCollectionResult GetResponseInputItemsAsync(string responseId, int? limit, string order, string after, string before, RequestOptions options);
-        public virtual CollectionResult<StreamingResponseUpdate> GetResponseStreaming(string responseId, IEnumerable<IncludedResponseProperty> include = null, int? startingAfter = null, bool? includeObfuscation = null, CancellationToken cancellationToken = default);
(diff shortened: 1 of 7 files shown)


api/OpenAI.net8.0.cs
```diff
@@ -5598,6 +5600,7 @@ public class OpenAIResponse : IJsonModel<OpenAIResponse>, IPersistableModel<Open
         public ResponseIncompleteStatusDetails IncompleteStatusDetails { get; }
         public string Instructions { get; }
         public int? MaxOutputTokenCount { get; }
+        public int? MaxToolCallCount { get; }
```
joseharriaga, line 5604: In addition to `MaxToolCallCount`, the other properties should also be echoed back in the Response here. In other words, we should also add `ConversationId`, `SafetyIdentifier`, and `TopLogProbabilityCount` to this output class, as wel as the corresponding factory method in the `OpenAIResponseModelFactory`.

specification/base/typespec/common/models.tsp
```diff
@@ -265,6 +268,10 @@ model ModelResponsePropertiesForRequest {
   /** A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse. [Learn more](/docs/guides/safety-best-practices#end-user-ids). */
   user?: string;
 
+  /**A stable identifier used to help detect users of your application that may be violating OpenAI's usage policies.
+  The IDs should be a string that uniquely identifies each user. We recommend hashing their username or email address, in order to avoid sending us any identifying information. [Learn more](https://platform.openai.com/docs/guides/safety-best-practices#safety-identifiers).*/
+  safety_identifier?: string;
```
joseharriaga, line 273: I suspect that part of the fix to my other comment is to add these properties to the `ModelResponsePropertiesForResponse` model too.

specification/base/typespec/responses/models.tsp
```diff
@@ -87,6 +87,25 @@ model CreateResponse {
    * for more information.
    */
   stream?: boolean | null = false;
+
+  /** The conversation that this response belongs to.
+   *  Items from this conversation are prepended to input_items for this response request.
+   *  Input items and output items from this response are automatically added to this conversation after this response completes. */
+  conversation?: ConversationParam | null;
```
joseharriaga, line 94: This property also needs to be added to the `Response` output model below. 

There is an interesting situation here though: 
* In the API reference docs, this property is an object with a single "id" property (just like we saw in the input). 
* However, in the latest spec, this property is defined as an object with the "id" property, but with a few other properties too, like "metadata", "created_at", etc. 

So now, the question is: Which one is correct? 

To find the answer, could you verify by testing against the service and seeing what it returns? Based on what you find, I see two paths:
1. If it returns just the "id" property, we can expose this property as a single string property called `ConversationId` exactly like you did in the input model. Additionally, we should also file a bug against the latest spec. 
2. But if it returns more than that just the "id", then we might need to expose a full `Conversation` object with all the other properties.

src/Generated/Models/Responses/ResponseCreationOptions.cs
```diff
@@ -57,10 +61,14 @@ internal ResponseCreationOptions(IDictionary<string, string> metadata, float? te
...
 
+        public string SafetyIdentifier { get; set; }
+
         public ResponseServiceTier? ServiceTier { get; set; }
 
         public string PreviousResponseId { get; set; }
 
         public string Instructions { get; set; }
+
+        public string ConversationId { get; set; }
```
joseharriaga, line 72: Could you add tests and/or samples for using the `ConversationId`? Here's more info about the main scenario, which would be nice to validate:
🔗 https://platform.openai.com/docs/guides/conversation-state?api-mode=responses#using-the-conversations-api

joseharriaga (changes requested):

specification/base/typespec/responses/models.tsp
ShivangiReja: I tested this and the conversation object in the response only contains the id field, none of the additional properties from the latest spec show up.

I’ve updated the Response output model accordingly and created an issue https://github.com/microsoft/openai-openapi-pr/issues/283 for the spec mismatch.

tests/Utility/TestHelpers.cs
```diff
@@ -70,6 +72,7 @@ public enum TestScenario
         TestScenario.Realtime => "gpt-4o-realtime-preview-2024-10-01",
         TestScenario.Responses => "gpt-4o-mini",
         TestScenario.Containers => "gpt-4o-mini",
+        TestScenario.Conversations => "gpt-4o-mini",
```
joseharriaga: nit: My understanding is that Conversations doesn't need a model, so this should be `null` instead.

tests/Responses/ResponsesTests.cs
```diff
@@ -522,6 +523,41 @@ public async Task CanOptOutOfStorage()
...
+        OpenAIResponseClient client = GetTestClient("gpt-4.1");
+        OpenAIResponse response = await client.CreateResponseAsync(
+            [ResponseItem.CreateUserMessageItem("Hello, model!")],
+            new ResponseCreationOptions()
+            {
+                ConversationId = conversationId,
+            });
+
+        Assert.That(response, Is.Not.Null);
+        Assert.That(response.ConversationId, Is.EqualTo(conversationId));
```
joseharriaga, line 558: nit: Could you add one more part to this test where you call `GetConversationItems` from the `ConversationClient` and validate that items were successfull added after you create the Response?

tests/Utility/TestHelpers.cs
ShivangiReja: Yes, I just copied it from the line above 🤣

PR #856, opened 2025-12-04
[Spec Ingestion] Embeddings, Images, and Models

joseharriaga:
* Ingested latest TypeSpec for Embeddings, Images, and Models
* Added `NumericPropertiesVisitor` to handle TypeSpec's `integer` type
* Added `ItemsPropertyVisitor` to handle serialization of classes that were customized to inherit from `ReadOnlyCollection<T>`
* Added `Background`, `OutputFileFormat`, `Quality` and `Size` properties to `GeneratedImageCollection`
* Disambiguated between `HD` and `High` values in the `GeneratedImageQuality` enum
* Added `InputFidelity`, `OutputCompressionFactor`, and `OutputFileFormat` properties to `ImageEditOptions`
* Added `OutputTokenDetails` property to `ImageTokenUsage`


api/OpenAI.net8.0.cs
```diff
@@ -3659,6 +3667,7 @@ public class GeneratedImageCollection : ObjectModel.ReadOnlyCollection<Generated
         public GeneratedImageQuality(string value);
         [Experimental("OPENAI001")]
         public static GeneratedImageQuality Auto { get; }
+        public static GeneratedImageQuality HD { get; }
```
ShivangiReja, line 3671: Should this be marked [Experimental]?

src/Custom/Images/GeneratedImageCollection.cs
```diff
@@ -1,44 +1,61 @@
...
+    [CodeGenMember("Background")]
+    public GeneratedImageBackground? Background { get; }
+
+    // CUSTOM:
+    // - Renamed.
+    // - Changed property type.
+    /// <summary> The format in which the generated images are returned. </summary>
+    [Experimental("OPENAI001")]
+    [CodeGenMember("OutputFormat")]
+    public GeneratedImageFileFormat? OutputFileFormat { get; set; }
```
ShivangiReja: This looks like an output model. Should `OutputFileFormat` be read-only?

codegen/generator/src/Visitors/NumericPropertiesVisitor.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        "OpenAI.Chat.ChatCompletionOptions.Seed",
+
+        "OpenAI.LegacyCompletions.InternalCreateCompletionRequest.Seed",
+
+        "OpenAI.Images.ImageTokenUsage.InputTokenCount",
+        "OpenAI.Images.ImageTokenUsage.OutputTokenCount",
+        "OpenAI.Images.ImageTokenUsage.TotalTokenCount",
+        "OpenAI.Images.ImageInputTokenUsageDetails.ImageTokens",
+        "OpenAI.Images.ImageInputTokenUsageDetails.TextTokens",
```
ShivangiReja: There is an inconsistency here between using the final C# property names vs the original generated names. 

`ImageTokenUsage` entries use the renamed properties (e.g. `InputTokenCount`), but `ImageInputTokenUsageDetails` entries use the original names (`TextTokens`, `ImageTokens`).

api/OpenAI.net8.0.cs
joseharriaga: Yes, good catch!

src/Custom/Images/GeneratedImageCollection.cs
joseharriaga: Yes, thank you!

codegen/generator/src/Visitors/NumericPropertiesVisitor.cs
joseharriaga: These were left over from when I was prototyping and need to be removed. Good catch!

PR #861, opened 2025-12-08
Clone options fully for ChatCompletion calls

### LINKED ISSUE #436: Race Condition: ChatClient.CompleteChatAsync Mutates Shared ChatCompletionOptions, Causing Incorrect Request Payloads in Parallel Calls
### Service

OpenAI

### Describe the bug

When using `ChatClient.CompleteChatAsync` (or `CompleteChatStreamingAsync`) within a parallel loop (e.g., `Parallel.ForEachAsync`), if the same `ChatCompletionOptions` instance is passed to multiple concurrent calls, a race condition occurs. The internal method `CreateChatCompletionOptions` (or a similar method responsible for preparing the request) modifies the `Messages` property (and potentially others) of the passed-in `ChatCompletionOptions` object by reference.

This leads to a scenario where the `Messages` (and thus the content of the API request) for one parallel operation can be overwritten by another before the request is fully constructed and sent. As a result, multiple distinct input messages can end up being processed by the OpenAI API as if they were the same, often the message from the last thread to modify the shared `options` object before requests are dispatched.

### Steps to reproduce

Image

### Code snippets

```C#
using Azure.AI.OpenAI;
using Azure.Identity;
using OpenAI.Chat;

ChatCompletionOptions _options = new()
{
    Temperature = 0,
};

AzureOpenAIClient _azureClient = new(new Uri("https://learn-shared-aoai-test-eus2.openai.azure.com/"), new DefaultAzureCredential(), new AzureOpenAIClientOptions());

ChatClient _chatClient = _azureClient.GetChatClient("gpt-4.1-2025-04-14");

IReadOnlyList<string> _messages =
[
    "2 + 4 = ?",
    "3 + 5 = ?",
    "4 + 6 = ?"
];

await Parallel.ForEachAsync(_messages, async (message, cancellationToken) =>
{
    List<ChatMessage> messages =
    [
        new UserChatMessage(message),
    ];

    ChatCompletion completion = await _chatClient.CompleteChatAsync(messages, _options, cancellationToken);

    Console.WriteLine($"Message: {message}, Response: {completion.Content[0].Text}");
});
```

### OS

windows

### .NET version

9.0.204

### Library version

2.1.0

christothes:
fixes #436


jsquire (approved):

src/Custom/Chat/ChatCompletionOptions.cs
```diff
@@ -215,4 +216,44 @@ public ChatResponseModalities ResponseModalities
...
+        foreach (var t in Tools) clone.Tools.Add(t);
+        clone.Temperature = Temperature;
+        clone.TopP = TopP;
+        clone.SafetyIdentifier = SafetyIdentifier;
+        clone.ServiceTier = ServiceTier;
+        clone.FrequencyPenalty = FrequencyPenalty;
+        clone.PresencePenalty = PresencePenalty;
+        clone.WebSearchOptions = WebSearchOptions;
+        clone.Seed = Seed;
+        return clone;
```
joseharriaga, line 256: Would using [`MemberwiseClone`](https://learn.microsoft.com/en-us/dotnet/api/system.object.memberwiseclone?view=net-10.0) work here? It would be nice to not have to update this method every time a new property is added. 

We are currently using `MemberwiseClone` in Responses:
https://github.com/openai/openai-dotnet/blob/aa8b0342177259a6d62bc2c35fec7eb4792ce782/src/Custom/Responses/CreateResponseOptions.cs#L125-L130

christothes: I'm not sure if it would do the same thing with the collection properties.

jsquire: It won't.   It doesn't do a great job of even a shallow clone inline.   We tried this with the messaging libraries.

joseharriaga: I think you are correct that it won't be exactly the same: The current implementation creates a new collection and adds references to the existing items, while `MemberwiseClone` will simply copy the reference to the existing collection. 

I'm thinking that `MemberwiseClone` might be sufficient because the intention here is not to make `ChatCompletionOptions` thread-safe, but rather to simply avoid modifying it. In other words, the user's instance of `ChatCompletionOptions` should be unchanged after they call `CompleteChat`. The reason why this doesn't work today is because we're trampling over a few internal properties like `Messages`, `Stream`, etc. But if we `MemberwiseClone` it, we are free to modify the internal properties of the clone while keeping the user's instance intact.

christothes: The customer issue motivating this change specifically deals with concurrency issues. Taking the memberwise clone approach would not resolve the issue, I believe.

joseharriaga: I think `MemberwiseClone` would solve the issue because the user is specifically talking about the `Messages` property, which is one of the internal properties that we're trampling over. By design, our models are not guaranteed to be thread-safe (only our clients are thread-safe), so if the user were asking us to make it thread-safe, I think we would choose not to fix it. However, in this particular case, I think this is still a valid issue because the customer is wondering: "If I'm not modifying the `ChatCompletionOptions` instance at all, shouldn't I be able to use it across requests even if it's not thread-safe?". And the reason why it doesn't work like that is because it is us who are modifying it as part of `CompletChat`, which we shouldn't do because the `ChatCompletionOptions` parameter is not a `ref` parameter.

Here's another way of thinking about this issue separate from concurrency: 
1. First, the user calls `ModelReaderWriter.Write` to serialize an instance of `ChatCompletionOptions`
2. Then, the user calls `CompleteChat` and passes this instance of `ChatCompletionOptions` as an argument
3. Finally, if the user were to call `ModelReaderWriter.Write` to serialize the same instance of `ChatCompletionOptions` again, the result should be the same as in Step 1. However, this is currently not the case because we are changing a handful of internal properties that appear in the serialization.

jsquire: We should not be mutating ANY of the options in an operation call.   Ever.    That's a risk for any concurrent calls to the client, which violates our thread safety guarantees.   Even if it works today, its by coincidence and not design.   I don't want to carry that debt forward.

Cloning the options is something that we should do for client-level options to protect us from the caller mutating after instantiation.   We shouldn't have to clone on an operation because we should be treating these as immutable.   We need to find a better way to set the per-request model.   The underlying pattern is a problem.

In the meantime, I'd rather we go with the heavier option that gives us full safety and concentrate efforts on fixing the client.

joseharriaga:
Could you do the same fix for the Responses API? It can be in a separate PR. 🙂

src/Custom/Chat/ChatClient.cs
```diff
@@ -381,19 +381,24 @@ public virtual ClientResult<ChatCompletionDeletionResult> DeleteChatCompletion(s
...
+    private ChatCompletionOptions CreateChatCompletionOptions(IEnumerable<ChatMessage> messages, ChatCompletionOptions options, bool stream = false)
     {
-        options.Messages = messages.ToList();
-        options.Model = _model;
+        var clonedOptions = options.Clone();
+        foreach (var message in messages)
+        {
+            clonedOptions.Messages.Add(message);
+        }
+        clonedOptions.Model ??= _model;
```
stephentoub, line 391: In 2.8.0 we were able to use JsonPatch on a ChatCompletionOptions to set the Model. Now in 2.9.0, that's being completely ignored and we're no longer able to set the model. I'm guessing it's because of this change. Was this intentional?

jsquire: No, that's not intentional.   Looks like we forgot to include the patch in the clone.   I've opened #988 to track.

PR #862, opened 2025-12-08
[EngSys] .NET 10 Migration

jsquire:
# Summary

The focus of these changes is to move the repository to .NET 10, including the SDK, target frameworks, and BCL package dependencies.


tests/OpenAI.Tests.csproj
```diff
@@ -1,28 +1,33 @@
...
-    <PackageReference Include="System.Linq.Async" Version="6.0.1" />
-		<PackageReference Include="Microsoft.ClientModel.TestFramework" Version="1.0.0-alpha.20250919.1" />
+    <PackageReference Include="NUnit" Version="4.4.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="6.0.0" />
 	</ItemGroup>
 
+  <!-- Conditionally include packages built into the latest runtime -->
+  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net10.0')) == 'false'">
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="10.0.0" />
+    <PackageReference Include="System.Linq.Async" Version="7.0.0" />
```
stephentoub: Can you reference System.Linq.AsyncEnumerable directly instead?

jsquire: It triggers an `NU1510` for `net10.0` but we could throw it behind the conditional if there's a benefit to that over the current package.   I went for minimal churn.

.github/workflows/live-test.yml
```diff
@@ -17,6 +17,11 @@ jobs:
       version_suffix_args: ${{ format('/p:VersionSuffix="alpha.{0}"', github.run_number) }}
     steps:
       - name: Setup .NET
```
christothes: Might as well be descriptive here too. 

```suggestion
      - name: Setup .NET 10
```

tests/OpenAI.Tests.csproj
stephentoub: I'm surprised given this is already guarded with a conditional to not be used on net10.0.

jsquire: I'll check it again here, but I recall that being one of the explicit removals we had to add for a similar upgrade due to the analyzer failure.

jsquire: Interesting.   It seems to work here just fine.

.github/workflows/live-test.yml
jsquire: Updated all instances.

PR #865, opened 2025-12-09
Apply protocol model pattern to missing types in Responses

### LINKED ISSUE #863: Modify ImplicitConversionToBinaryContentVisitor to allow for exception list
We have this [visitor[(https://github.com/openai/openai-dotnet/blob/main/codegen/generator/src/Visitors/ImplicitConversionToBinaryContentVisitor.cs) which removes the generated BinaryContent implicit conversion because it wasn't useful for any of our models:

However, this is definitely useful for our protocol models. Let's modify the visitor so that we can keep a list of types that we can "opt-out" of the visitor, so that this code gets generated instead of having to keep it as custom code.

joseharriaga:
- Added a visitor to generate the protocol model pattern: 1) properties with setters, 2) public default constructor, 3) implicit conversion from `BinaryContent` for input models.
- Applied protocol model pattern to missing types in Responses, including the derived types of `StreamingResponseUpdate`, `ResponseDeletionResult`, and `ResponseItemCollectionPage`.
- Fixed an issue with `CreateResponse` methods overriding the value of `StreamingEnabled`. It is now the responsibility of the user to correctly set `StreamingEnabled` based on if they are calling the streaming or non-streaming method.
- Added validation that checks the value of `StreamingEnabled` based on which method it is being used with and throws an exception in case of invalid combinations.
- Changed the `ConversationId` property in `CreateResponseOptions` to match the `ConversationOptions` property of `ResponseResult`, i.e. type is now `ResponseConversationOptions` instead of just a `string`.
- Renamed `MaxToolCalls` in `CreateResponseOptions` to `MaxToolCallCount` to match the property in `ResponseResult`.
- Renamed `Id` in `ResponseDeletionResult` to `ResponseId` for clarity and consistency with other APIs in the library.
- fixes #863

### DIFF
--- a/api/OpenAI.net8.0.cs
+++ b/api/OpenAI.net8.0.cs
@@ -6103,11 +6103,13 @@ namespace OpenAI.Responses {
         public virtual Task<ClientResult> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> include, bool? stream, int? startingAfter, bool? includeObfuscation, RequestOptions options);
         public virtual Task<ClientResult<ResponseResult>> GetResponseAsync(string responseId, CancellationToken cancellationToken = default);
         public virtual ClientResult<ResponseItemCollectionPage> GetResponseInputItemCollectionPage(ResponseItemCollectionOptions options, CancellationToken cancellationToken = default);
+        public virtual ClientResult GetResponseInputItemCollectionPage(string responseId, int? limit, string order, string after, string before, RequestOptions options);
         public virtual Task<ClientResult<ResponseItemCollectionPage>> GetResponseInputItemCollectionPageAsync(ResponseItemCollectionOptions options, CancellationToken cancellationToken = default);
+        public virtual Task<ClientResult> GetResponseInputItemCollectionPageAsync(string responseId, int? limit, string order, string after, string before, RequestOptions options);
         public virtual CollectionResult<ResponseItem> GetResponseInputItems(ResponseItemCollectionOptions options, CancellationToken cancellationToken = default);
-        public virtual ClientResult GetResponseInputItems(string responseId, int? limit, string order, string after, string before, RequestOptions options);
+        public virtual CollectionResult<ResponseItem> GetResponseInputItems(string responseId, CancellationToken cancellationToken = default);
         public virtual AsyncCollectionResult<ResponseItem> GetResponseInputItemsAsync(ResponseItemCollectionOptions options, CancellationToken cancellationToken = default);
-        public virtual Task<ClientResult> GetResponseInputItemsAsync(string responseId, int? limit, string order, string after, string before, RequestOptions options);
+        public virtual AsyncCollectionResult<ResponseItem> GetResponseInputItemsAsync(string responseId, CancellationToken cancellationToken = default);
         public virtual CollectionResult<StreamingResponseUpdate> GetResponseStreaming(GetResponseOptions options, CancellationToken cancellationToken = default);
         public virtual CollectionResult<StreamingResponseUpdate> GetResponseStreaming(string responseId, CancellationToken cancellationToken = default);
         public virtual AsyncCollectionResult<StreamingResponseUpdate> GetResponseStreamingAsync(GetResponseOptions options, CancellationToken cancellationToken = default);
--- a/api/OpenAI.netstandard2.0.cs
+++ b/api/OpenAI.netstandard2.0.cs
@@ -5372,11 +5372,13 @@ namespace OpenAI.Responses {
         public virtual Task<ClientResult> GetResponseAsync(string responseId, IEnumerable<IncludedResponseProperty> include, bool? stream, int? startingAfter, bool? includeObfuscation, RequestOptions options);
         public virtual Task<ClientResult<ResponseResult>> GetResponseAsync(string responseId, CancellationToken cancellationToken = default);
         public virtual ClientResult<ResponseItemCollectionPage> GetResponseInputItemCollectionPage(ResponseItemCollectionOptions options, CancellationToken cancellationToken = default);
+        public virtual ClientResult GetResponseInputItemCollectionPage(string responseId, int? limit, string order, string after, string before, RequestOptions options);
         public virtual Task<ClientResult<ResponseItemCollectionPage>> GetResponseInputItemCollectionPageAsync(ResponseItemCollectionOptions options, CancellationToken cancellationToken = default);
+        public virtual Task<ClientResult> GetResponseInputItemCollectionPageAsync(string responseId, int? limit, string order, string after, string before, RequestOptions options);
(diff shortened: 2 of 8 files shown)


codegen/generator/src/Visitors/ProtocolModelVisitor.cs
```diff
@@ -0,0 +1,125 @@
...
+{
+    // Classes that are protocol models.
+    private static readonly HashSet<string> _protocolModels = new(StringComparer.OrdinalIgnoreCase)
+    {
+        "OpenAI.Responses.CreateResponseOptions",
+        "OpenAI.Responses.ResponseDeletionResult",
+        "OpenAI.Responses.ResponseItemCollectionPage",
+        "OpenAI.Responses.ResponseResult",
+
+        // "OpenAI.Responses.StreamingResponseUpdate",
```
christothes: Is this intentional?

codegen/generator/src/OpenAILibraryGenerator.cs
```diff
@@ -27,12 +27,13 @@ protected override void Configure()
             AddVisitor(new OpenAILibraryVisitor());
             AddVisitor(new VirtualMessageCreationVisitor());
             AddVisitor(new ProhibitedNamespaceVisitor());
-            AddVisitor(new ImplicitConversionToBinaryContentVisitor());
+            // AddVisitor(new ImplicitConversionToBinaryContentVisitor());
```
christothes: Should we remove this and the visitor?

src/Custom/Responses/Streaming/StreamingResponseContentPartAddedUpdate.cs
```diff
@@ -1,14 +1,10 @@
...
-// - Added Experimental attribute.
-// - Renamed.
+// CUSTOM: Renamed.
 [CodeGenType("ResponseContentPartAddedEvent")]
 public partial class StreamingResponseContentPartAddedUpdate
 {
     // CUSTOM: Apply generalized content type.
     [CodeGenMember("Part")]
-    public ResponseContentPart Part { get; }
+    public ResponseContentPart Part { get; set; }
```
christothes, line 9: Will the `ProtocolModelVisitor` do this for us?

src/Custom/Responses/Streaming/StreamingResponseContentPartDoneUpdate.cs
```diff
@@ -1,14 +1,10 @@
...
-// - Added Experimental attribute.
-// - Renamed.
+// CUSTOM: Renamed.
 [CodeGenType("ResponseContentPartDoneEvent")]
 public partial class StreamingResponseContentPartDoneUpdate
 {
     // CUSTOM: Apply generalized content type.
     [CodeGenMember("Part")]
-    public ResponseContentPart Part { get; }
+    public ResponseContentPart Part { get; set; }
```
christothes, line 9: Will the `ProtocolModelVisitor` do this for us?

src/Custom/Responses/Streaming/StreamingResponseUpdate.cs
```diff
@@ -1,12 +1,11 @@
...
-// - Added Experimental attribute.
-// - Renamed.
+// CUSTOM: Renamed.
 [CodeGenType("ResponseStreamEvent")]
 [CodeGenSuppress("StreamingResponseUpdate", typeof(System.ClientModel.ClientResult))]
 public partial class StreamingResponseUpdate
 {
+    // CUSTOM: Added setter because this is a property of protocol model.
+    [CodeGenMember("SequenceNumber")]
+    public int SequenceNumber { get; set; }
```
christothes: Will the `ProtocolModelVisitor` do this for us?

api/OpenAI.net8.0.cs
```diff
@@ -5896,13 +5900,12 @@ public class ResponseItem : IJsonModel<ResponseItem>, IPersistableModel<Response
...
     [Experimental("OPENAI001")]
     public class ResponseItemCollectionOptions : IJsonModel<ResponseItemCollectionOptions>, IPersistableModel<ResponseItemCollectionOptions> {
-        public ResponseItemCollectionOptions();
         public ResponseItemCollectionOptions(string responseId);
         public string AfterId { get; set; }
         public string BeforeId { get; set; }
         public ResponseItemCollectionOrder? Order { get; set; }
         public int? PageSizeLimit { get; set; }
-        public string ResponseId { get; set; }
+        public string ResponseId { get; }
```
christothes, line 5908: Is this intentional?

codegen/generator/src/Visitors/ProtocolModelVisitor.cs
joseharriaga: I originally excluded `StreamingResponseUpdate` from this list because the visitor would make its default constructor public, but because this is the base type in a hierarchy, I think its default constructor should remain internal. Instead, the default constructor must be exposed in the derived types only. With Jorge's help, I found a fix to be able to do this with the visitor, and `StreamingResponseUpdate` can be included in this list. 🙂

codegen/generator/src/OpenAILibraryGenerator.cs
joseharriaga: Yes. Fixed!

src/Custom/Responses/Streaming/StreamingResponseUpdate.cs
joseharriaga: This is now possible with the fix I just added to the visitor to be able to handle `StreamingResponseUpdate`. Fixed!

src/Custom/Responses/Streaming/StreamingResponseContentPartAddedUpdate.cs
joseharriaga: It will once this property is moved to generated code, but for now it needs to remain in custom code for reasons unrelated to this change (i.e., to apply the generalized ResponseContentPart type as per the CUSTOM comment). I suspect we can fix it with client.tsp, but we can do it separately in the future.

src/Custom/Responses/Streaming/StreamingResponseContentPartDoneUpdate.cs
joseharriaga: It will once this property is moved to generated code, but for now it needs to remain in custom code for reasons unrelated to this change (i.e., to apply the generalized ResponseContentPart type as per the CUSTOM comment). I suspect we can fix it with client.tsp, but we can do it separately in the future.

api/OpenAI.net8.0.cs
joseharriaga: For now, yes. Here's my reasoning: `ResponseItemCollectionOptions` is technically not a protocol model. Instead, it is a property bag that we invented to wrap path and query parameter. At this moment, we're still unsure what path we want to take with these: Should they behave like "status quo" models? Or should they behave like protocol models? Or should they behave in their own way?

The change I'm making here is to make them behave the "status quo" way, which I think is the safest for now while we close on the design.

christothes (approved):

PR #870, opened 2025-12-11
Restore `Model` in `ResponsesClient`

joseharriaga:
* Restored `Model` in `ResponseClient`
* Converted `ResponseTokenUsage`, `ResponseInputTokenUsageDetails`, and `ResponseOutputTokenUsageDetails` into protocol models
* Fixed examples

### DIFF
--- a/src/Custom/Responses/ResponsesClient.cs
+++ b/src/Custom/Responses/ResponsesClient.cs
@@ -3,7 +3,6 @@ using System.ClientModel;
 using System.ClientModel.Primitives;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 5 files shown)


api/OpenAI.net8.0.cs
```diff
@@ -36,7 +36,7 @@ public class OpenAIClient {
         [Experimental("OPENAI002")]
         public virtual RealtimeClient GetRealtimeClient();
         [Experimental("OPENAI001")]
-        public virtual ResponsesClient GetResponsesClient();
+        public virtual ResponsesClient GetResponsesClient(string model);
```
christothes, line 39: Did we decide not to go with `defaultModel`?

joseharriaga: It sounds like we will want to move this property into `ResponsesClientOptions`, which doesn't exist yet because we're waiting on the Config feature from System.ClientModel that Michael is working on. With the intention to minimize churn for users, I'm simply reverting the `Model` property to what we had before for now, and we can perform the change as a whole in the future.

PR #877, opened 2025-12-16
Add reasoning effort 'none' for GPT 5.1 and onward

jpaarhuis:
### Summary

This PR adds support for the `none` value in the `reasoning_effort` parameter, which was introduced for GPT-5.1 and newer models.

### Changes

- Added `"none"` to the `ReasoningEffort` union in the TypeSpec specification
- Updated documentation to reflect that this applies to "Reasoning models" (not just o-series)
- Regenerated code to include `ChatReasoningEffortLevel.None` and `ResponseReasoningEffortLevel.None`

### Background

According to the [OpenAI API documentation](https://platform.openai.com/docs/api-reference/chat/create#chat_create-reasoning_effort), the `reasoning_effort` parameter now supports the following values: `none`, `minimal`, `low`, `medium`, and `high`.

### Usage

```csharp
var options = new ChatCompletionOptions
{
    ReasoningEffortLevel = ChatReasoningEffortLevel.None
};
```

### DIFF
--- a/api/OpenAI.net8.0.cs
+++ b/api/OpenAI.net8.0.cs
@@ -2090,6 +2090,7 @@ namespace OpenAI.Chat {
         public static ChatReasoningEffortLevel Low { get; }
         public static ChatReasoningEffortLevel Medium { get; }
         public static ChatReasoningEffortLevel Minimal { get; }
+        public static ChatReasoningEffortLevel None { get; }
         public readonly bool Equals(ChatReasoningEffortLevel other);
         [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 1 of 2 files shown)


christothes:
Hi @jpaarhuis - 
One step you'll need to complete before this is merged is to run the [Export-API](https://github.com/openai/openai-dotnet/blob/main/scripts/Export-Api.ps1) script to update the public API surface file.

You'll need the GenAPI tools which can be found here https://github.com/dotnet/sdk/tree/main/src/Compatibility/GenAPI

jpaarhuis:
Ok @christothes, just ran that and committed the additional code.

PR #885, opened 2026-01-06
Update sample codes for v2.8.0 breaking changes

justinyoo:
## Background
The v2.8.0 release introduced breaking changes in the public surface area used by our documentation samples (notably around Responses, tools, and related usage patterns). This PR updates the docs/quickstart snippets so they stay copy/paste-able and match the current SDK.

## List of changes
- Refresh Responses-based samples across docs and quickstarts to align with the v2.8.0 API surface (tools/function calling, web search, file search, remote MCP, approvals, streaming, and image analysis snippets).
- Update docs SDK guidance to target stable .NET 10 (update install instructions and pin docs/global.json to 10.0.100 with roll-forward behavior).
- Add an internal agent definition to support consistent future SDK migration work.

## How to verify the changes
- Ensure a .NET 10 SDK is installed and active:
  - `dotnet --version` (should be 10.x; docs guidance now targets 10.0.100)
- Ensure the OpenAI API key is stored as an environment variable, `OPENAI_API_KEY`.
- Run individual sample codes respectively by:
  - `dotnet run ./docs/overview/responses/developer_quickstart.cs`, for example.
- Spot-check that updated docs snippets compile when copied into a small console app that references the current package version (especially the Responses + tools examples).

### DIFF
--- a/docs/global.json
+++ b/docs/global.json
@@ -1,7 +1,0 @@
-{
-  "sdk": {
-    "version": "10.0.100",
-    "rollForward": "latestFeature",
-    "allowPrerelease": false
-  }
-}


.github/agents/expert-sdk-migration.agent.md
```diff
@@ -0,0 +1,51 @@
+---
+description: "Provide expert .NET software engineering guidance using modern software design patterns."
```
jsquire: I think we want to remove this file.

docs/global.json
```diff
@@ -1,5 +1,7 @@
 {
```
jsquire: Please remove the file.   This should bubble up to the root global now that the repo is unified on .NET 10.

justinyoo: Oops! Removed.

docs/global.json
jsquire, line 1: Can we please remove the file?   There's a global at the root that should apply to all areas - including samples.   That'll ensure that we can manage the SDK version in a single location.

justinyoo: Removed! It was there due to the discrepancy of SDK and the sample code targeting different versions back in the day. Now, it's gone.

jsquire (approved):
Thanks, @justinyoo!    LGTM

PR #887, opened 2026-01-06
Update net9.0 to 10.0 and add correct using statements

MaiLinhP:
Latest codegen moved some attributes from namespace OpenAI to namespace Microsoft.TypeSpec.Generator.Customizations. This PR adds the using statements to files that use these attributes.

Also upgraded TargetFramework from net9.0 to net10.0.

### DIFF
--- a/src/Custom/Administration/Internal/GeneratorStubs.cs
+++ b/src/Custom/Administration/Internal/GeneratorStubs.cs
@@ -1,3 +1,5 @@
+using Microsoft.TypeSpec.Generator.Customizations;
+
 namespace OpenAI.Administration;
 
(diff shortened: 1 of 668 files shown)


src/Custom/Administration/Internal/GeneratorStubs.cs
```diff
@@ -1,3 +1,5 @@
+using Microsoft.TypeSpec.Generator.Customizations;
```
jorgerangel-msft, line 1: We are missing the generated updates to the CodeGen attributes

jorgerangel-msft (commented):
Please also include the updates to bump the generator version, similar to [this](https://github.com/openai/openai-dotnet/pull/883/files#diff-a8982328bdcbe070d823bf6b522131c1ca0a8e12d125d78d00d7897e59058c17).

Issue #850, opened 2025-11-26
[BUG] CodeInterpreterCallImageOutput.DeserializeCodeInterpreterCallImageOutput throws invalid url length


shanepowell:
### Describe the bug

On the line with new Uri(item.Value.GetString()) where the item.Value.GetString() returns: "data:image/png;base64,long base64 string"

i.e.:
			if (item.NameEquals("url"u8))
			{
				imageUri = new Uri(item.Value.GetString());
				continue;
			}

I'm using OpenAI package version 2.7.0.

This using the gpt5.1 model with medium reasoning effort using the question:
Parse the time and city from the airport board shown in this image into a list.

I can provide the image if useful.

### Steps to reproduce

This using the gpt5.1 model with medium reasoning effort against the responses API using the following question:
Parse the time and city from the airport board shown in this image into a list.

I can provide the image if useful. The image is sent the modal as a embedded image part after the text part above.

### Code snippets

```C#

```

### OS

Windows

### .NET version

9

### Library version

2.7.0

jsquire:
Hi @shanepowell.  Thanks for reaching out and we regret that you're experiencing difficulties.   The behavior that you're inquiring about is a platform limitation with the `Uri` constructor, not something that the OpenAI client has insight into nor influence over.   The limitation was removed in .NET 10 via [dotnet/runtime/117287](https://github.com/dotnet/runtime/pull/117287).    For more context, please see [BUG: MessageContent.FromImageUrl(Uri) incompatible with "data url" (#135)](https://github.com/openai/openai-dotnet/issues/135)

In order to use the longer data URLs returned by the OpenAI service, the recommended mitigation is to run against .NET 10.   If you're unable to do so, the best approach on .NET 9 would be to use the protocol overload for the operation and read the data URL directly a s a string.   A protocol example for the code interpreter tool can be found in [this comment](https://github.com/openai/openai-dotnet/issues/552#issuecomment-3105233071) and an example of parsing responses to extract a base64 image can be found in [this comment](https://github.com/openai/openai-dotnet/issues/561#issuecomment-3130032021).

shanepowell:
That is not data coming from me, but the openai code interpreter tool via the responses api.  So the bug is that the OpenAi library does not support embedded images that the openai code interpreter tool can produce.

jsquire:
> That is not data coming from me, but the openai code interpreter tool via the responses api. So the bug is that the OpenAi library does not support embedded images that the openai code interpreter tool can produce.

[...]

PR #888, opened 2026-01-08
[Spec Ingestion] Moderations

ShivangiReja:
This PR updates the following:
* Ingested latest TypeSpec for Moderations.
* Added support for multimodal input, enabling moderation of both text and image content.


api/OpenAI.net8.0.cs
```diff
@@ -3984,13 +3994,37 @@ public class ModerationClient {
         [Experimental("OPENAI001")]
         public string Model { get; }
         public ClientPipeline Pipeline { get; }
+        [Experimental("OPENAI001")]
+        public virtual ClientResult<ModerationResult> ClassifyModerationInputs(IEnumerable<ModerationInputPart> inputParts, CancellationToken cancellationToken = default);
+        [Experimental("OPENAI001")]
+        public virtual Task<ClientResult<ModerationResult>> ClassifyModerationInputsAsync(IEnumerable<ModerationInputPart> inputParts, CancellationToken cancellationToken = default);
```
joseharriaga: I think we could simply call these methods "Classify". What do you think?

api/OpenAI.net8.0.cs
```diff
@@ -3984,13 +3994,37 @@ public class ModerationClient {
...
         public virtual ClientResult<ModerationResult> ClassifyText(string input, CancellationToken cancellationToken = default);
         public virtual Task<ClientResult> ClassifyTextAsync(BinaryContent content, RequestOptions options = null);
         public virtual Task<ClientResult<ModerationResultCollection>> ClassifyTextAsync(IEnumerable<string> inputs, CancellationToken cancellationToken = default);
         public virtual Task<ClientResult<ModerationResult>> ClassifyTextAsync(string input, CancellationToken cancellationToken = default);
     }
+    [Experimental("OPENAI001")]
+    public class ModerationInputPart : IJsonModel<ModerationInputPart>, IPersistableModel<ModerationInputPart> {
+        public BinaryData ImageBytes { get; }
+        public string ImageBytesMediaType { get; }
+        public Uri ImageUri { get; }
```
joseharriaga: Back in November while you were OOF, we discussed these URLs that can either either be an HTTP url (pointing to an image) or a data URL (with the base64-encoded image data). One of the main motivations for the pattern that you have replicated here was the fact that data URLs are typically very long, and that .NET's `Uri` type had a length limit that was often broken by them. This limitation was fixed in .NET 10. Based on this, we have decided to no longer continue implementing this pattern going forward. In other words, we will represent this simply as a `Uri` instead of splitting in two: `Uri` and `BinaryData` (plus a `string` for the media type).

You can read more about this discussion in Jesse's response here:
🔗 https://github.com/openai/openai-dotnet/issues/850#issuecomment-3599407656

Since this is new code that you're adding, I recommend removing the custom code and just leaving it as a `Uri`. 🙂

codegen/generator/src/Visitors/NumericPropertiesVisitor.cs
```diff
@@ -25,11 +25,17 @@ public class NumericPropertiesVisitor : ScmLibraryVisitor
...
         if (propertyProvider is not null
             && propertyProvider.Type.Equals(typeof(long))
-            && !_excludedProperties.Contains($"{propertyProvider.EnclosingType.Type.Namespace}.{propertyProvider.EnclosingType.Name}.{propertyProvider.Name}"))
+            && !_excludedLongProperties.Contains($"{propertyProvider.EnclosingType.Type.Namespace}.{propertyProvider.EnclosingType.Name}.{propertyProvider.Name}"))
         {
             propertyProvider.Update(type: new CSharpType(typeof(int), propertyProvider.Type.IsNullable));
         }
 
+        if (propertyProvider is not null
+            && propertyProvider.Type.Equals(typeof(double)))
```
joseharriaga: nit: Could you add a `HashSet` for "_excludedDoubleProperties" (even if it's empty for now) and check for it here? I'm thinking it would be nice to have it ready for future use.

src/Custom/Moderations/ModerationInputPartKind.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace OpenAI.Moderations;
+
+// CUSTOM:
+// - Added Experimental attribute.
+// - Renamed.
+// - Plain enum type to convert from an underlying extensible enum
+[Experimental("OPENAI001")]
+public enum ModerationInputPartKind
```
joseharriaga, line 12: Should this have `[CodeGenType("InternalModerationInputPartType")]` ?

src/Custom/Moderations/ModerationInputPart.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace OpenAI.Moderations;
+
+// CUSTOM: Renamed.
+[CodeGenType("CreateModerationRequestInput")]
+public partial class ModerationInputPart
+{
+    // CUSTOM: Converted to public enum from internal extensible type.
+    [CodeGenMember("Type")]
+    internal InternalModerationInputPartType InternalType { get; set; }
```
joseharriaga: Do we need both: an internal property called `InternalType` of type `InternalModerationInputPartType` and a public property called `Kind` of type `ModerationInputPartKind`? Would it be possible to just keep the public one?

api/OpenAI.net8.0.cs
```diff
@@ -3984,13 +3994,37 @@ public class ModerationClient {
...
         public string Model { get; }
         public ClientPipeline Pipeline { get; }
+        [Experimental("OPENAI001")]
+        public virtual ClientResult<ModerationResult> ClassifyModerationInputs(IEnumerable<ModerationInputPart> inputParts, CancellationToken cancellationToken = default);
+        [Experimental("OPENAI001")]
+        public virtual Task<ClientResult<ModerationResult>> ClassifyModerationInputsAsync(IEnumerable<ModerationInputPart> inputParts, CancellationToken cancellationToken = default);
         public virtual ClientResult ClassifyText(BinaryContent content, RequestOptions options = null);
         public virtual ClientResult<ModerationResultCollection> ClassifyText(IEnumerable<string> inputs, CancellationToken cancellationToken = default);
         public virtual ClientResult<ModerationResult> ClassifyText(string input, CancellationToken cancellationToken = default);
         public virtual Task<ClientResult> ClassifyTextAsync(BinaryContent content, RequestOptions options = null);
```
joseharriaga, line 4012: The name `ClassifyText` does not really fit the protocol methods anymore since they can now do more than just classify text. I think we should add protocol methods with a name that matches the new convenience methods that take `IEnumerable<ModerationInputPart>` (e.g., we can call them simply "Classify" instead of "ClassifyText"). To do this, one suggestion is to keep the old protocol methods as custom code (that we hopefully will never touch again), and let the new methods be generated. In other words, you might be able to simply cut-and-paste the existing protocol methods from generated code over to custom code, and then change the name of the methods in client.tsp with `@clientName` from "ClassifyText" to "Classify" and finally regenerate the code.

api/OpenAI.net8.0.cs
ShivangiReja: I like the idea of shortening it! However, just `Classify` on its own feels a little vague next to the existing `ClassifyText` methods, it doesn't clearly signal what's different about it.

How about `ClassifyInputs` or `ClassifyMultimodalInput`?

`ClassifyInputs` is concise, pairs nicely with `ClassifyText`, and the parameter type (`IEnumerable<ModerationInputPart>`) already makes it clear you're passing structured multimodal content. The naming pattern would be:

`ClassifyText(string)` — simple text
`ClassifyText(IEnumerable<string>)` — multiple texts
`ClassifyInputs(IEnumerable<ModerationInputPart>)` — structured text + images

What do you think?

joseharriaga: I like `ClassifyInputs` 🙂

src/Custom/Moderations/ModerationInputPart.cs
ShivangiReja: Yes, I was thinking the same, but I followed the existing [ResponseContentPart ](https://github.com/openai/openai-dotnet/blob/f3562bf15db7e7dad0dc4e8c7f9bb9319131f9bc/src/Custom/Responses/ResponseContentPart.cs#L15-L21) pattern.

There’s no need for both, so I removed the internal type and used the public one instead.

src/Custom/Moderations/ModerationInputPartKind.cs
ShivangiReja: https://github.com/openai/openai-dotnet/pull/888#discussion_r2692617227

api/OpenAI.net8.0.cs
ShivangiReja: Added `ClassifyInputs` protocol method.

src/Custom/Moderations/ModerationClient.Protocol.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.ClientModel;
+using System.ClientModel.Primitives;
+using System.Threading.Tasks;
+
+namespace OpenAI.Moderations
+{
+    public partial class ModerationClient
+    {
+        public virtual ClientResult ClassifyText(BinaryContent content, RequestOptions options = null)
```
joseharriaga, line 11: nit: Could you add CUSTOM comments on both of these methods? Maybe something like "CUSTOM: Legacy." or similar to indicate that these only exist because we had previously GAed them.

tests/Moderations/ModerationsTests.cs
```diff
@@ -54,4 +56,80 @@ public async Task ClassifyMultipleInputs()
...
+        ModerationResult moderation = await client.ClassifyInputsAsync(inputs);
+        Assert.That(moderation, Is.Not.Null);
+        Assert.That(moderation.Flagged, Is.False);
+        Assert.That(moderation.Violence.Flagged, Is.False);
+        Assert.That(moderation.Violence.Score, Is.LessThan(0.2));
+        Assert.That(moderation.Violence.ApplicableInputKinds, Is.EqualTo(ModerationApplicableInputKinds.Image));
+    }
+
+    [RecordedTest]
+    public async Task ClassifyMultipleTexts()
```
joseharriaga: nit: What do you think about renaming the ModerationTests to indicate which method they are intended to test? I'm thinking this could be useful to differentiate scenarios such as testing the legacy `ClassifyText` with multiple strings vs. testing the new `ClassifyInputs` also with multiple strings.

For example, this method currently called "ClassifyMultipleTexts" could be renamed to "ClassifyInputsWithMultipleTexts". Similarly, the existing tests above called "ClassifySingleInput" and "ClassifyMultipleInputs" could be renamed respectively to "ClassifyTextWithSingleString" and "ClassifyTextWithMultipleStrings", or something similar.

Since this is just a little improvement to make our tests more readable, this is not a blocker and can be done separately. 🙂

PR #889, opened 2026-01-08
Factor Responses into its own assembly

christothes:


src/Shared/debug.targets
```diff
@@ -0,0 +1,5 @@
+<Project>
```
jsquire: I'm assuming these targets are for debugging purposes?  Looks like they're just message dumps.

tests/Responses/OpenAI.Responses.Tests.csproj
```diff
@@ -0,0 +1,75 @@
...
+
+  <ItemGroup>
+    <!-- Reference only the OpenAI.Responses assembly to validate it works independently -->
+    <ProjectReference Include="..\..\src\Responses\OpenAI.Responses.csproj" />
+    <ProjectReference Include="..\..\src\Shared\OpenAI.Shared.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- Include all test files from this folder -->
+    <Compile Include="ResponsesTests.cs" />
```
jsquire: All items in the folder are included automatically for a modern project, aren't they?

jsquire (commented):
I've got a PR out to move the repository configuration and package management to a centralized version, which would remove a lot of the redundancies that you needed in the project files and such.

tests/Responses/OpenAI.Responses.Tests.csproj
christothes: Because we are cherry picking other files, ` <EnableDefaultCompileItems>false</EnableDefaultCompileItems>` is set which disables the default folder compilation behavior.

christothes: Looks like we can remove it and let things happen naturally though..

src/Custom/Responses/OpenAIContext.Definition.cs
```diff
@@ -0,0 +1,7 @@
+using System.ClientModel.Primitives;
+
+namespace OpenAI;
```
KrzysztofCwalina: can we not have any types in the root namespace in responses dll?

src/Custom/TypeForwarding.cs
```diff
@@ -0,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: TypeForwardedTo(typeof(OpenAI.OpenAIClientOptions))]
```
KrzysztofCwalina: we need a responses specific client options. we dont want to type forward.

src/Shared/OpenAI.Shared.csproj
```diff
@@ -0,0 +1,100 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
KrzysztofCwalina: why do we need this new assembly if we create responses specific client options?

KrzysztofCwalina (changes requested):

src/Shared/OpenAI.Shared.csproj
christothes: there are more shared helper types than just the options.

api/OpenAI.Responses.netstandard2.0.cs
```diff
@@ -0,0 +1,1864 @@
+//------------------------------------------------------------------------------
+// <auto-generated>
+//     This code was generated by a tool.
+//
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+namespace OpenAI {
+    public class OpenAIContext : ModelReaderWriterContext {
```
KrzysztofCwalina: can this type be also renamed and in the .responses namespace?

KrzysztofCwalina (approved):

scripts/Export-Api.ps1
```diff
@@ -84,11 +85,30 @@ if ($LASTEXITCODE -ne 0) {
     exit $LASTEXITCODE
```
joseharriaga, line 85: nit: Edit the error message to something like "GenAPI failed for OpenAI with exit code $LASTEXITCODE" to differentiate it from failing for OpenAI.Responses.

scripts/Export-Api.ps1
```diff
@@ -84,11 +85,30 @@ if ($LASTEXITCODE -ne 0) {
...
+
+Write-Host ""
+Write-Host "Running GenAPI for OpenAI.Responses..." -ForegroundColor Cyan
+Write-Host ""
+
+& dotnet @responsesBuildArgs
+if ($LASTEXITCODE -ne 0) {
+    Write-Error "GenAPI failed for OpenAI.Responses with exit code $LASTEXITCODE"
+    exit $LASTEXITCODE
+}
```
joseharriaga: I wonder if this should be a small function, especially if we proceed to split out other libraries too.

examples/OpenAI.Examples.csproj
```diff
@@ -9,6 +9,7 @@
     <Compile Remove="aspnet-core/**" />
     <Content Remove="aspnet-core/**" />
     <None Remove="aspnet-core/**" />
+    <Compile Remove="Responses/**" />
```
joseharriaga: Is removing the Responses examples from compilation intentional?

scripts/Export-Api.ps1
joseharriaga, line 105: I see that the API listings for `OpenAI` still contain all the stuff from `OpenAI.Responses`, and I wonder if it would be preferable to not modify this script. I'm thinking it might be better to not produce individual API listings for Responses for a few reasons:
1. It feels redundant.
2. It sets the precedent of having to modify this script every time we split a library out of the base library.
3. It adds at least some latency to this script (which is already a little slow).

In general, if our infra can continue to work without modifications, I think that would be ideal. I'm thinking of this script, but I'm also thinking of other validation scripts (Test-ApiCompatibility.ps1, Test-AotCompatibility.ps1, etc.), as well as our pipelines.

tests/Responses/OpenAI.Responses.Tests.csproj
```diff
@@ -0,0 +1,57 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
joseharriaga, line 1: Are there benefits to splitting the `OpenAI.Tests` project too, or is it reasonable to keep it unchanged? If keeping one single project works fine, I would strongly prefer that for simplicity. Otherwise, I worry that this will be difficult to scale and impact our productivity.

tests/Responses/ResponseStoreTests.cs
```diff
@@ -390,4 +390,6 @@ public async Task GetInputItemsWithCombinedOptions()
 
         Assert.That(items, Has.Count.GreaterThan(0));
     }
+
+    private ResponsesClient GetTestClient(string overrideModel = null) => GetProxiedOpenAIClient<ResponsesClient>(overrideModel);
```
joseharriaga: nit: This helper is not used anymore, so it can be deleted. All test methods now call `GetProxiedOpenAIClient` directly.

tests/Responses/ResponsesTests.cs
```diff
@@ -1009,4 +1019,6 @@ public async Task CanCompactConversation()
             }
             """),
         strictModeEnabled: false);
+
+    private ResponsesClient GetTestClient(string overrideModel = null) => GetProxiedOpenAIClient<ResponsesClient>(overrideModel);
```
joseharriaga: nit: This helper is not used anymore, so it can be deleted. All test methods now call GetProxiedOpenAIClient directly.

README.md
```diff
@@ -101,6 +101,8 @@ The library is organized into namespaces by feature areas in the OpenAI REST API
 | `OpenAI.Responses`            | `ResponsesClient`            |
 | `OpenAI.VectorStores`         | `VectorStoreClient`          |
 
+The `OpenAI.Shared` namespace contains common types and utilities used across the library.
```
joseharriaga: I think this addition is outdated and can be removed.

Directory.Packages.props
```diff
@@ -36,8 +36,8 @@
     <PackageVersion Include="Microsoft.ClientModel.TestFramework" Version="1.0.0-alpha.20260212.1" />
   </ItemGroup>
 
-  <!-- Generator-Only Dependencies -->
-  <ItemGroup Condition="'$(IsGeneratorProject)' == 'true' or $(MSBuildProjectName.EndsWith('Plugin'))">
+  <!-- Generator Dependencies (for code customization attributes) -->
+  <ItemGroup>
```
joseharriaga: Is removing this generator-only condition intentional? I noticed that the generator project doesn't have `IsGeneratorProject` set to true, so we might have to add that instead of removing this condition?

Directory.Build.props
```diff
@@ -99,6 +99,7 @@
     <Configurations>Debug;Release</Configurations>
     <IsAotCompatible>false</IsAotCompatible>
     <IsTrimmable>false</IsTrimmable>
+    <EnableSourceLink>false</EnableSourceLink>
```
joseharriaga: Is disabling Source Link intentional?

joseharriaga (changes requested):

examples/OpenAI.Examples.csproj
christothes: I think there was an assumption here that there would be a responses example, but that probably doesn't make sense so we'll remove this.

Directory.Build.props
christothes: As I recall, this caused some type of build break, but it appears not to be a problem anymore. Will remove.

scripts/Export-Api.ps1
christothes: I think as we have new assemblies it's important to understand if there are any differences between what shows up in the main assembly and what shows up in the sub-assembly.  I think it will always be true that the main assembly will contain everything, but it's not always clear what subset would be available in the subassembly.

tests/Responses/OpenAI.Responses.Tests.csproj
christothes: This is to validate that tests still pass when loading just the subassembly and not the main assembly.

Directory.Packages.props
christothes: I'm not sure what value this provides. I tried adding it back in my local branch and it made no change to code generation. As you called out, `IsGeneratorProject` isn't defined anywhere so this hasn't done anything for a while.

tests/Responses/OpenAI.Responses.Tests.csproj
joseharriaga: I agree that this is valuable. Can we think of an alternative that allows us to get this benefit in some way that does not require maintaining separate projects and separate test infrastructure for each new library that we onboard? I'm a little worried because I'm envisioning a nearby future where we have over a dozen libraries, and the cost of standing up test infra for each, along with modifying pipelines, scripts, etc. and then maintaining all of it will hinder us. 

I'm thinking about it in the following way: Because the `OpenAI` library is by design just a passthrough for the `OpenAI.Responses` library, testing `OpenAI.Responses` _*through*_ `OpenAI` is a valid way of testing its functionality. Then, the question changes from "does this work at all?" to "can we validate that this is all actually coming from `OpenAI.Responses`?". The funny part is that because all the functionality related to the Responses API must be in `OpenAI.Responses` and we must not put any of it in `OpenAI` by design, the answer is technically an automatic "yes" by definition. So while I agree that this validation would be good to have to ensure that we don't accidentally break our design somehow, it also feels a little superfluous and makes the trade-off in complexity a bit of a hard pill to swallow. 😕

What could be considered a "good enough" way to validate that the functionality is actually coming from `OpenAI.Responses`? I'm even considering something like the following: "What if we had a job in our nightly pipeline that checks out `main`, deletes everything in the test project except Responses, swaps the dependency from `OpenAI` to `OpenAI.Responses`, and runs the recorded tests?". While a solution like this is still something that we will need to maintain, maintaining a single pipeline job is simpler compared to multiple test projects and test infras. It would be entirely isolated in a single YAML file, and because we're delegating the validation to a pipeline, we can basically forget about it during our day-to-day work. Am I crazy? 🤔

christothes: Let me see what I can do to improve it - worst case we can rely on just the main assembly

christothes: Just pushed an update that refactors things a bit to minimize duplication. Let me know if this looks ok to you.

src/Responses/Internal/ResponsesModelContext.Custom.cs
```diff
@@ -0,0 +1,223 @@
...
+    [ModelReaderWriterBuildable(typeof(StreamingResponseWebSearchCallSearchingUpdate))]
+    [ModelReaderWriterBuildable(typeof(UnknownResponseStreamEvent))]
+    [ModelReaderWriterBuildable(typeof(UriCitationMessageAnnotation))]
+    [ModelReaderWriterBuildable(typeof(WebSearchCallResponseItem))]
+    [ModelReaderWriterBuildable(typeof(WebSearchPreviewTool))]
+    [ModelReaderWriterBuildable(typeof(WebSearchTool))]
+    [ModelReaderWriterBuildable(typeof(WebSearchToolApproximateLocation))]
+    [ModelReaderWriterBuildable(typeof(WebSearchToolFilters))]
+    [ModelReaderWriterBuildable(typeof(WebSearchToolLocation))]
+    public partial class ResponsesModelContext
```
joseharriaga, line 220: This should be auto-generated.

src/Responses/Internal/ErrorTypes/OpenAIError.Generated.cs
```diff
@@ -0,0 +1,45 @@
+// <auto-generated/>
+
+#nullable disable
+
+using System;
+using System.Collections.Generic;
+
+namespace OpenAI.Responses.Internal
+{
+    internal partial class OpenAIError
```
joseharriaga, line 10: Question: The name of this file says "Generated", but it's not in the `Generated` folder. Is this file generated or is it a custom copy of a generated file?

src/Responses/Internal/ResponsesModelContext.Custom.cs
christothes: Are you suggesting we do this via a Visitor or something like that? We can certainly do that longer term.

src/Responses/Internal/ErrorTypes/OpenAIError.Generated.cs
christothes: custom copy.

src/Custom/OpenAIClientOptions.cs
```diff
@@ -104,4 +105,20 @@ internal OpenAIClientOptions(IConfigurationSection section)
             UserAgentApplicationId = userAgentApplicationId;
         }
     }
+
+    /// <summary>
+    /// Creates a new <see cref="ResponsesClientOptions"/> instance initialized with the same property values
+    /// as the specified <see cref="OpenAIClientOptions"/>.
+    /// </summary>
+    /// <returns> A new <see cref="ResponsesClientOptions"/> with the same property values. </returns>
+    internal ResponsesClientOptions ToResponsesClientOptionss()
```
ShivangiReja, line 114:
```suggestion
    internal ResponsesClientOptions ToResponsesClientOptions()
```

src/Custom/Responses/OpenAIClientExtensions.cs
ShivangiReja, line 1: Should [these](https://github.com/christothes/openai-dotnet/blob/e38677ab2a94b5873f0589d4a6bb373abfde4539/src/Custom/OpenAIHostBuilderExtensions.cs#L409-L431) methods move to the Responses assembly as extension methods?

PR #890, opened 2026-01-10
[EngSys] Central package management and properties

jsquire:
# Summary

The focus of these changes is to prepare the repository to support multiple packages by centralizing package versions and common build properties.  In addition, the solution file has also been converted to the
.NET 10 format (.slnx) which significantly improves readability and allows for easier manual updates.

### DIFF
--- a/codegen/Directory.Build.props
+++ b/codegen/Directory.Build.props
@@ -6,7 +6,7 @@
 
   <!-- Custom Properties -->
   <PropertyGroup>
-    <RepositoryRoot>$(MSBuildThisFileDirectory)</RepositoryRoot>
+    <RepositoryRoot>$([MSBuild]::NormalizeDirectory('$(MSBuildThisFileDirectory)..'))</RepositoryRoot>
     <GeneratorTargetFrameworks>net10.0</GeneratorTargetFrameworks>
   </PropertyGroup>
 


m-nash (approved):

Directory.Build.props
```diff
@@ -0,0 +1,114 @@
...
+  <PropertyGroup>
+    <VersionPrefix>2.8.0</VersionPrefix>
+    <VersionSuffix></VersionSuffix>
+    <InformationalVersion>$(Version)</InformationalVersion>
+  </PropertyGroup>
+
+  <!-- Publishing -->
+  <PropertyGroup>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageIcon>OpenAI.png</PackageIcon>
```
joseharriaga, line 58: I think this image was moved.

Directory.Build.props
```diff
@@ -0,0 +1,114 @@
...
+    <NoWarn>$(NoWarn);0618</NoWarn>
+
+    <!-- Disable unused fields warnings -->
+    <NoWarn>$(NoWarn);0169</NoWarn>
+
+    <!-- Disable warnings for experimental APIs -->
+    <NoWarn>$(NoWarn);OPENAI001;OPENAI002;OPENAICUA001;SCME0001</NoWarn>
+
+    <!-- Enable AOT for compatible target frameworks -->
+    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</IsAotCompatible>
```
joseharriaga, line 45: Is `IsAotCompatible` enough, or should we also keep `IsTrimmable`?

Directory.Build.props
```diff
@@ -0,0 +1,114 @@
...
+    <VersionSuffix></VersionSuffix>
+    <InformationalVersion>$(Version)</InformationalVersion>
+  </PropertyGroup>
+
+  <!-- Publishing -->
+  <PropertyGroup>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageIcon>OpenAI.png</PackageIcon>
+    <PackageProjectUrl>$(RepositoryUrl)</PackageProjectUrl>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
```
joseharriaga, line 60: nit: Should this be `$(RepositoryRoot)README.md` ?

Directory.Build.props
jsquire: `IsTrimmable` is set implicitly by `IsAotCompatible`, so both are set.

joseharriaga: Got it! One more question: I just noticed that the condition for `IsTrimmable` is against net6.0, while the condition for `IsAotCompatible` is against net7.0. Is that difference worth keeping?

jsquire: It made sense when `net6.0` was a supported framework because you could trim on .NET 6.   Since anything < `net8.0` is longer a supported runtime, there's no benefit to allowing trimming there.

Directory.Build.props
jsquire: Yes, in the source tree - but NOT in the package.   This is a relative path in the package, which gets configured in via the `OpenAI.csproj` metadata:

```xml
<ItemGroup>
  <None Include="$(RepositoryRoot)eng\OpenAI.png" Pack="true" PackagePath="\" />
  <None Include="$(RepositoryRoot)CHANGELOG.md" Pack="true" PackagePath="\" />
  <None Include="$(RepositoryRoot)README.md" Pack="true" PackagePath="\" />
 </ItemGroup>
  ```

Since those copy the files into the root of the package, that's where these publishing props have to point.

Directory.Build.props
jsquire: Same as the image.  😄

PR #901, opened 2026-01-20
Add implicit conversions to `McpToolCallApprovalPolicy` union

joseharriaga:
- Added implicit conversion from `CustomMcpToolCallApprovalPolicy` to `McpToolCallApprovalPolicy` 
- Added implicit conversion from `GlobalMcpToolCallApprovalPolicy ` to `McpToolCallApprovalPolicy` 
- Added asserts in the constructors of `McpToolCallApprovalPolicy` to verify that the received argument is not null

### DIFF
--- a/src/Custom/Responses/Tools/McpTool/McpToolCallApprovalPolicy.cs
+++ b/src/Custom/Responses/Tools/McpTool/McpToolCallApprovalPolicy.cs
@@ -12,11 +12,8 @@ public partial class McpToolCallApprovalPolicy
     // CUSTOM: Added to support the corresponding component of the union.
     public McpToolCallApprovalPolicy(GlobalMcpToolCallApprovalPolicy globalPolicy)
     {
-        Argument.AssertNotNull(globalPolicy, nameof(globalPolicy));
-
         GlobalPolicy = globalPolicy;
     }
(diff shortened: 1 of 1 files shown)


examples/Responses/Example05_RemoteMcp.cs
```diff
@@ -25,7 +25,7 @@ public void Example05_RemoteMcp()
                 new McpTool(serverLabel: "dmcp", serverUri: new Uri("https://dmcp-server.deno.dev/sse"))
                 {
                     ServerDescription = "A Dungeons and Dragons MCP server to assist with dice rolling.",
-                    ToolCallApprovalPolicy = new McpToolCallApprovalPolicy(GlobalMcpToolCallApprovalPolicy.NeverRequireApproval)
+                    ToolCallApprovalPolicy = GlobalMcpToolCallApprovalPolicy.NeverRequireApproval
```
christothes: Since this is an example, would it be of any value to comment about the implicit conversion? Perhaps it's obvious though.

joseharriaga: I think this is a good idea to add context since implicit conversions are not super obvious.

PR #903, opened 2026-01-21
Fix MPFD tests - normalize filepaths and improve perf in playback

m-redding:


tests/Utility/OpenAIRecordedTestBase.cs
```diff
@@ -10,11 +11,20 @@ public class OpenAIRecordedTestBase : RecordedTestBase<OpenAITestEnvironment>
     {
         public OpenAIRecordedTestBase(bool isAsync, RecordedTestMode? mode = null) : base(isAsync, mode)
         {
+            // Normalizes filepaths used as filenames in Content-Disposition headers, so record/playback works across different OSes
+            CustomSanitizers.Add(new ContentDispositionFilePathSanitizer());
+
+            // Turn off default sanitizers to improve performance in large recordings (audio/image specifically)
+            // This turns off all sanitizers exception Authorization headers, so we have to be very careful to explicitly sanitize
+            // all other sensitive information
+            UseDefaultSanitizers = false;
```
christothes, line 20: Which headers are we explicitly missing with this change?

jsquire: @christothes made me curious, so I went looking:
https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs#L40

joseharriaga (approved):

PR #904, opened 2026-01-21
Clone CreateResponseOptions

christothes:


src/Custom/Responses/CreateResponseOptions.cs
```diff
@@ -122,10 +122,46 @@ public CreateResponseOptions(IEnumerable<ResponseItem> inputItems, string model
...
+        clone.ServiceTier = ServiceTier;
+        clone.PreviousResponseId = PreviousResponseId;
+        clone.Model = Model;
+        clone.ReasoningOptions = ReasoningOptions;
+        clone.BackgroundModeEnabled = BackgroundModeEnabled;
+        clone.MaxOutputTokenCount = MaxOutputTokenCount;
+        clone.MaxToolCallCount = MaxToolCallCount;
+        clone.Instructions = Instructions;
+        clone.TextOptions = TextOptions;
+        foreach (var tool in Tools)
```
jsquire, line 148: Since you're walking the lists inline, we may want to consider using `CollectionsMarshal.AsSpan` in the `foreach` to elide the allocation for the enumerators.

src/Custom/Responses/ResponsesClient.cs
```diff
@@ -633,7 +633,7 @@ internal virtual CreateResponseOptions CreatePerCallOptions(CreateResponseOption
     {
         CreateResponseOptions clonedOptions = userProvidedOptions is null
```
jsquire, line 634: Should we consider doing a thread-static shared instance for default options rather than creating an instance per-call?

src/Custom/Responses/CreateResponseOptions.cs
christothes: I don't think there are typically enough items here to move the needle, but it's a good suggestion!

src/Custom/Responses/ResponsesClient.cs
christothes: What scenario did you have in mind where we'd use the default options? My assumption is that we'll typically only clone when someone is passing in a non-default options instance.

jsquire: We're doing a `new()` if there's no custom options passed.   My suggestion is to replace that with a thread-static version.    We need to be able to safely mutate for this call but there's no reason why we need a fresh instance each time, is there?

jsquire (approved):

src/Custom/Responses/ResponsesClient.cs
christothes: Ah, makes sense. For some reason I thought this comment was on the clone method.

src/Custom/Responses/CreateResponseOptions.cs
```diff
@@ -8,6 +8,8 @@ namespace OpenAI.Responses;
 [CodeGenType("CreateResponse")]
 public partial class CreateResponseOptions
 {
+    internal static readonly CreateResponseOptions Default = new CreateResponseOptions();
```
jsquire, line 11: Will this work without being `[ThreadStatic]`?   Won't we end up with concurrency issues because of the mutation during operations?

PR #907, opened 2026-01-22
Bump to version 1.0.0-alpha.20260116.3 - Add internal partial structs for new enums

MaiLinhP:
Made enums internal to fix build issues

### DIFF
--- a/src/Custom/Conversations/IncludedConversationItemProperty.cs
+++ b/src/Custom/Conversations/IncludedConversationItemProperty.cs
@@ -9,7 +9,6 @@ namespace OpenAI.Conversations
     // - Renamed.
     // - Implemented in custom code due to a bug in the code generator that skips generation for this type when only
     //   protocol methods are generated.
-    [Experimental("OPENAI001")]
     [CodeGenType("IncludeEnum")]
     public readonly partial struct IncludedConversationItemProperty : IEquatable<IncludedConversationItemProperty>
     {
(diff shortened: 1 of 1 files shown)


src/Custom/Videos/Internal/InternalOrderEnum.cs
```diff
@@ -0,0 +1,11 @@
+using Microsoft.TypeSpec.Generator.Customizations;
+
+namespace OpenAI.Videos;
+
+// CUSTOM:
+// - Renamed.
+// - Made internal until we support the convenience methods.
+[CodeGenType("OrderEnum")]
+internal readonly partial struct InternalOrderEnum
```
jorgerangel-msft, line 9: This is technically a correct way of addressing the changes here, but I took a look as to why this was happening and it is a bug fix from the upstream version [bump](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/CHANGELOG.md#bug-fixes-1) of the TCGC dependency.

Since the usage for these enum's is now  `Input`, we generate these types. Before, tcgc gave us a usage of `None` so we didn't generate them, or we generated them and they were removed as part of the roslyn post processing.

I think the cleaner fix here would be to manually add the "None" usage for the effected enums using the [usage](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/decorators/#@Azure.ClientGenerator.Core.usage) decorator in their respective client.tsp spec file. That would avoid introducing custom code.

jorgerangel-msft: Looks like tcgc doesn't support `usage.none` 😞, so you can ignore this suggestion.

jorgerangel-msft, line 9: The only other tip / suggestion I have is to use the [access](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/decorators/#@Azure.ClientGenerator.Core.access) decorator to set this type to internal. Though I don't see us using that anywhere else, so feel free to ignore and see what Jose thinks!

PR #909, opened 2026-01-23
[EngSys] Replace GenAPI tool with GenAPI MSBuild task in Export-Api.ps1

joseharriaga:
* Replaced the GenAPI tool with the GenAPI MSBuild task, which is able to automatically pull all assembly dependencies at build time. This means that we no longer need to manually maintain a list of dependencies just so that we can pass them to the GenAPI tool.
* Added the public .NET dev feed to our nuget.config from where the GenAPI MSBuild task can be restored. This means that we no longer need to manually build and install a copy of the GenAPI tool locally.

### DIFF
--- a/scripts/Export-Api.ps1
+++ b/scripts/Export-Api.ps1
@@ -98,8 +98,9 @@ Get-ChildItem -Path $outputDirectory -Filter "OpenAI.*.cs" | ForEach-Object {
     $content = $content -creplace '\r?\n\r?\n', "`n"
     $content = $content -creplace '\r?\n *{', " {"
 
-    # Remove fully-qualified System namespace prefixes.
+    # Remove fully-qualified namespace prefixes.
     @(
+        "Diagnostics\.CodeAnalysis\.",
         "System\.ComponentModel\.",
         "System\.ClientModel\.Primitives\.",
         "System\.ClientModel\.",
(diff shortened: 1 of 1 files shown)


joseharriaga:
@copilot Modify the Export-Api.ps1 script in this PR so that it deletes the contents of the "api" folder before running the GenAPI task.

joseharriaga:
@copilot Simplify the changes in this PR by taking the following into consideration:

1. The Export-Api.ps1 script does not need the `Configuration` parameter. It should always be "Release".
2. The Export-Api.ps1 script does not need the `TargetFramework` parameter. It should always use all the TFMs specified in the .csproj file.

joseharriaga:
@copilot The current Export-Api.ps1 script in this PR runs `dotnet build` once for each target framework. Is it necessary to run them one by one? Or is it possible to run them all at once? Would the GenAPI target support that?

scripts/Export-Api.ps1
```diff
@@ -1,279 +1,140 @@
...
+        "System\.ClientModel\.Primitives\.",
+        "System\.ClientModel\.",
+        "System\.Collections\.Generic\.",
+        "System\.Collections\.",
+        "System\.Threading\.Tasks\.",
+        "System\.Threading\.",
+        "System\.Text\.Json\.",
+        "System\.Text\.",
+        "System\.IO\.",
+        "System\."
```
jsquire: Wouldn't this last entry satisfy the above conditions as well?   Or, does this not act as a wildcard?

Directory.Build.targets
jsquire, line 1: Nice!

scripts/Export-Api.ps1
joseharriaga: Correct. It's not a wildcard, but it does have to go at the end to avoid conflicting with the other entries above it.

PR #913, opened 2026-01-23
[Spec Ingestion] Audio

ShivangiReja:
Ingested latest TypeSpec for Audio

### DIFF
--- a/src/Custom/Audio/AudioTranscription.Serialization.cs
+++ b/src/Custom/Audio/AudioTranscription.Serialization.cs
@@ -30,6 +30,10 @@ public partial class AudioTranscription
         return DeserializeAudioTranscription(document.RootElement, null);
     }
 
+    // CUSTOM: The explicit operator is not auto-generated because the response type
+    // has been customized in the specification (DotNetCombinedJsonTranscriptionResponse).
+    // Auto-generation only occurs for root output types that directly match the operation's
+    // response schema without customization.
     [Experimental("OPENAI001")]
     public static explicit operator AudioTranscription(ClientResult result)
     {


src/Custom/Audio/Internal/GeneratorStubs.cs
```diff
@@ -2,18 +2,39 @@
...
+[CodeGenType("TranscriptTextDeltaEvent")] internal partial class InternalTranscriptTextDeltaEvent { }
+[CodeGenType("TranscriptTextDeltaEventLogprobs")] internal partial class InternalTranscriptTextDeltaEventLogprobs { }
+[CodeGenType("TranscriptTextDoneEvent")] internal partial class InternalTranscriptTextDoneEvent { }
+[CodeGenType("TranscriptTextDoneEventLogprobs")] internal partial class InternalTranscriptTextDoneEventLogprobs { }
+[CodeGenType("TranscriptTextSegmentEvent")] internal partial class InternalTranscriptTextSegmentEvent { }
+
+// Remove these after streaming events are implemented
+[CodeGenType("DotNetCreateSpeechStreamingResponseType")] internal readonly partial struct InternalDotNetCreateSpeechStreamingResponseType { }
+[CodeGenType("SpeechAudioDoneEventUsage")] internal partial class InternalSpeechAudioDoneEventUsage { }
+[CodeGenType("DotNetCreateSpeechStreamingResponse")] internal partial class InternalDotNetCreateSpeechStreamingResponse { }
```
ShivangiReja, line 37: We will add support for speech streaming events in a separate PR: https://github.com/openai/openai-dotnet/issues/914

src/Custom/Audio/Internal/GeneratorStubs.cs
```diff
@@ -2,18 +2,39 @@
...
+[CodeGenType("CreateSpeechRequestStreamFormat")] internal readonly partial struct InternalCreateSpeechRequestStreamFormat { }
+[CodeGenType("CreateTranscriptionResponseJsonUsageType")] internal readonly partial struct InternalCreateTranscriptionResponseJsonUsageType{ }
+[CodeGenType("CreateTranscriptionResponseJsonUsage")] internal partial class InternalCreateTranscriptionResponseJsonUsage { }
+[CodeGenType("UnknownCreateTranscriptionResponseJsonUsage")] internal partial class InternalUnknownCreateTranscriptionResponseJsonUsage { }
+[CodeGenType("TranscriptTextUsageTokens")] internal partial class InternalTranscriptTextUsageTokens { }
+[CodeGenType("TranscriptTextUsageTokensInputTokenDetails")] internal partial class InternalTranscriptTextUsageTokensInputTokenDetails { }
+[CodeGenType("TranscriptTextUsageDuration")] internal partial class InternalTranscriptTextUsageDuration { }
+[CodeGenType("CreateTranscriptionResponseDiarizedJson")] internal partial class InternalCreateTranscriptionResponseDiarizedJson { }
+[CodeGenType("TranscriptionDiarizedSegment")] internal partial class InternalTranscriptionDiarizedSegment { }
```
ShivangiReja, line 25: We will add support for diarized transcription response in a separate PR: https://github.com/openai/openai-dotnet/issues/916

api/OpenAI.net8.0.cs
```diff
@@ -1347,6 +1348,18 @@ public class StreamingAudioTranscriptionTextDoneUpdate : StreamingAudioTranscrip
         protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
     }
     [Experimental("OPENAI001")]
+    public class StreamingAudioTranscriptionTextSegmentUpdate : StreamingAudioTranscriptionUpdate, IJsonModel<StreamingAudioTranscriptionTextSegmentUpdate>, IPersistableModel<StreamingAudioTranscriptionTextSegmentUpdate> {
+        public TimeSpan End { get; }
+        public string Id { get; }
+        public string Speaker { get; }
+        public TimeSpan Start { get; }
+        public string Text { get; }
```
joseharriaga, line 1356: For consistency with other names in the library, what do you think about the following renames?

* End     --> EndTime
* Start   --> StartTime

And then, for clarity, what do you think about these ones too?

* Id      --> SegmentId
* Speaker --> SpeakerLabel

specification/base/typespec/audio/models.tsp
```diff
@@ -223,174 +398,406 @@ model CreateTranscriptionResponseVerboseJson {
...
 
-  @doc("""
-    The log probabilities of the individual tokens in the transcription. Only included if you [create a transcription](/docs/api-reference/audio/create-transcription) with the `include[]` parameter set to `logprobs`.
-    """)
-  logprobs?: TranscriptTextDoneEventLogprob[];
+  /** The log probabilities of the individual tokens in the transcription. Only included if you [create a transcription](https://platform.openai.com/docs/api-reference/audio/create-transcription) with the `include[]` parameter set to `logprobs`. */
+  logprobs?: TranscriptTextDoneEventLogprobs[];
+
+  #suppress "@azure-tools/typespec-azure-core/documentation-required" "Auto-suppressed warnings non-applicable rules during import."
+  usage?: TranscriptTextUsageTokens;
```
joseharriaga, line 673: I was going to recommend exposing this property as part of this PR, but this property is interesting: I was not expecting the type of this property to be `TranscriptTextUsageTokens`. Instead I was expecting it to be the base model `CreateTranscriptionResponseJsonUsage`, which is what we see in the non-streaming transcription result.

Context: In transcriptions, usage can be measure in two different ways depending on the model that you use. Some models measure usage in audio input duration (`TranscriptTextUsageDuration`), while other models measure usage in tokens (`TranscriptTextUsageTokens`).

Perhaps this means that models that measure usage in audio input duration do not support streaming? Or could it be that the spec is wrong here? I recommend investigating to confirm.

specification/base/typespec/audio/operations.tsp
```diff
@@ -1,64 +1,1541 @@
...
-import "../common";
 import "./models.tsp";
-
+import "@typespec/sse";
 using TypeSpec.Http;
 using TypeSpec.OpenAPI;
-
+using TypeSpec.SSE;
 namespace OpenAI;
+/** Generates audio from the input text. */
```
joseharriaga, line 9: Just a heads-up: There appears to be a new Voices API.
🔗 https://platform.openai.com/docs/api-reference/audio/createVoice

It appears that it is not yet in the TypeSpec, but let's keep an eye on it so that we can add protocol methods for it as soon as it is. 🙂

api/OpenAI.net8.0.cs
```diff
@@ -1331,6 +1331,7 @@ public class SpeechGenerationOptions : IJsonModel<SpeechGenerationOptions>, IPer
     [Experimental("OPENAI001")]
     public class StreamingAudioTranscriptionTextDeltaUpdate : StreamingAudioTranscriptionUpdate, IJsonModel<StreamingAudioTranscriptionTextDeltaUpdate>, IPersistableModel<StreamingAudioTranscriptionTextDeltaUpdate> {
         public string Delta { get; }
+        public string SegmentId { get; }
```
joseharriaga, line 1334: I just updated the Export-Api.ps1 script and there is a new "OpenAI.net10.0.cs" API listing. I recommend syncing to the latest. 🙂

src/Custom/Audio/AudioTranslation.cs
```diff
@@ -6,9 +6,6 @@ namespace OpenAI.Audio;
 [CodeGenType("CreateTranslationResponseVerboseJson")]
 public partial class AudioTranslation
 {
-    // CUSTOM: Made private. This property does not add value in the context of a strongly-typed class.
-    private string Task { get; } = "translate";
```
joseharriaga, line 10: It's interesting that this property just vanished. I guess it's not super important since it was kind of useless and it wasn't part of the public API.

I see you're removing it from `AudioTranslation` here, but you might want to also remove it from `AudioTranscription` for consistency.

src/Custom/Audio/AudioTranscription.Serialization.cs
```diff
@@ -16,12 +16,12 @@ internal static AudioTranscription FromResponse(PipelineResponse response)
             contentType.StartsWith("text/plain", StringComparison.Ordinal))
         {
             return new AudioTranscription(
-                task: default,
                 language: null,
                 duration: null,
                 text: response.Content?.ToString(),
                 words: new ChangeTrackingList<TranscribedWord>(),
                 segments: new ChangeTrackingList<TranscribedSegment>(),
+                usage: null,
```
joseharriaga, line 24: Unrelated, but I just noticed that there is a custom explicit operation from ClientResult to AudioTranscription defined in this file. These days I believe it should be possible to remove it and let it be generated?

src/Custom/Audio/AudioTranslation.Serialization.cs
```diff
@@ -12,7 +12,7 @@ internal static AudioTranslation FromResponse(PipelineResponse response)
         if (response?.Headers?.TryGetValue("Content-Type", out string contentType) == true &&
             contentType.StartsWith("text/plain", StringComparison.Ordinal))
         {
-            return new AudioTranslation(language: null, text: response.Content?.ToString(), segments: null, task: default, duration: null, additionalBinaryDataProperties: null);
+            return new AudioTranslation(language: null, text: response.Content?.ToString(), segments: null, duration: null, additionalBinaryDataProperties: null);
```
joseharriaga: I feel a little bad for asking something so nit-picky, but could you put the arguments in separate lines like in AudioTranscription.Serialization.cs so that it's easier to read and compare with one another? 😁😅

src/Custom/Audio/AudioTranslation.cs
ShivangiReja: Nice catch!

specification/base/typespec/audio/models.tsp
ShivangiReja: After testing, here's what I found:

### Streaming Support by Model:

| Model | Streaming Support | Streaming `usage` type |
|-------|-------------------|------------------------|
| `whisper-1` | ❌ Not supported | N/A |
| `gpt-4o-transcribe` | ✅ | `tokens` |
| `gpt-4o-mini-transcribe` | ✅ | `tokens` |
| `gpt-4o-transcribe-diarize` | ✅ | `tokens` |

### Key Finding:
All streaming-capable models return `usage.type: "tokens"` in the `transcript.text.done` event. This was confirmed through live API testing.

### Conclusion:
**The spec is correct.** `TranscriptTextDoneEvent.usage` being typed as `TranscriptTextUsageTokens` is accurate.

src/Custom/Audio/AudioTranscription.Serialization.cs
ShivangiReja: I tried removing the explicit operator, but it doesn’t seem to be auto-generated.

The reason it isn’t being generated is that we’ve customized the response type here: [https://github.com/openai/openai-dotnet/blob/main/specification/client/models/audio.models.tsp](https://github.com/openai/openai-dotnet/blob/main/specification/client/models/audio.models.tsp), and it isn’t directly returned by any operation in the spec.

specification/base/typespec/audio/models.tsp
ShivangiReja: I was thinking of exposing it as part of this PR, but that would require making multiple additional types public, which we’ll likely need anyway for [https://github.com/openai/openai-dotnet/issues/916](https://github.com/openai/openai-dotnet/issues/916). So I’m planning to handle this as part of #916 instead, to keep this PR focused.

joseharriaga (approved):

Issue #602, opened 2025-08-14
[Docs] Convert samples to snippets


jsquire:
# Summary

Currently, the code snippets used in samples and the README are local to the document and exist only in raw markdown form.  This makes them brittle as the API of the library evolves, as they cannot be automatically validated during compilation and/or test runs.   These should all be converted to snippet form and injected via the `snippet-generator` tool to ensure they do not get stale.

## Scope

- Update stand alone samples form the `examples` directory in source.
- Update the README

## Out of scope

- The examples in the `docs` area are already runnable units and utilize the new C# single-file runnable format.   These don't make sense to convert to snippets.

## Success criteria

- The code snippets in the README are injected from test compilablecode
- The code snippets in the `examples` section are injected from compilable sample code

jsquire:
Fixed by #923

PR #923, opened 2026-01-31
[Docs] README snippetization and content updates

jsquire:
# Summary

The focus of these changes is to extract the README examples into code snippets and update the content to reflect the new Responses API.   Also riding along are updates to dependency package versions.


jsquire:
closes #602

jsquire:
Test failure is also consistent locally and is the result of some test utility issues.  After discussion, I'm going to merge this with the failure and address the test issues separately.

PR #928, opened 2026-02-03
Realtime GA

christothes:


specification/base/typespec/realtime/custom/content_parts.tsp
```diff
@@ -4,33 +4,41 @@ namespace OpenAI;
...
+  /** Input text content from the user or system. */
   input_text: "input_text",
+  /** Input audio content from the user. */
   input_audio: "input_audio",
-  text: "text",
-  audio: "audio",
+  /** Output text content from the assistant (GA format). */
+  output_text: "output_text",
+  /** Output audio content from the assistant (GA format). */
+  output_audio: "output_audio",
```
ShivangiReja, line 14: I believe we should use an exact copy of the latest base TypeSpec. In the [latest spec](https://github.com/microsoft/openai-openapi-pr/tree/main/packages/openai-typespec/src/realtime), there’s no `custom/` folder or `custom.tsp`, so can we remove these extra files here to match the base spec exactly?

If anything breaks, we should first try to fix it via [`client.tsp`](https://github.com/christothes/openai-dotnet/tree/edad8eec926df70a29ea8840f3414939c4180f81/specification/client) or [`model.tsp`](https://github.com/christothes/openai-dotnet/tree/edad8eec926df70a29ea8840f3414939c4180f81/specification/client/models).

If it can’t be fixed through those overlays, then we should open an issue to fix the latest base spec and address it upstream.

joseharriaga (approved):

docs/realtime-discriminator-issue.md
```diff
@@ -0,0 +1,111 @@
+# Realtime API TypeSpec uses unions instead of discriminators for polymorphic model types
```
joseharriaga: The `./docs` folder is intended for user docs. I recommend moving this somewhere else.

api/OpenAI.net10.0.cs
```diff
@@ -4707,10 +4716,10 @@ public event EventHandler<BinaryData> OnSendingCommand {
             add;
             remove;
         }
-        public virtual ClientResult CreateEphemeralToken(BinaryContent content, RequestOptions options = null);
-        public virtual Task<ClientResult> CreateEphemeralTokenAsync(BinaryContent content, RequestOptions options = null);
-        public virtual ClientResult CreateEphemeralTranscriptionToken(BinaryContent content, RequestOptions options = null);
-        public virtual Task<ClientResult> CreateEphemeralTranscriptionTokenAsync(BinaryContent content, RequestOptions options = null);
+        public virtual ClientResult<RealtimeCreateClientSecretResponse> CreateRealtimeClientSecret(RealtimeCreateClientSecretRequest body, CancellationToken cancellationToken = default);
```
KrzysztofCwalina, line 4719: Can we remove RealtimeCLient ctor that takes OpenAIClientOptions? We talked about this at one of the arch meetings. Realtime client might need different options and also we want to be able to not GA new ties to the shared type

christothes: OpenAIClientOptions are pretty simple and are primarily used to create the pipeline. Do we want to just create a copy of it that has the same properties or do we want to do something different?

KrzysztofCwalina: I would do the same we are doing in ResponsesClinet, a copy?

PR #934, opened 2026-02-05
Represent image URIs in Responses as `Uri` instead of `string`

joseharriaga:
* Changed the type of image URIs properties from `string` to `Uri`
* Removed the custom code used to generate data URI strings from `BinaryData` as per our design discussion.
* Fixed a few failing tests related to Responses.

### DIFF
--- a/tests/Responses/ResponsesTests.cs
+++ b/tests/Responses/ResponsesTests.cs
@@ -444,6 +444,7 @@ public partial class ResponsesTests : OpenAIRecordedTestBase
         Assert.That(response, Is.Not.Null);
     }
 
+#if NET10_0_OR_GREATER
     [RecordedTest]
     public async Task ImageInputWorks()
     {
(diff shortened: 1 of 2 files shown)


tests/Responses/ResponsesTests.cs
```diff
@@ -988,18 +791,19 @@ ResponseToolChoice toolChoice
         Assert.That(functionCall.FunctionName, Is.EqualTo(toolChoice.FunctionName));
     }
 
-    [Ignore("Failing")]
+    // [Ignore("Failing")]
```
christothes: Is this intentional?

joseharriaga: Yes, but I meant to remove the attribute entirely, not just comment it out. Fixed!

jsquire (approved):

tests/Responses/ResponsesTests.cs
```diff
@@ -444,6 +444,7 @@ public async Task MessageHistoryWorks()
         Assert.That(response, Is.Not.Null);
     }
 
+#if NET10_0_OR_GREATER
```
jsquire, line 447: Why?

joseharriaga: Because otherwise it hits the length limit of `Uri` in frameworks <10.0.

PR #935, opened 2026-02-05
[Spec Ingestion] VectorStore

ShivangiReja:
Ingested latest TypeSpec for VectorStore


codegen/generator/src/Visitors/NumericTypesVisitor.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Collections.Generic;
+
+namespace OpenAILibraryPlugin.Visitors;
+
+/// <summary>
+/// We prefer to use 32-bit numeric types (`int`, `float`) for numeric properties and parameters unless there is a specific reason to
+/// use 64-bit numeric types (`long`, `double`). Because using 64-bit types is therefore the exception, this visitor
+/// converts all `long` properties/parameters to `int` and all `double` properties to `float` by default unless they are explicitly excluded.
+/// </summary>
+public class NumericTypesVisitor : ScmLibraryVisitor
```
christothes, line 15: If this is mostly a rename, could you do this with `git mv` to preserve the history and make it easier to see the diff?

ShivangiReja: It wasn't just a rename, I also added 2 new override methods (`VisitField`, `VisitMethod`) and 2 new exclusion sets. I tried git mv but Git doesn't detect it as a rename because the content similarity is too low (the file nearly doubled from 46 to 87 lines).

Issue #912, opened 2026-01-23
[BUG] RealtimeItem.CreateFunctionCallOutput creates null MessageContentParts which can't be null even with valid output passed to function


JohnGalt1717:
### Describe the bug

In the past RealTimeItem had a nullable MessageContentParts. It has now been made not nullable.

The .Output is still not public.

And this:

```
            var output = await aiFunction.InvokeAsync(new(jsonArgs), cancellationToken);
            return RealtimeItem.CreateFunctionCallOutput(
                update.FunctionCallId,
                output?.ToString() ?? ""
            );

```

ALWAYS results in MessageContentParts being null despite being not nullable.

It should be putting the value in or at least have a single entry that is "" in the worst case. It doesn't.

### Steps to reproduce

```
            var output = await aiFunction.InvokeAsync(new(jsonArgs), cancellationToken);
            return RealtimeItem.CreateFunctionCallOutput(
                update.FunctionCallId,
                output?.ToString() ?? ""
            );

```

Evaluate the RetimeItem that is returned.

### Code snippets

```C#

```

### OS

macOs 26

### .NET version

10.0.101

### Library version

OpenAI 2.8.0

jsquire:
Hi @JohnGalt1717.   Thanks for reaching out and we regret that you're experiencing difficulties.   According to the [docs](https://platform.openai.com/docs/api-reference/realtime-client-events/conversation/item/create#realtime_client_events-conversation-item-create-item-realtime_function_call_output_item) and the [REST API Spec](https://github.com/openai/openai-openapi), the Realtime FunctionCallOutput item does not have a `MessageContentParts`, let alone require one.    Can you help me understand your end-to-end scenario?

_(REST API spec, Realtime models, line 46,679):_
```yaml
RealtimeConversationItemFunctionCallOutput:
      type: object
      title: Realtime function call output item
      description: A function call output item in a Realtime conversation.
      properties:
        id:
          type: string
          description: The unique ID of the item. This may be provided by the client or generated by the server.
        object:
          type: string
          enum:
            - realtime.item
          description: >-
            Identifier for the API object being returned - always `realtime.item`. Optional when creating a
            new item.
          x-stainless-const: true
        type:
          type: string
          enum:
            - function_call_output
          description: The type of the item. Always `function_call_output`.
          x-stainless-const: true
        status:
          type: string
          enum:
            - completed
            - incomplete
            - in_progress
          description: The status of the item. Has no effect on the conversation.
        call_id:
[...]

PR #937, opened 2026-02-08
Fix: Return empty list instead of null for MessageContentParts in Fun…

swamy18:
…ctionCallOutput items (#912)

The MessageContentParts property was returning null for FunctionCallOutput items, even though the property is declared as non-nullable. This fix adds a null-coalescing fallback that returns an empty read-only list instead, ensuring the property never returns null.


jsquire (changes requested):
@swamy18:  Thank you for your contribution and interest in improving the OpenAI developer experience.   

Unfortunately, this is not a change that we'll accept at the moment.   As discussed in the issue that you've linked the current return of `null` is intentional and the value that we believe is correct for this scenario.

The issue author has a misunderstanding of how nullable annotations work, which you've reflected in your PR description here.    An application opting into nullability does not extend to its dependencies unless those dependencies have also opted in and are annotated - which the OpenAI client library intentionally does not.   As a result, the contract for this API makes no guarantees about null returns, as detailed in the linked issue.

PR #950, opened 2026-02-13
[Test Infrastructure] Unify test environment

jsquire:
# Summary

Removes legacy test utilities in favor of unifying environment operations on the test framework's  `TestEnvironment` abstraction.  Overhauls the test environment setup to support `.env` files for local configuration, upgrades the test framework dependency, and re-records Chat test sessions.

### Test environment configuration (OpenAITestEnvironment)

   - Adds layered environment variable resolution with clear precedence: hard-coded defaults → repo-root .env → local .env → process environment variables. This enables easier environment updates.
   - Introduces a basic `.env `file parser with line-by-line parsing, comment support, and clear error reporting for malformed files.
   - Allows for predictable and consistent environment overrides for the test framework defaults, which previously were only available via true environment variables being set.
   - Moves global state initialization for the test framework's `TestEnvironment` base to the instance constructor with concurrency guards to prevent the interdependencies in static construction.
   - Integrates test environment reads with the recording framework; previously it was inconsistent and callers could not predict "recorded/environment/file" or "just true environment variables"
   - Refactors OpenApiKey resolution to prefer recorded variables with a fallback to the live environment, improving flexibility across test modes.

  ### Developer experience

   - Adds `.env.sample` as a template for optional environment variables.
   - Adds `.env` to `.gitignore` to prevent accidental secret commits.

  ### Test cleanup

   - Refactors UserAgentTests to use TestEnvironment.GetTestClient<T>() instead of manually constructing clients, aligning with the standard test patterns.
   - Removes unused using directive in OpenAIRecordedTestBase.
   - Removes the legacy `TestHelpers` class, unifying test environment and state management through the `OpenAITestEnvironment`

### Dependencies

   - Bumps Microsoft.ClientModel.TestFramework from 1.0.0-alpha.20260121.1 to 1.0.0-alpha.20260212.1.

### DIFF
--- a/tests/Utility/OpenAITestEnvironment.cs
+++ b/tests/Utility/OpenAITestEnvironment.cs
@@ -69,7 +69,7 @@ public class OpenAITestEnvironment : TestEnvironment
         EnvironmentFile = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
         {
             { EnvironmentVariables.OpenAIKey, "api-key" },
-            { EnvironmentVariables.TestMode, "live" }
+            { EnvironmentVariables.TestMode, "Live" }
         };
         _ = TryReadEnvFile(Path.Combine(RepositoryRoot, ".env"), EnvironmentFile);
         _ = TryReadEnvFile(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), ".env"), EnvironmentFile);
@@ -113,12 +113,18 @@ public class OpenAITestEnvironment : TestEnvironment
                 // OpenAI does not support bootstrapping of test resources, so disable it.
                 DisableBootstrapping = true;
 
-                // If a test mode was set, then use that as the default global mode for the environment.
+                // If a test mode was set, then use that as the default global mode for the environment.  Enum.Parse
+                // will accept undefined members, such as "9999", so validate that the parsed value is defined before accepting it.
                 var testMode = GetOptionalVariable(EnvironmentVariables.TestMode);
 
-                if ((testMode is { }) && (Enum.IsDefined(typeof(RecordedTestMode), testMode)))
+                if ((testMode is { }) && (Enum.TryParse<RecordedTestMode>(testMode, ignoreCase: true, out var parsedTestMode)))
                 {
-                    GlobalTestMode = Enum.Parse<RecordedTestMode>(testMode);
+                    if (!Enum.IsDefined(typeof(RecordedTestMode), parsedTestMode))
+                    {
+                        throw new FormatException($"The test mode value '{testMode}' is not valid. Accepted values are: {string.Join(", ", Enum.GetNames(typeof(RecordedTestMode)))}");
+                    }
+
+                    GlobalTestMode = parsedTestMode;
                 }
 
                 s_baseInitialized = true;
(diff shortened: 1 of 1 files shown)


.env.sample
```diff
@@ -0,0 +1,5 @@
+# Application environment variables sample file
+# Copy this file to .env and fill in the required values
+
+OPENAI_AI_KEY=<< YOUR API KEY >>
```
ShivangiReja: nit: Typo - `OPENAI_AI_KEY` should be `OPENAI_API_KEY`

tests/Batch/BatchTests.cs
```diff
@@ -11,17 +11,14 @@
...
 namespace OpenAI.Tests.Batch;
 
 [Category("Batch")]
 [TestFixture(true)]
 [TestFixture(false)]
 public class BatchTests : OpenAIRecordedTestBase
-{
-    private BatchClient GetTestClient() => GetProxiedOpenAIClient<BatchClient>(TestScenario.Batch);
-    private static readonly DateTimeOffset s_2024 = new(2024, 01, 01, 0, 0, 0, TimeSpan.Zero);
+{    private static readonly DateTimeOffset s_2024 = new(2024, 01, 01, 0, 0, 0, TimeSpan.Zero);
```
ShivangiReja: nit: The opening brace and field declaration ended up on the same line.

tests/Embeddings/EmbeddingsTests.cs
```diff
@@ -178,7 +177,4 @@ public async Task EmbeddingFromStringAndEmbeddingFromTokenIdsAreEqual()
         {
             Assert.That(vector1.Span[i], Is.EqualTo(vector2.Span[i]).Within(0.0005));
         }
-    }
-
-    private EmbeddingClient GetTestClient() => GetProxiedOpenAIClient<EmbeddingClient>(TestScenario.Embeddings);
-}
+    }}
```
ShivangiReja: nit: Removing the `GetTestClient()` method collapsed two closing braces onto a single line. These should be on separate lines

tests/Responses/ResponsesTests.cs
```diff
@@ -212,7 +211,7 @@ public async Task ResponsesHelloWorldWithTool(string model)
     [RecordedTest]
     public async Task ResponsesWithReasoning()
     {
-        ResponsesClient client = GetTestClient("gpt-5");
+        ResponsesClient client = TestEnvironment.GetTestClient<ResponsesClient>("gpt-5");
```
ShivangiReja, line 214: Is it intentional that this test uses TestEnvironment.GetTestClient (bypassing the test proxy) while most other tests in this file use GetProxiedOpenAIClient? Same question for `CanGetInputTokenCounts` and `CanCompactConversation`.

jsquire: These are prior art.   My intent was to not change behavior and just update where the call was made.   Some of these appear to be tagged as `[RecordedTest]` but are really just unit/mock style.   Chat has several as well.

.env.sample
jsquire: Fun story:  Chat test recordings had that, as there was a typo in the existing `OpenAITestEnvironment`.   I didn't notice until debugging and, apparently, after I copy/pasted here.  😄

ShivangiReja (approved):

Issue #949, opened 2026-02-12
[BUG]Error mesage: The requested operation requires an element of type 'String', but the target element has type 'Array'.


sevzas:
### Describe the bug

I call ResponseSvc.CreateResponse(responseOptions) and after a few seconds I see:

The requested operation requires an element of type 'String', but the target element has type 'Array'.

   at System.Text.Json.ThrowHelper.ThrowJsonElementWrongTypeException(JsonTokenType expectedType, JsonTokenType actualType)
   at System.Text.Json.JsonDocument.GetString(Int32 index, JsonTokenType expectedType)
   at OpenAI.Responses.ResponseResult.DeserializeResponseResult(JsonElement element, BinaryData data, ModelReaderWriterOptions options)
   at OpenAI.Responses.ResponseResult.op_Explicit(ClientResult result)
   at OpenAI.Responses.ResponsesClient.CreateResponse(CreateResponseOptions options, CancellationToken cancellationToken)
   at <<<My code>>>

I'm migrating well-tested code from Assistants to Prompts.  I'm using Prompts to provide instructions, model, etc.  

I set the prompt id using 

                responseOptions.Patch.Set(Encoding.UTF8.GetBytes("$.prompt.id").AsSpan(), GetPromptId())

The inputItems contains 1 ResponseItem that is made up of 1 ResponseContentPart which is built from a string.

Since it takes a few seconds for the exception to appear after calling ResponseSvc.CreateResponse(), it appears that there is a problem extracting the response, not sending the request.  How can I troubleshoot this further and maybe see the raw message that is being returned?  Is there something else I can try?

### Steps to reproduce

1. create inputItems with 1 ResponseItem that is made up of 1 ResponseContentPart which is built from a string.
2. create a responseOptions with the inputItems 
3. responseOptions.Patch.Set(Encoding.UTF8.GetBytes("$.prompt.id").AsSpan(), GetPromptId())   // Unfortunately I can't share my Prompt publicly
4. create a response ResponseSvc.CreateResponse()

### Code snippets

```C#

```

### OS

windows

### .NET version

9.0.0

### Library version

2.8.0

jsquire:
[...]

PR #963, opened 2026-02-19
Update Responses with latest tsp

christothes:


src/Generated/Models/Responses/ResponseResult.cs
```diff
@@ -95,6 +96,8 @@ internal ResponseResult(IDictionary<string, string> metadata, float? temperature
```
jsquire, line 96: `Instructions` is now `string | string[] | null` and breaking serialization.    That exists in the OpenAPI upstream spec, we'll need to be sure it is included here for Feb.

src/Generated/Models/Responses/CreateResponseOptions.cs
```diff
@@ -42,6 +42,7 @@ internal CreateResponseOptions(IDictionary<string, string> metadata, float? temp
             TextOptions = textOptions;
             Tools = tools ?? new ChangeTrackingList<ResponseTool>();
             ToolChoice = toolChoice;
+            Prompt = prompt;
```
jsquire, line 45: Prompt has been a big request item for a while, if we're updating to the new spec, I think we need to make sure that saved prompts are in the convenience API.    This looks to be still hidden.

src/Generated/Models/Responses/InternalPrompt.cs
```diff
@@ -0,0 +1,44 @@
...
+#nullable disable
+
+using System.ClientModel.Primitives;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json.Serialization;
+
+namespace OpenAI.Responses
+{
+    internal partial class InternalPrompt
```
jsquire, line 12: Make me public!

src/Generated/Models/Responses/ResponseResult.cs
christothes: It still looks like this in the upstream spec:

`instructions?: string | null;`

jsquire: In the AIP spec?   Then they're behind.    Details are in #949, but the service docs and OpenAPI spec disagree.

- The reference docs are confirmed by the latest [OpenAI REST API spec](https://app.stainless.com/api/spec/documented/openai/openapi.documented.yml) on line 60,408:
```yaml
instructions:
  anyOf:
    - description: >
        A system (or developer) message inserted into the model's
        context.
        When using along with `previous_response_id`, the
        instructions from a previous
        response will not be carried over to the next response. This
        makes it simple
        to swap out system (or developer) messages in new responses.
      oneOf:
        - type: string
          description: >
            A text input to the model, equivalent to a text input
            with the
            `developer` role.
        - type: array
          title: Input item list
          description: >
            A list of one or many input items to the model,
            containing
            different content types.
          items:
            $ref: "#/components/schemas/InputItem"
    - type: "null"
```

ShivangiReja: It is up to date in the latest spec, and `instructions` property exists in both the input and output models for Create Response.

>It still looks like this in the upstream spec:

>instructions?: string | null;

@christothes That line is for the Create Response input(see [here](https://github.com/microsoft/openai-openapi-pr/blob/66dec95fdcffb9c1cb66e8370ecabbd4714d5557/packages/openai-typespec/src/responses/models.tsp#L2366))

@jsquire is referring to the Create Response output, which is also up to date(see [here](https://github.com/microsoft/openai-openapi-pr/blob/66dec95fdcffb9c1cb66e8370ecabbd4714d5557/packages/openai-typespec/src/responses/models.tsp#L3040C3-L3040C45))

>instructions: string | InputItem[] | null;

christothes: Thanks, I was getting the two confused.

PR #964, opened 2026-02-20
Add workflow to record tests

joseharriaga:
This is intended for the following scenario:

1. There is branch with changes that need new test recordings (e.g., a branch with tests written by the Copilot Coding Agent).
2. A user kicks off this workflow off that branch.
3. The workflow does the recordings and pushes them into a new PR. This new PR targets the corresponding branch.
4. The user merges the generated PR into the corresponding branch.


.github/workflows/record-test.yml
```diff
@@ -0,0 +1,111 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          BASE_BRANCH="${{ github.ref_name }}"
+          RECORDING_BRANCH="record-test/${{ github.run_id }}"
+          
+          git config --local user.email "github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git checkout -b "$RECORDING_BRANCH"
+          git add tests/SessionRecords/
+          git commit -m "Record tests matching: ${{ inputs.filter }}"
```
joseharriaga: @copilot Change the commit message to simply include the name of the branch that the workflow is targeting instead of including the filter.

.github/workflows/record-test.yml
```diff
@@ -0,0 +1,111 @@
...
+          git push origin "$RECORDING_BRANCH"
+          
+          gh pr create \
+            --base "$BASE_BRANCH" \
+            --head "$RECORDING_BRANCH" \
+            --title "Record tests matching: ${{ inputs.filter }}" \
```
joseharriaga: @copilot Change the PR title to simply include the name of the branch that the workflow is targeting instead of including the filter.

.github/workflows/record-test.yml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Run tests in Record mode
+        run: dotnet test ./tests/OpenAI.Tests.csproj
+          --configuration Release
+          --filter="${{ inputs.filter }}"
+          --logger "trx;LogFilePrefix=record"
+          --results-directory ${{ github.workspace }}/artifacts/test-results
+          ${{ env.version_suffix_args }}
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          CLIENTMODEL_TEST_MODE: Record
```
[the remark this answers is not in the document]
joseharriaga: @copilot Please fix this. Here is more information about character escaping: https://learn.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=nunit#character-escaping

.github/workflows/record-test.yml
```diff
@@ -0,0 +1,110 @@
...
+          git push origin "$RECORDING_BRANCH"
+          
+          gh pr create \
+            --base "$BASE_BRANCH" \
+            --head "$RECORDING_BRANCH" \
+            --title "Record tests matching: ${{ inputs.filter }}" \
+            --body "This PR contains session recordings generated by the Record test workflow.
+
+          **Filter:** \`${{ inputs.filter }}\`
+          **Workflow run:** ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
```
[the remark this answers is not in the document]
joseharriaga: @copilot After you make the changes to the commit message and the PR title, this issue will only affect the PR body. Please fix this. Here is more information about character escaping: https://learn.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=nunit#character-escaping

.github/workflows/record-test.yml
```diff
@@ -0,0 +1,131 @@
+# This workflow is triggered manually and runs tests in Record mode, then commits the session
+# recordings and creates a PR against the target branch.
+name: "[Test] Record tests"
```
christothes, line 5: We should be clear about if this is meant for only recording new tests or updating existing tests. I can imagine this being a bad thing if it were to update test recordings where we didn't expect the recordings to change.

joseharriaga: I updated the description. This workflow is intended to be used for recording new tests or re-recording existing tests on behalf of the Copilot Coding Agent. 
* It must be intentionally triggered manually against a branch in our repo. In other words, it cannot be triggered by third-parties and cannot target third-party branches.
* The workflow also does not make any changes automatically to the target branch. Instead, it creates a PR against it that must be approved and merged manually and intentionally.

.github/workflows/record-test.yml
```diff
@@ -0,0 +1,133 @@
...
+# Runs targeting different branches are allowed in parallel.
+concurrency:
+  group: record-test-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  record:
+    name: Record tests
+    runs-on: ubuntu-latest
+    environment: Live Testing
```
ShivangiReja, line 29: Do we need to add some timeout here? Tests call the live OpenAI API and could hang forever. Without a timeout, that burns runner minutes indefinitely.

.github/workflows/record-test.yml
```diff
@@ -0,0 +1,133 @@
...
+          if [ -z "$STATUS_OUTPUT" ]; then
+            echo "has_changes=false" >> $GITHUB_OUTPUT
+            echo "### No recording changes detected" >> $GITHUB_STEP_SUMMARY
+          else
+            echo "has_changes=true" >> $GITHUB_OUTPUT
+            echo "### Recording changes detected:" >> $GITHUB_STEP_SUMMARY
+            echo '```' >> $GITHUB_STEP_SUMMARY
+            echo "$STATUS_OUTPUT" >> $GITHUB_STEP_SUMMARY
+            echo '```' >> $GITHUB_STEP_SUMMARY
+          fi
```
ShivangiReja, line 81: Can we add a step to scan recordings for accidentally leaked secrets. The sanitizers in OpenAIRecordedTestBase strip known sensitive headers, properties but if a new field appears in a response that contains a key or token, it would get auto-committed and pushed to a public PR. Can add something to prevent this? Maybe add a quick grep before playback, something like:

```
      - name: Scan recordings for unsanitized secrets
        if: steps.changes.outputs.has_changes == 'true'
        run: |
          if grep -rP 'sk-[a-zA-Z0-9]{20,}' tests/SessionRecords/ --include='*.json'; then
            echo "::error::Potential unsanitized API key found in recordings!"
            exit 1
          fi
```

joseharriaga: The workflow has automatic secret detection, but I added the step just in case.

.github/workflows/record-test.yml
joseharriaga: Our clients have built-in timeouts, so this shouldn't be necessary.

.github/workflows/record-test.yml
```diff
@@ -0,0 +1,145 @@
...
+          
+          BODY_FILE=$(mktemp)
+          trap 'rm -f "$BODY_FILE"' EXIT
+
+          printf 'This PR contains session recordings generated by the record-test workflow.\n\n**Filter:** `%s`\n**Workflow run:** %s' \
+            "$TEST_FILTER" \
+            "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" \
+            > "$BODY_FILE"
+          
+          PR_URL=$(gh pr create \
```
jsquire, line 130: So the intent is:

- We have a PR where we need to record tests
- This thing kicks off and does the recordings, pushing to assets
- This thing opens a PR to push the assets.json update as a new PR?
- I would have to know to merge that new PR, then either rebase my original or update the assets file by hand?

joseharriaga: The session recordings live in this repo, so it looks more like this:

1. User has a branch where they need to record tests.
2. User kicks off this workflow off their branch.
3. The workflow does the recordings and pushes them into a new PR. This new PR targets the user's branch.
4. User merges the generated PR into their branch. They now have the recordings they wanted.

jsquire: Why not just push to the active PR branch as a co-author?

joseharriaga: Mostly just as a precaution for now, so that it doesn't accidentally mess up our branches or something. But if this ends up working end-to-end as well as I'm picturing it in my head, I would be down for changing it to be more seamless like that.

PR #965, opened 2026-02-20
fix(docs): correct broken link and formatting

alty-ir:


joseharriaga:
Thank you for your contribution, @alty-ir !

PR #966, opened 2026-02-21
Add support for ResponseResult.Instructions to handle string and ResponseItem[]

ShivangiReja:
The OpenAI Responses API returns the `instructions` field in the response as either a `string`, an array of `InputItem[]`, or `null`. The SDK was only handling the `string | null` case, causing a `JsonElementWrongTypeException`(issue https://github.com/openai/openai-dotnet/issues/949) when the service returned an array (e.g., when using saved prompts).

### Changes

**TypeSpec specification** (models.tsp)
- Moved `instructions` from the shared `ResponseProperties` model into `CreateResponse` (remains `string | null`) and `Response` (now `string | InputItem[] | null`) separately, reflecting that the request and response have different shapes.

**Custom code**
- **`ResponseResult.cs`** — Changed the `Instructions` property type from `BinaryData` to `IList<ResponseItem>` to support both string and array instruction formats.
- **`ResponseResult.Serialization.cs`** *(new file)* — Custom deserialization override that handles all three `instructions` formats:
  - `string` → wraps in a single `ResponseItem.CreateDeveloperMessageItem()`
  - `array` → deserializes each element as a `ResponseItem`
  - `null` → initializes an empty `ChangeTrackingList<ResponseItem>`
- **`CreateResponseOptions.cs`** — Fixed `Clone()` to copy `_patch`, preventing Patch operations from being lost when the convenience method calls `CreatePerCallOptions()`.

**Tests**
- `SingleResponseInstruction` — Verifies that a string `instructions` value round-trips correctly as a single `ResponseItem` with `Developer` role.
- `MultipleResponseInstructions` — Uses a saved prompt ID to trigger an array-format `instructions` response and verifies multiple `ResponseItem` entries are deserialized.

### DIFF
--- a/tests/Responses/ResponsesTests.cs
+++ b/tests/Responses/ResponsesTests.cs
@@ -992,7 +992,7 @@ public partial class ResponsesTests : OpenAIRecordedTestBase
         ResponsesClient client = GetProxiedOpenAIClient<ResponsesClient>();
 
         const string instructionsText = "Always begin your replies with 'Ahoy'";
-        CreateResponseOptions options = new([ResponseItem.CreateUserMessageItem("Hello, model!")])
+        CreateResponseOptions options = new("gpt-5", [ResponseItem.CreateUserMessageItem("Hello, model!")])
         {
             Instructions = instructionsText,
         };
@@ -1014,7 +1014,7 @@ public partial class ResponsesTests : OpenAIRecordedTestBase
     {
         ResponsesClient client = GetProxiedOpenAIClient<ResponsesClient>();
 
-        CreateResponseOptions options = new([
+        CreateResponseOptions options = new("gpt-5", [
             ResponseItem.CreateUserMessageItem([
                 ResponseContentPart.CreateInputTextPart("Hello, model!")
             ])
--- a/tests/SessionRecords/ResponsesTests/MultipleResponseInstructions.json
+++ b/tests/SessionRecords/ResponsesTests/MultipleResponseInstructions.json
@@ -36,9 +36,9 @@
       "ResponseHeaders": {
         "Alt-Svc": "h3=\":443\"",
         "cf-cache-status": "DYNAMIC",
-        "CF-RAY": "9d1272080f5872dd-SEA",
+        "CF-RAY": "9d39a0cf29e12814-SEA",
         "Connection": "keep-alive",
-        "Content-Length": "3570",
+        "Content-Length": "2605",
         "Content-Type": "application/json",
         "Date": "Sanitized",
         "openai-organization": "Sanitized",
@@ -52,21 +52,21 @@
         "x-ratelimit-limit-requests": "15000",
         "x-ratelimit-limit-tokens": "40000000",
         "x-ratelimit-remaining-requests": "14999",
-        "x-ratelimit-remaining-tokens": "39999856",
+        "x-ratelimit-remaining-tokens": "40000000",
         "x-ratelimit-reset-requests": "4ms",
         "x-ratelimit-reset-tokens": "0s",
         "X-Request-ID": "Sanitized"
       },
       "ResponseBody": {
-        "id": "resp_099a1de531fad7e400699909585ec8819589cb9edcf88b5fb5",
+        "id": "resp_04df1c2eda66ca3700699f4e4575388195bb315dc87931b3ee",
         "object": "response",
-        "created_at": 1771637080,
+        "created_at": 1772047941,
         "status": "completed",
         "background": false,
         "billing": {
           "payer": "developer"
         },
-        "completed_at": 1771637090,
+        "completed_at": 1772047946,
         "error": null,
         "frequency_penalty": 0.0,
         "incomplete_details": null,
@@ -97,12 +97,12 @@
         "model": "gpt-5.2-2025-12-11",
         "output": [
           {
-            "id": "rs_099a1de531fad7e40069990958c7ac8195a70c1554d285e5bb",
+            "id": "rs_04df1c2eda66ca3700699f4e4605b08195ac14dbef1ad1d8c2",
             "type": "reasoning",
             "summary": []
           },
           {
-            "id": "msg_099a1de531fad7e4006999095b225c8195a82e01a637c6e657",
+            "id": "msg_04df1c2eda66ca3700699f4e47a4c48195ae99f93972e424d3",
             "type": "message",
             "status": "completed",
             "content": [
@@ -110,7 +110,7 @@
                 "type": "output_text",
                 "annotations": [],
                 "logprobs": [],
-                "text": "HELLOOOO! And buckle up—here comes the full tale, quick and punchy.\n\nThree little pigs move out to build their own homes.  \n- Pig #1 goes fast: a house of **straw**.  \n- Pig #2 goes “good enough”: a house of **sticks**.  \n- Pig #3 goes full mastermind: a house of **bricks**, taking the time to do it right.\n\nAlong comes the Big Bad Wolf, hungry and dramatic. He storms up to the straw house: “Let me in!” The pig says no, so the wolf huffs, puffs, and **blows it down**. Pig #1 sprints to Pig #2’s stick house.\n\nThe wolf follows. Same demand, same refusal. He huffs, he puffs, and **blows the stick house down**, too. Now both pigs rocket over to Pig #3’s brick house.\n\nThe wolf tries his famous move again—huffing and puffing like he’s auditioning for a windstorm—but the brick house doesn’t budge. Furious, he decides to get sneaky and climbs onto the roof to come down the chimney. The third pig is ready: he lights a fire under a big pot. The wolf drops in, gets a scorching surprise, yelps, and **flees for good**.\n\nThe three pigs stay safe in the brick house—two of them learning, very loudly, that “quick and easy” isn’t always the strongest plan."
+                "text": "HELLO THERE! I’M HERE, I’M AWAKE, AND I’M READY TO GO.\n\nWant the **Three Little Pigs** story—brief, complete, and entertaining—or did you have something else in mind? If you say “pigs,” I’m delivering the whole tale like a freight train."
               }
             ],
             "role": "assistant"
@@ -150,11 +150,11 @@
           "input_tokens_details": {
             "cached_tokens": 0
           },
-          "output_tokens": 425,
+          "output_tokens": 148,
           "output_tokens_details": {
-            "reasoning_tokens": 111
+            "reasoning_tokens": 75
           },
-          "total_tokens": 494
+          "total_tokens": 217
         },
         "user": null,
         "metadata": {}
--- a/tests/SessionRecords/ResponsesTests/MultipleResponseInstructionsAsync.json
+++ b/tests/SessionRecords/ResponsesTests/MultipleResponseInstructionsAsync.json
@@ -36,9 +36,9 @@
       "ResponseHeaders": {
         "Alt-Svc": "h3=\":443\"",
         "cf-cache-status": "DYNAMIC",
-        "CF-RAY": "9d12724ec86572dd-SEA",
+        "CF-RAY": "9d39a10588d32814-SEA",
         "Connection": "keep-alive",
-        "Content-Length": "2567",
+        "Content-Length": "3980",
         "Content-Type": "application/json",
         "Date": "Sanitized",
         "openai-organization": "Sanitized",
@@ -51,21 +51,21 @@
         "x-ratelimit-limit-requests": "15000",
         "x-ratelimit-limit-tokens": "40000000",
         "x-ratelimit-remaining-requests": "14999",
-        "x-ratelimit-remaining-tokens": "40000000",
+        "x-ratelimit-remaining-tokens": "39999812",
         "x-ratelimit-reset-requests": "4ms",
         "x-ratelimit-reset-tokens": "0s",
         "X-Request-ID": "Sanitized"
       },
       "ResponseBody": {
-        "id": "resp_05902d8a608a62040069990963b23c8193b552429025c9bf69",
+        "id": "resp_0a38ae07be03205100699f4e4dbe2081908c38dd08471e8c1f",
         "object": "response",
-        "created_at": 1771637091,
+        "created_at": 1772047949,
         "status": "completed",
         "background": false,
         "billing": {
           "payer": "developer"
         },
-        "completed_at": 1771637095,
+        "completed_at": 1772047961,
         "error": null,
         "frequency_penalty": 0.0,
         "incomplete_details": null,
@@ -96,12 +96,12 @@
         "model": "gpt-5.2-2025-12-11",
         "output": [
           {
-            "id": "rs_05902d8a608a6204006999096415b48193bf6a47603a411db2",
+            "id": "rs_0a38ae07be03205100699f4e4e3370819081dbdd9c63078b8f",
             "type": "reasoning",
             "summary": []
           },
           {
-            "id": "msg_05902d8a608a62040069990966ad6c819395ed7a1200573399",
+            "id": "msg_0a38ae07be03205100699f4e5047d081909e5952f6d73ba1d6",
             "type": "message",
             "status": "completed",
             "content": [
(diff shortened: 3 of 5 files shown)


src/Custom/Responses/CreateResponseOptions.cs
```diff
@@ -161,6 +161,7 @@ internal CreateResponseOptions Clone()
         clone.StoredOutputEnabled = StoredOutputEnabled;
         clone.StreamingEnabled = StreamingEnabled;
         clone.ConversationOptions = ConversationOptions;
+        clone._patch = _patch;
```
jsquire: I don't think we can do a direct assignment here, because that transfers the reference from the source object to the clone.   If we mutate internally or the caller mutates externally, we've got an inconsistent representation in the clone.

I think we'd need to clone the patch as well.

api/OpenAI.net10.0.cs
```diff
@@ -6319,7 +6319,7 @@ public class ResponseResult : IJsonModel<ResponseResult>, IPersistableModel<Resp
         public ResponseError Error { get; set; }
         public string Id { get; set; }
         public ResponseIncompleteStatusDetails IncompleteStatusDetails { get; set; }
-        public string Instructions { get; set; }
+        public IList<ResponseItem> Instructions { get; set; }
```
jsquire: Shouldn't this be `IReadOnlyList<ResponseItem>` for an output type?

specification/base/typespec/responses/models.tsp
```diff
@@ -157,6 +164,13 @@ model Response {
     """)
   output: OutputItem[];
 
+  /**
+   * A system (or developer) message inserted into the model's context.
```
jsquire:
```suggestion
   * A collection of system (or developer) messages inserted into the model's context.
```

src/Custom/Responses/ResponseResult.Serialization.cs
```diff
@@ -0,0 +1,360 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Text.Json;
+using OpenAI;
+
+namespace OpenAI.Responses
+{
+    public partial class ResponseResult : IJsonModel<ResponseResult>
+    {
+        internal static ResponseResult DeserializeResponseResult(JsonElement element, BinaryData data, ModelReaderWriterOptions options)
```
jsquire: Is there any way that we can mark the generated serialization stub as "ignore instructions" and make it an internal member?   

That would let us handle just the "instructions is a string" case here with a patch transform and then just forward to the generated version.   That'd save us from having to write/maintain this.   

Maybe a visitor could help?    Or, maybe we do the transform in a custom policy that rewrites the raw response JSON and just use the generated serialization where it's always an array?

src/Custom/Responses/CreateResponseOptions.cs
ShivangiReja: Yes, and I looked into alternatives:

1. **Round-trip serialization** (`WriteTo` → `new JsonPatch(bytes)`) doesn't work because `WriteTo` is designed for *final output*, not for cloning:
```c#
// Deep-copy _patch via round-trip serialization
using var stream = new MemoryStream();
using (var writer = new Utf8JsonWriter(stream))
{
    _patch.WriteTo(writer);
}
clone._patch = new JsonPatch(stream.ToArray());
clone._patch.SetPropagators(clone.PropagateSet, clone.PropagateGet);
```
- It [drops removed properties](https://github.com/Azure/azure-sdk-for-net/blob/ec4f6315a02becc705f8ff4d2658b78d0fec704d/sdk/core/System.ClientModel/src/ModelReaderWriter/JsonPatch.Serialization.cs#L30-L33), so [`Patch.Remove("$.model"u8)`](https://github.com/ShivangiReja/openai-dotnet/blob/Shreja_FixResponseInstructions/tests/Responses/ResponsesTests.cs#L1026) would be lost, and `model` would reappear in the serialized output.
- The reconstructed patch [doesn't write its data](https://github.com/Azure/azure-sdk-for-net/blob/ec4f6315a02becc705f8ff4d2658b78d0fec704d/sdk/core/System.ClientModel/src/ModelReaderWriter/JsonPatch.Serialization.cs#L101-L106) when the `Utf8JsonWriter` already has content (which it does, because generated serialization writes known properties first).

2. **Deep copy** isn't possible. `JsonPatch` doesn't expose a `Clone()` method, and its internal dictionary type isn't accessible.

------------------------

Current fix is only the short term fix I could find. A proper `Clone()` method on `JsonPatch` would be the right long-term fix. Happy to take a different approach if you have a suggestion.

If we go with the current approach, then I think we should also add `SetPropagators`:

```csharp
clone._patch = _patch;
clone._patch.SetPropagators(clone.PropagateSet, clone.PropagateGet);
```

Without it, the clone's patch holds delegate references to the **original's** `PropagateGet`/`PropagateSet`, so during serialization the clone would read the original's property values instead of its own. The shared dictionary reference is safe here because `Clone()` is only called from [`CreatePerCallOptions()`](https://github.com/openai/openai-dotnet/blob/13f5f01c98f1532ece6733d93ea146a50c726967/src/Custom/Responses/ResponsesClient.cs#L632), where the clone is serialized immediately and never mutated further.

jsquire: Huh...  yeah.   I'm seeing what you're seeing.   I thought we could seed based on the full JSON object, but you're right - that doesn't pull the removal forward.

```csharp
var clonePatch = new JsonPatch(Encoding.UTF8.GetBytes(responseOptions.Patch.ToString("J")));
```

```text
responseOptions.Patch.ToString()
[{"op":"add","path":"/prompt","value":{"id":"pmpt_<<Number>>"}},{"op":"remove","path":"/model"}]

clonePatch.ToString()
[]

clonePatch.ToString("J")
{"prompt":{"id":"pmpt_<<Number>>"}}
```

We'll need to get a clone for patch in place, but I think your approach and suggestion to reset propagators is probably the best path in the short term.

src/Custom/Responses/ResponseResult.Serialization.cs
ShivangiReja: Added the `CodeGenSerialization` attribute. It lets us customize deserialization for an individual property, so we don’t have to move the entire model deserialization into custom code.

See here: https://github.com/ShivangiReja/openai-dotnet/blob/Shreja_FixResponseInstructions/src/Custom/Responses/ResponseResult.Serialization.cs#L9

src/Custom/Responses/ResponseResult.cs
```diff
@@ -10,6 +11,7 @@ namespace OpenAI.Responses;
 // - Suppressed client-only OutputText property in favor of a method.
 [CodeGenType("Response")]
 [CodeGenSuppress("OutputText")]
+[CodeGenSerialization(nameof(Instructions), DeserializationValueHook = nameof(DeserializeInstructions))]
```
jsquire: 👀 

I had no idea you could do this!

src/Custom/Responses/ResponseResult.Serialization.cs
```diff
@@ -0,0 +1,33 @@
...
+                    array.Add(ResponseItem.CreateDeveloperMessageItem(property.Value.GetString()));
+                }
+                else if (property.Value.ValueKind == JsonValueKind.Array)
+                {
+                    foreach (var item in property.Value.EnumerateArray())
+                    {
+                        array.Add(ResponseItem.DeserializeResponseItem(item, item.GetUtf8Bytes(), ModelSerializationExtensions.WireOptions));
+                    }
+                }
+                instructions = array;
```
jsquire:
```suggestion
                instructions = array.AsReadOnly();
```

Don't we need this to make it actually read only?

src/Custom/Responses/ResponseResult.Serialization.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        private static void DeserializeInstructions(JsonProperty property, ref IReadOnlyList<ResponseItem> instructions)
+        {
+            if (property.Value.ValueKind == JsonValueKind.Null)
+            {
+                instructions = new ChangeTrackingList<ResponseItem>();
+            }
+            else
+            {
+                List<ResponseItem> array = new List<ResponseItem>();
```
jsquire: nit:  Seeing `array` be defined as a list is weird.   Maybe we should consider a different name?

jsquire (approved):

src/Custom/Responses/ResponseResult.Serialization.cs
ShivangiReja: I don't think this is needed. This follows the established pattern used by all generated deserialization code. The property is typed as IReadOnlyList<ResponseItem>, so consumers already interact through the read-only interface.

src/Custom/Responses/ResponseResult.Serialization.cs
ShivangiReja: Agree but the generated code uses `array` as the variable name for lists throughout (e.g. [here](https://github.com/openai/openai-dotnet/blob/c12f45377f132ba8b2df7d4297108bd329f4e48e/src/Generated/Models/Responses/ResponseResult.Serialization.cs#L482)). I copied the pattern from there 😆 But sure, happy to rename it in this custom code for clarity.

jsquire: Your call.  😄

api/OpenAI.net10.0.cs
```diff
@@ -6319,7 +6319,7 @@ public class ResponseResult : IJsonModel<ResponseResult>, IPersistableModel<Resp
         public ResponseError Error { get; set; }
         public string Id { get; set; }
         public ResponseIncompleteStatusDetails IncompleteStatusDetails { get; set; }
-        public string Instructions { get; set; }
+        public IReadOnlyList<ResponseItem> Instructions { get; }
```
joseharriaga: I think that because this is a "protocol model", this does need to be an `IList<ResponseItem>` instead of `IReadOnly<ResponseItem>` even though this is an output model.

cc @christothes

src/Custom/Responses/ResponseResult.cs
```diff
@@ -10,6 +11,7 @@ namespace OpenAI.Responses;
...
 public partial class ResponseResult
```
joseharriaga, line 14: nit: I recommend moving the `[CodeGenSerialization]` attribute to the "ResponseResult.Serialization.cs" file so that it lives together with the implementation. 🙂

src/Custom/Responses/ResponseResult.Serialization.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.ClientModel.Primitives;
+using System.Collections.Generic;
+using System.Text.Json;
+
+namespace OpenAI.Responses
+{
+    public partial class ResponseResult : IJsonModel<ResponseResult>
+    {
+        private static void DeserializeInstructions(JsonProperty property, ref IReadOnlyList<ResponseItem> instructions)
```
joseharriaga: This signature is missing the `options` parameter:

 ```csharp
private static void DeserializeInstructions(JsonProperty property, ref IReadOnlyList<ResponseItem> instructions, ModelReaderWriterOptions options = null)
```

 Here's an example from another deserialization hook:
 🔗 https://github.com/openai/openai-dotnet/blob/f786f16acc21df99774683115d39ad3eea4172ee/src/Custom/Responses/Items/FunctionTool/FunctionCallResponseItem.Serialization.cs#L26

Comment (joseharriaga) [reply src/Custom/Responses/ResponseResult.Serialization.cs]: Also, add the `[MethodImpl(MethodImplOptions.AggressiveInlining)]` attribute for a minor performance improvement.

File: src/Custom/Responses/ResponseResult.Serialization.cs [hunk 11]
```diff
@@ -0,0 +1,33 @@
...
+                List<ResponseItem> items = new List<ResponseItem>();
+                if (property.Value.ValueKind == JsonValueKind.String)
+                {
+                    items.Add(ResponseItem.CreateDeveloperMessageItem(property.Value.GetString()));
+                }
+                else if (property.Value.ValueKind == JsonValueKind.Array)
+                {
+                    foreach (var item in property.Value.EnumerateArray())
+                    {
+                        items.Add(ResponseItem.DeserializeResponseItem(item, item.GetUtf8Bytes(), ModelSerializationExtensions.WireOptions));
```
joseharriaga: Once you have the `options` parameter, you no longer have to hard-code this to `ModelSerializationExtensions.WireOptions` and should be able to flow the `options` parameter instead.

src/Generated/Models/Responses/InternalCodeInterpreterToolCallItemParam.cs
```diff
@@ -29,7 +29,7 @@ internal InternalCodeInterpreterToolCallItemParam(InternalItemType kind, in Json
 
         public string ContainerId { get; set; }
 
-        public string Code { get; }
+        public string Code { get; set; }
```
joseharriaga, line 32: Do you know what caused all these internal classes to change? It doesn't seem to be a problem given that they are internal, but it's a little unexpected. 🤔

src/Generated/Models/Responses/ResponseResult.Serialization.cs
```diff
@@ -709,6 +726,16 @@ private bool PropagateGet(ReadOnlySpan<byte> jsonPath, out JsonPatch.EncodedValu
...
+            if (local.StartsWith("instructions"u8))
+            {
+                int propertyLength = "instructions"u8.Length;
+                ReadOnlySpan<byte> currentSlice = local.Slice(propertyLength);
+                if (!currentSlice.TryGetIndex(out int index, out int bytesConsumed))
+                {
+                    return false;
+                }
+                return Instructions[index].Patch.TryGetEncodedValue([.. "$"u8, .. currentSlice.Slice(bytesConsumed)], out value);
+            }
```
joseharriaga, line 738: @m-nash: 

### CONTEXT
This "instructions" property is a union between a `string` and a collection of `ResponseItem`. This change is exposing the `Instructions` property as an `IList<ResponseItem>`, and the deserialization code is customized so that if we ever receive a `string`, we transform it into a collection with a single `ResponseItem` made from that `string`. In other words, we are trying to hide the union and always treat it as a collection.

### QUESTION
Do we need to customize the `PropagateGet` and/or `PropagateSet` methods too to be able to handle a union like this one?

src/Custom/Responses/ResponseResult.Serialization.cs
ShivangiReja: @joseharriaga I already tried adding a `ModelReaderWriterOptions options` parameter, but the generated code doesn’t pass it through to the call site where this method is invoked. Using your example, here’s the spot I mean: https://github.com/openai/openai-dotnet/blob/f786f16acc21df99774683115d39ad3eea4172ee/src/Generated/Models/Responses/FunctionCallResponseItem.Serialization.cs#L127

I will create an issue in the code gen.

src/Generated/Models/Responses/InternalCodeInterpreterToolCallItemParam.cs
ShivangiReja: No idea!

ShivangiReja: No idea!

src/Custom/Responses/ResponseResult.Serialization.cs
joseharriaga: Good catch! Please link the issue that you create back to this PR for reference. 

I recommend making the change anyway even if it doesn't work today so that when the CodeGen issue is fixed, it fixes it here too.

ShivangiReja: https://github.com/microsoft/typespec/issues/9800

src/Generated/Models/Responses/ResponseResult.Serialization.cs
ShivangiReja: As far as I understand this, I think we don't need to customize PropagateGet/PropagateSet.

`instructions` is a named property, so it is always matched and deserialized via `DeserializeInstructions`, it never falls through to the catch-all that would store it in the Patch. By the time PropagateGet/PropagateSet could ever run, the string has already been turned into a list item.

Also, `ResponseResult.Instructions` is output-only, users read it, they don't modify it. So PropagateGet/PropagateSet for `instructions` would never actually be called. The generator just emits them for every collection property automatically.

PR #983, opened 2026-02-27
Remove GA prefix from Realtime types

joseharriaga:


src/Custom/Realtime/GA/AudioFormat/RealtimePcmaAudioFormat.cs
```diff
@@ -4,6 +4,6 @@ namespace OpenAI.Realtime;
 
 // CUSTOM: Renamed.
 [CodeGenType("RealtimeAudioFormatsAudioPcmaGA")]
-public partial class GARealtimePcmaAudioFormat
+public partial class RealtimePcmaAudioFormat
```
ShivangiReja, line 7: Do we want to keep the `GA` folder?

joseharriaga: I didn't remove it to not make the diff weirder. We'll clean it up after. 🙂

PR #990, opened 2026-03-01
Fix UploadFile filename extraction from filepath

### LINKED ISSUE #308: UploadFile/Async with filePath sets whole path as filename
### Service

OpenAI

### Describe the bug

When using OpenAIFileClient.UploadFile/Async directly with a filepath. The Uploaded File will have the complete Path as a name

### Steps to reproduce

1. Upload file using a path
2. Check filename of uploaded file

### Code snippets

_No response_

### OS

winOS

### .NET version

net8.0

### Library version

2.0.0

04cb:
Fixes #308. Extracts filename from filepath using Path.GetFileName() instead of using the full path as the filename.

### DIFF
--- a/src/Custom/Files/OpenAIFileClient.cs
+++ b/src/Custom/Files/OpenAIFileClient.cs
@@ -213,8 +213,9 @@ public partial class OpenAIFileClient
     {
         Argument.AssertNotNullOrEmpty(filePath, nameof(filePath));
 
+        string fileName = Path.GetFileName(filePath);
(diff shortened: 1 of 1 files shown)


ShivangiReja:
Hi @04cb, thanks for the contribution! We already have a PR that addresses the same issue: https://github.com/openai/openai-dotnet/pull/951. We’re just waiting for it to merge, so I’m going to close this PR to avoid duplication.

PR #1007, opened 2026-03-05
Update tar to version 7.5.10

joseharriaga:


package-lock.json
```diff
@@ -150,431 +150,6 @@
         "node": ">=18"
       }
     },
-    "node_modules/@esbuild/aix-ppc64": {
```
jsquire, line 153: Did we expect the large amount of deletions?

joseharriaga: Not exactly, but I deleted these dependencies and everything appears to continue working fine, so I think it's a good thing that they are gone? Thoughts, @jorgerangel-msft ?

jorgerangel-msft: I think it should be fine. If there were issues with the deps, they _should_ get caught in ci when we run `npm ci`

Issue #916, opened 2026-01-23
[Audio] Add support for diarized transcription response


ShivangiReja:
Add client support for the diarized transcription response in the `TranscribeAudio`(`createTranscription`) operation.

### Implementation
Spread [`CreateTranscriptionResponseDiarizedJson`](https://github.com/openai/openai-dotnet/blob/fcc535799be0b5b8bb9bdd32a060959ae77c2a95/specification/base/typespec/audio/models.tsp#L185) into [`DotNetCombinedJsonTranscriptionResponse`](https://github.com/openai/openai-dotnet/blob/fcc535799be0b5b8bb9bdd32a060959ae77c2a95/specification/client/models/audio.models.tsp#L10) in `audio.models.tsp`.

### Reference
- [`CreateTranscriptionResponseDiarizedJson`](https://github.com/openai/openai-dotnet/blob/fcc535799be0b5b8bb9bdd32a060959ae77c2a95/src/Generated/Models/Audio/InternalCreateTranscriptionResponseDiarizedJson.cs)

PR #1008, opened 2026-03-05
[Audio] Add support for diarized audio transcription

ShivangiReja:
## Summary

Adds client support for the diarized audio transcription.

### New API surface

- **`AudioClient.TranscribeAudioDiarized` / `TranscribeAudioDiarizedAsync`** — four new method overloads (sync/async × `Stream`/file path) that return `DiarizedAudioTranscription`
- **`DiarizedAudioTranscription`** — response model containing `Text`, `Duration`, `Segments`, and `Usage`
- **`DiarizedTranscriptionSegment`** — segment model with `Speaker`, `Text`, `StartTime`, `EndTime`, and `Logprob`
- **`TranscriptionUsage`** — base usage type with discriminated subtypes:
  - `TranscriptionTokenUsage` (with `InputTokens`, `OutputTokens`, `TotalTokens`, `InputTokenDetails`)
  - `TranscriptionDurationUsage` (with `Seconds`)
- **`TranscriptionUsageKind`** — extensible enum (`Tokens`, `Duration`)
- **`TranscriptionInputTokenDetails`** — `TextTokens`, `AudioTokens`
- **`AudioTranscriptionFormat.DiarizedJson`** — now public
- **`AudioTranscriptionOptions.KnownSpeakerNames`** and **`KnownSpeakerReferences`** — now public

### Tests & examples

- Five new tests: `DiarizedTranscriptionWorks`, `DiarizedTranscriptionWithKnownSpeakersWorks`, `DiarizedTranscriptionHasUsage`, `DiarizedTranscriptionSegmentsAreOrdered`, `TranscriptionUsageWorks`
- Two new examples: `Example05_DiarizedTranscription` (sync and async)
- New test model entry: `gpt-4o-transcribe-diarize`

Fixes: https://github.com/openai/openai-dotnet/issues/916

### DIFF
--- a/src/Custom/Audio/AudioClient.cs
+++ b/src/Custom/Audio/AudioClient.cs
@@ -202,7 +202,7 @@ public partial class AudioClient
         {
             throw new InvalidOperationException(
                 $"{nameof(AudioTranscriptionOptions.ResponseFormat)} must not be set to {nameof(AudioTranscriptionFormat.Diarized)} when calling {nameof(TranscribeAudio)}. "
-                + $"For diarized transcription, call {nameof(TranscribeAudioDiarized)} instead.");
+                + $"For diarized transcription, call {nameof(TranscribeAudioDiarizedAsync)} instead.");
         }
 
         using MultiPartFormDataBinaryContent content
@@ -299,7 +299,7 @@ public partial class AudioClient
         {
             throw new InvalidOperationException(
                 $"{nameof(AudioTranscriptionOptions.ResponseFormat)} must be {nameof(AudioTranscriptionFormat.Diarized)} when calling {nameof(TranscribeAudioDiarized)}. "
-                + $"For non-diarized transcription, call {nameof(TranscribeAudio)} instead.");
+                + $"For non-diarized transcription, call {nameof(TranscribeAudioAsync)} instead.");
         }
 
         using MultiPartFormDataBinaryContent content


api/OpenAI.net10.0.cs
```diff
@@ -1462,6 +1501,58 @@ public class StreamingAudioTranscriptionUpdate : IJsonModel<StreamingAudioTransc
         public TimeSpan StartTime { get; }
         public string Word { get; }
     }
+    [Experimental("OPENAI001")]
+    public class TranscriptionDurationUsage : TranscriptionUsage, IJsonModel<TranscriptionDurationUsage>, IPersistableModel<TranscriptionDurationUsage> {
```
joseharriaga: For consistency, could we preface the names of these classes with "Audio", e.g.: "AudioTranscriptionDurationUsage", "AudioTranscriptionInputTokenUsageDetails", etc.

api/OpenAI.net10.0.cs
```diff
@@ -1462,6 +1501,58 @@ public class StreamingAudioTranscriptionUpdate : IJsonModel<StreamingAudioTransc
...
+        public TimeSpan Seconds { get; }
+        protected override TranscriptionUsage JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected override TranscriptionUsage PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class TranscriptionInputTokenDetails : IJsonModel<TranscriptionInputTokenDetails>, IPersistableModel<TranscriptionInputTokenDetails> {
```
joseharriaga: For consistency, we should add the word "Usage", i.e., "TranscriptionInputTokenUsageDetails" (or "AudioTranscriptionInputTokenUsageDetails" as per my other comment).

api/OpenAI.net10.0.cs
```diff
@@ -1462,6 +1501,58 @@ public class StreamingAudioTranscriptionUpdate : IJsonModel<StreamingAudioTransc
         public TimeSpan StartTime { get; }
         public string Word { get; }
     }
+    [Experimental("OPENAI001")]
+    public class TranscriptionDurationUsage : TranscriptionUsage, IJsonModel<TranscriptionDurationUsage>, IPersistableModel<TranscriptionDurationUsage> {
+        public TimeSpan Seconds { get; }
```
joseharriaga: Since this is a `TimeSpan` and not a numeric type, we should change the name of the property. In Realtime, this property is called "Duration".

api/OpenAI.net10.0.cs
```diff
@@ -1462,6 +1501,58 @@ public class StreamingAudioTranscriptionUpdate : IJsonModel<StreamingAudioTransc
...
+        protected override TranscriptionUsage JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected override TranscriptionUsage PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class TranscriptionInputTokenDetails : IJsonModel<TranscriptionInputTokenDetails>, IPersistableModel<TranscriptionInputTokenDetails> {
+        public int? AudioTokens { get; }
+        public int? TextTokens { get; }
```
joseharriaga: These two properties should be renamed to "AudioTokenCount" and "TextTokenCount".

api/OpenAI.net10.0.cs
```diff
@@ -1462,6 +1501,58 @@ public class StreamingAudioTranscriptionUpdate : IJsonModel<StreamingAudioTransc
...
+        protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected virtual TranscriptionInputTokenDetails PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class TranscriptionTokenUsage : TranscriptionUsage, IJsonModel<TranscriptionTokenUsage>, IPersistableModel<TranscriptionTokenUsage> {
+        public TranscriptionInputTokenDetails InputTokenDetails { get; }
+        public int InputTokens { get; }
+        public int OutputTokens { get; }
+        public int TotalTokens { get; }
```
joseharriaga: These properties should be renamed to "InputTokenCount", "OutputTokenCount", and "TotalTokenCount".

api/OpenAI.net10.0.cs
```diff
@@ -1462,6 +1501,58 @@ public class StreamingAudioTranscriptionUpdate : IJsonModel<StreamingAudioTransc
...
+    [Experimental("OPENAI001")]
+    public class TranscriptionUsage : IJsonModel<TranscriptionUsage>, IPersistableModel<TranscriptionUsage> {
+        public TranscriptionUsageKind Kind { get; }
+        protected virtual TranscriptionUsage JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected virtual TranscriptionUsage PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public readonly partial struct TranscriptionUsageKind : IEquatable<TranscriptionUsageKind> {
```
joseharriaga: Does the `TranscriptionUsageKind` and the `Kind` property of `TranscriptionUsage` need to be public? Or can they be kept internal as produced by the generator?

api/OpenAI.net10.0.cs
```diff
@@ -1174,6 +1174,14 @@ public class AudioClient {
...
         public virtual Task<ClientResult<AudioTranscription>> TranscribeAudioAsync(string audioFilePath, AudioTranscriptionOptions options = null);
         [Experimental("OPENAI001")]
+        public virtual ClientResult<DiarizedAudioTranscription> TranscribeAudioDiarized(Stream audio, string audioFilename, AudioTranscriptionOptions options = null, CancellationToken cancellationToken = default);
+        [Experimental("OPENAI001")]
+        public virtual ClientResult<DiarizedAudioTranscription> TranscribeAudioDiarized(string audioFilePath, AudioTranscriptionOptions options = null);
+        [Experimental("OPENAI001")]
+        public virtual Task<ClientResult<DiarizedAudioTranscription>> TranscribeAudioDiarizedAsync(Stream audio, string audioFilename, AudioTranscriptionOptions options = null, CancellationToken cancellationToken = default);
+        [Experimental("OPENAI001")]
+        public virtual Task<ClientResult<DiarizedAudioTranscription>> TranscribeAudioDiarizedAsync(string audioFilePath, AudioTranscriptionOptions options = null);
```
joseharriaga, line 1183: I think we need streaming counterparts of these methods too: "TranscribeAudioDiarizedStreaming"

api/OpenAI.net10.0.cs
ShivangiReja: Made these public for consistency with other discriminated types in the library (e.g., [`ResponseMessageAnnotation`](https://github.com/openai/openai-dotnet/blob/a95ca8194e174fea01bf1ed787b3707b2bc8cf2f/api/OpenAI.net10.0.cs#L7440)). Users could use pattern matching (`if (usage is AudioTranscriptionTokenUsage tokenUsage)`), but having `Kind` public keeps the API consistent and lets users inspect the discriminator without casting.

api/OpenAI.net10.0.cs
ShivangiReja: For streaming, separate diarized methods aren't needed. The existing `TranscribeAudioStreaming` methods already support diarization, users just need to set `options.ResponseFormat = AudioTranscriptionFormat.DiarizedJson`. The streaming events (e.g., `StreamingAudioTranscriptionTextSegmentUpdate`) will include diarization data like speaker labels when that format is specified.

api/OpenAI.net10.0.cs
```diff
@@ -1229,8 +1239,17 @@ public class AudioTranscription : IJsonModel<AudioTranscription>, IPersistableMo
...
+    public class AudioTranscriptionDurationUsage : AudioTranscriptionUsage, IJsonModel<AudioTranscriptionDurationUsage>, IPersistableModel<AudioTranscriptionDurationUsage> {
+        public TimeSpan Duration { get; }
+        protected override AudioTranscriptionUsage JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected override AudioTranscriptionUsage PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected override BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
     public readonly partial struct AudioTranscriptionFormat : IEquatable<AudioTranscriptionFormat> {
         public AudioTranscriptionFormat(string value);
+        public static AudioTranscriptionFormat DiarizedJson { get; }
```
joseharriaga: Since the user doesn't directly interact with JSON in the .NET library, we should rename this enum value to simply be "Diarized".

api/OpenAI.net10.0.cs
```diff
@@ -1254,9 +1273,22 @@ public enum AudioTranscriptionIncludes {
...
+        protected virtual AudioTranscriptionInputTokenUsageDetails PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
     public class AudioTranscriptionOptions : IJsonModel<AudioTranscriptionOptions>, IPersistableModel<AudioTranscriptionOptions> {
         [Experimental("OPENAI001")]
         public AudioTranscriptionIncludes Includes { get; set; }
+        [Experimental("OPENAI001")]
+        public IList<string> KnownSpeakerNames { get; }
+        [Experimental("OPENAI001")]
+        public IList<string> KnownSpeakerReferences { get; }
```
joseharriaga: This one should be a `IList<Uri>` since I think this is supposed to be a list of data URIs. I wonder if we should also call it `KnownSpeakerReferenceUris` for consistency with other data URI properties.

api/OpenAI.net10.0.cs
joseharriaga: Personally, I agree that the `Kind` property is useful. Having said that, we have historically been very inconsistent with how we apply this pattern across the library. I think we need to formalize it. In particular, exposing the `Kind` property is something that we want in our protocol models, which is why `ResponseMessageAnnotation` has it. (Note that only the Responses API has protocol models today).

I think we need a plan to become consistent. This is what I'm currently thinking:
1. We should not expose the `Kind` property manually anywhere going forward to avoid propagating the inconsistent behavior.
2. Instead we should add this functionality as part of the [ProtocolModelVisitor](https://github.com/openai/openai-dotnet/blob/main/codegen/generator/src/Visitors/ProtocolModelVisitor.cs).

Regarding Responses:
3. The `Kind` property of the `ResponseMessageAnnotation` class needs to be fixed: First, we revert the custom changes and then we expose it via the visitor instead.
4. We should do the same to all applicable types in Responses to achieve consistency.

Finally, circling back to the Audio API:
5. Because the Audio API doesn't have protocol models today, we should simply adopt the generator's default behavior for now, which is to keep the `Kind` property hidden. This will help us stop propagating inconsistencies.
6. I would go as far as to suggest that we should remove the public `StreamingAudioTranscriptionUpdateKind` that exists today too for consistency.
7. When we add protocol models to the Audio API (which might or might not be the same models that we have today, given that we have already GAed some of them, contrary to the models in Responses), we will expose the `Kind` property in those.

api/OpenAI.net10.0.cs
```diff
@@ -1319,6 +1369,29 @@ public class AudioTranslationOptions : IJsonModel<AudioTranslationOptions>, IPer
...
+        protected virtual DiarizedAudioTranscription JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected virtual DiarizedAudioTranscription PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class DiarizedTranscriptionSegment : IJsonModel<DiarizedTranscriptionSegment>, IPersistableModel<DiarizedTranscriptionSegment> {
+        public TimeSpan EndTime { get; }
+        public string Id { get; }
+        public string Speaker { get; }
```
joseharriaga: Rename to `SpeakerLabel` to align with `StreamingAudioTranscriptionTextSegmentUpdate`. This should also help make it clear that this is related to the "known speaker NAME" provided in the input options, but different in that if a speaker name is not provided, the transcription generates a LABEL for it.

api/OpenAI.net10.0.cs
```diff
@@ -1319,6 +1369,29 @@ public class AudioTranslationOptions : IJsonModel<AudioTranslationOptions>, IPer
...
+        public IReadOnlyList<DiarizedTranscriptionSegment> Segments { get; }
+        public string Text { get; }
+        public AudioTranscriptionUsage Usage { get; }
+        protected virtual DiarizedAudioTranscription JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        protected virtual DiarizedAudioTranscription PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class DiarizedTranscriptionSegment : IJsonModel<DiarizedTranscriptionSegment>, IPersistableModel<DiarizedTranscriptionSegment> {
```
joseharriaga: I wonder if we should make this a `public readonly partial struct` instead of a `public class`, and call it `DiarizedTranscribedSegment` to align with the existing `TranscribedSegment` and `TranscribedWord`. I recall we made those structs because a typical transcription would have a ton of segments, which I expect is the case with diarized transcriptions too. 🤔 What do you think?

src/Custom/Audio/AudioClient.cs
```diff
@@ -263,6 +264,94 @@ public virtual ClientResult<AudioTranscription> TranscribeAudio(string audioFile
...
+    ///     The filename associated with the audio stream. The filename's extension (for example: .mp3) will be used to
+    ///     validate the format of the input audio. The request may fail if the filename's extension and the actual
+    ///     format of the input audio do not match.
+    /// </param>
+    /// <param name="options"> The options to configure the audio transcription. </param>
+    /// <param name="cancellationToken"> A token that can be used to cancel this method call. </param>
+    /// <exception cref="ArgumentNullException"> <paramref name="audio"/> or <paramref name="audioFilename"/> is null. </exception>
+    /// <exception cref="ArgumentException"> <paramref name="audioFilename"/> is an empty string, and was expected to be non-empty. </exception>
+    [Experimental("OPENAI001")]
+    public virtual async Task<ClientResult<DiarizedAudioTranscription>> TranscribeAudioDiarizedAsync(Stream audio, string audioFilename, AudioTranscriptionOptions options = null, CancellationToken cancellationToken = default)
```
joseharriaga, line 293: We should add a check to these new methods and also to the old methods to detect if the wrong `AudioTranscriptionFormat` is passed in the `AudioTranscriptionOptions`. For example, if someone calls this `TranscribeAudioDiarizedAsync` method, but the format is something other than `AudioTranscriptionFormat.Diarized`, it should throw. Ideally, we should also add smoke tests to validate that this exception is thrown.

Throwing the exception is important because not adding this check means that the request could succeed, but we won't be able to correctly deserialize the response because it's a different type than we expected (for example, we get a diarized transcription when we expected a simple transcription, or vice versa).

We did something similar in Responses for streaming vs. non-streaming that you can probably get inspiration from, here:
🔗 https://github.com/openai/openai-dotnet/blob/a95ca8194e174fea01bf1ed787b3707b2bc8cf2f/src/Custom/Responses/ResponsesClient.cs#L113-L118

tests/Audio/TranscriptionTests.cs
```diff
@@ -184,6 +184,30 @@ public async Task TranscriptionFormatsWork(string responseFormat)
         }
     }
 
+    [RecordedTest]
+    public async Task TranscriptionUsageWorks()
```
joseharriaga: Should we add a fixture to ensure that this test validates both paths: token usage and duration usage. If I recall correctly, this depends on what model is used, correct?

src/Custom/Audio/AudioClient.cs
ShivangiReja: Discussed offline and updated based on: The `TranscribeAudioDiarized` method is strictly for when you want to use the `diarized_json` format and nothing else. Then, we need two checks:
 * If calling the `TranscribeAudioDiarized` method and the `ResponseFormat` **is not** `AudioTranscriptionFormat.Diarized`, then throw.
 * If calling the `TranscribeAudio` method and the `ResponseFormat` **is** `AudioTranscriptionFormat.Diarized`, then throw.

src/Custom/Audio/AudioClient.cs
```diff
@@ -197,6 +198,13 @@ public virtual async Task<ClientResult<AudioTranscription>> TranscribeAudioAsync
         Argument.AssertNotNull(audio, nameof(audio));
         Argument.AssertNotNullOrEmpty(audioFilename, nameof(audioFilename));
 
+        if (options?.ResponseFormat == AudioTranscriptionFormat.Diarized)
+        {
+            throw new InvalidOperationException(
+                $"{nameof(AudioTranscriptionOptions.ResponseFormat)} must not be set to {nameof(AudioTranscriptionFormat.Diarized)} when calling {nameof(TranscribeAudio)}. "
+                + $"For diarized transcription, call {nameof(TranscribeAudioDiarized)} instead.");
```
joseharriaga: nit: "call {nameof(TranscribeAudioDiarizedAsync)} instead."

src/Custom/Audio/AudioClient.cs
```diff
@@ -263,6 +278,108 @@ public virtual ClientResult<AudioTranscription> TranscribeAudio(string audioFile
...
+    public virtual async Task<ClientResult<DiarizedAudioTranscription>> TranscribeAudioDiarizedAsync(Stream audio, string audioFilename, AudioTranscriptionOptions options = null, CancellationToken cancellationToken = default)
+    {
+        Argument.AssertNotNull(audio, nameof(audio));
+        Argument.AssertNotNullOrEmpty(audioFilename, nameof(audioFilename));
+
+        if (options?.ResponseFormat is not null && options.ResponseFormat != AudioTranscriptionFormat.Diarized)
+        {
+            throw new InvalidOperationException(
+                $"{nameof(AudioTranscriptionOptions.ResponseFormat)} must be {nameof(AudioTranscriptionFormat.Diarized)} when calling {nameof(TranscribeAudioDiarized)}. "
+                + $"For non-diarized transcription, call {nameof(TranscribeAudio)} instead.");
```
joseharriaga: nit: "call {nameof(TranscribeAudioAsync)} instead."

joseharriaga (approved):

PR #1036, opened 2026-03-19
[Audio] Add ChunkingStrategy for audio transcription

ShivangiReja:
Adds chunking strategy support for audio transcription requests.

- New type: `AudioTranscriptionChunkingStrategy` (supports both string `"auto"` or `VadConfig` object)
- Renamed `VadConfig` to `AudioTranscriptionCustomChunkingStrategy` with `TimeSpan` duration properties
- Added `AudioTranscriptionDefaultChunkingStrategy` struct with `Auto` value
- Custom serialization writes bare string for default strategy and object for custom strategy 
- Exposed `AudioTranscriptionOptions.ChunkingStrategy` publicly
- Recorded tests for auto and custom chunking strategies

### DIFF
--- a/api/OpenAI.net10.0.cs
+++ b/api/OpenAI.net10.0.cs
@@ -1044,19 +1044,11 @@ namespace OpenAI.Audio {
         public AudioTranscriptionChunkingStrategy(AudioTranscriptionDefaultChunkingStrategy defaultChunkingStrategy);
         public AudioTranscriptionCustomChunkingStrategy CustomChunkingStrategy { get; }
         public AudioTranscriptionDefaultChunkingStrategy? DefaultChunkingStrategy { get; }
-        protected virtual AudioTranscriptionChunkingStrategy JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
-        protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
         public static implicit operator AudioTranscriptionChunkingStrategy(AudioTranscriptionCustomChunkingStrategy customChunkingStrategy);
         public static implicit operator AudioTranscriptionChunkingStrategy(AudioTranscriptionDefaultChunkingStrategy defaultChunkingStrategy);
-        protected virtual AudioTranscriptionChunkingStrategy PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
-        protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
     }
     [Experimental("OPENAI001")]
     public class AudioTranscriptionCustomChunkingStrategy : IJsonModel<AudioTranscriptionCustomChunkingStrategy>, IPersistableModel<AudioTranscriptionCustomChunkingStrategy> {
-        protected virtual AudioTranscriptionCustomChunkingStrategy JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
-        protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
-        protected virtual AudioTranscriptionCustomChunkingStrategy PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
-        protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
     }
     [Experimental("OPENAI001")]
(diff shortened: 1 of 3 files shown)


src/Custom/Audio/AudioTranscriptionCustomChunkingStrategy.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Microsoft.TypeSpec.Generator.Customizations;
+using System;
+
+namespace OpenAI.Audio;
+
+// CUSTOM: Renamed.
+[CodeGenType("VadConfig")]
+[CodeGenVisibility(nameof(Kind), CodeGenVisibility.Internal)]
```
joseharriaga: In order to futureproof this API, I think we need to expose this as a hierarchy (a base type and a derived type) in case that another type of chunking strategy is added later. There are hints that this might actually happen. For example: In Realtime, server VAD is used for turn-detection, but there is also an alternative called semantic VAD. I could semantic VAD making its way here as a chunking strategy. (For more info, see `RealtimeServerVadTurnDetection` versus `RealtimeSemanticVadTurnDetection`).

Because of this, I recommend having a base class with this name (`AudioTranscriptionCustomChunkingStrategy`) and adding a derived type called `AudioTranscriptionCustomServerVadChunkingStrategy` (a little bit of a mouthful...).

api/OpenAI.net10.0.cs
```diff
@@ -1240,6 +1240,44 @@ public class AudioTranscription : IJsonModel<AudioTranscription>, IPersistableMo
...
+        protected virtual AudioTranscriptionChunkingStrategy JsonModelCreateCore(ref Utf8JsonReader reader, ModelReaderWriterOptions options);
+        protected virtual void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options);
+        public static implicit operator AudioTranscriptionChunkingStrategy(AudioTranscriptionCustomChunkingStrategy customChunkingStrategy);
+        public static implicit operator AudioTranscriptionChunkingStrategy(AudioTranscriptionDefaultChunkingStrategy defaultChunkingStrategy);
+        protected virtual AudioTranscriptionChunkingStrategy PersistableModelCreateCore(BinaryData data, ModelReaderWriterOptions options);
+        protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options);
+    }
+    [Experimental("OPENAI001")]
+    public class AudioTranscriptionCustomChunkingStrategy : IJsonModel<AudioTranscriptionCustomChunkingStrategy>, IPersistableModel<AudioTranscriptionCustomChunkingStrategy> {
+        public float? ChunkingStrategyThreshold { get; set; }
```
joseharriaga: nit: Rename to "DetectionThreshold" to tie it back to the concept of voice activity detection (VAD) and be consistent with `RealtimeServerVadTurnDetection`.

src/Custom/Audio/AudioTranscriptionCustomChunkingStrategy.cs
christothes: Is "custom" necessary in the name? That seems implied. 

Maybe we could shorten the base class to `AudioTranscriptionChunkingStrategy` or maybe even `TranscriptionChunkingStrategy` with the derrived class being `TranscriptionServerVadChunkingStrategy` or `ServerVadChunkingStrategy`

ShivangiReja: @joseharriaga: Done. I've added the hierarchy following the `RealtimeTurnDetection` pattern. Previously I looked into [RealtimeServerVadTurnDetection](https://github.com/openai/openai-dotnet/blob/main/src/Custom/Realtime/GA/TurnDetection/RealtimeServerVadTurnDetection.cs#L8) and didn't realize there was a base class to it.

@christothes: [`AudioTranscriptionChunkingStrategy`](https://github.com/ShivangiReja/openai-dotnet/blob/Shreja_AudioChunkingStrategy/src/Custom/Audio/AudioTranscriptionChunkingStrategy.cs) is already taken — it's the top-level union type that wraps both components: the enum component with `"auto"` value (`AudioTranscriptionDefaultChunkingStrategy`) and the object-based component. "Custom" distinguishes the object hierarchy (where you manually configure parameters) from the "default" string enum.

This matches the existing Realtime tracing pattern:
- Union: `RealtimeTracing`  
- Enum component: `RealtimeDefaultTracing`  
- Object component: `RealtimeCustomTracing`

Dropping "Custom" would require renaming the union type itself, breaking the established convention.

PR #1037, opened 2026-03-19
Fix generator version upgrade

MaiLinhP:
1. Named unions for image streaming events
Added images.models.tsp with named unions (GeneratedImageSize, GeneratedImageQuality, GeneratedImageBackground, GeneratedImageOutputFormat) to replace inline unions that caused duplicate-client-name errors.

2. Removed broken generated files
Deleted src/Generated/Models/.cs and .Serialization.cs — generated with empty type names, causing compile errors.

3. Internalized speech audio event types
Remapped SpeechAudioDeltaEvent/SpeechAudioDoneEvent from public (root namespace) to internal (OpenAI.Audio) via [CodeGenType] stubs. Added OpenAIClientSettings stub and _keyCredential field for new generated dependencies.

### DIFF
--- a/api/OpenAI.net10.0.cs
+++ b/api/OpenAI.net10.0.cs
@@ -9,6 +9,8 @@
 namespace OpenAI {
     public class OpenAIClient {
         protected OpenAIClient();
+        [Experimental("SCME0002")]
+        public OpenAIClient(OpenAIClientSettings settings);
         public OpenAIClient(ApiKeyCredential credential, OpenAIClientOptions options);
         public OpenAIClient(ApiKeyCredential credential);
         [Experimental("OPENAI001")]
@@ -55,6 +57,12 @@ namespace OpenAI {
         public string OrganizationId { get; set; }
         public string ProjectId { get; set; }
         public string UserAgentApplicationId { get; set; }
+    }
+    [Experimental("SCME0002")]
+    public class OpenAIClientSettings : ClientSettings {
+        public Uri Endpoint { get; set; }
+        public OpenAIClientOptions Options { get; set; }
+        protected override void BindCore(Microsoft.Extensions.Configuration.IConfigurationSection section);
     }
     [Experimental("OPENAI001")]
     public class OpenAIContext : ModelReaderWriterContext {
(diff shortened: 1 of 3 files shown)


api/OpenAI.net10.0.cs
```diff
@@ -9,6 +9,8 @@
 namespace OpenAI {
     public class OpenAIClient {
         protected OpenAIClient();
+        [Experimental("SCME0002")]
+        public OpenAIClient(OpenAIClientSettings settings);
```
joseharriaga, line 13: The `OpenAIClient` is special. For now, we don't want it to have a Settings class, so we should suppress this constructor and the `OpenAIClientSettings` class. In our case, the individual feature clients (such as `AudioClient`) each have a Settings class instead (such as `AudioClientSettings`).

src/Custom/OpenAIClient.cs
```diff
@@ -93,6 +93,7 @@ private static class KnownHeaderNames
     }
 
     private readonly OpenAIClientOptions _options;
+    private ApiKeyCredential _keyCredential;
```
joseharriaga, line 96: Is this needed, or can it be removed?

specification/client/models/images.models.tsp
```diff
@@ -0,0 +1,63 @@
+import "../../base/typespec/images/main.tsp";
+import "@azure-tools/typespec-client-generator-core";
+
+using Azure.ClientGenerator.Core;
+
+namespace OpenAI;
+
+// Named unions to replace inline unions in image streaming events
+// These fix duplicate-client-name errors for unnamed inline unions
```
joseharriaga, line 9: Could you help me understand what caused this fix to be necessary? At a glance, it looks like a regression in how either TCGC or the generator deal with anonymous unions. 🤔

joseharriaga (changes requested):

specification/client/models/images.models.tsp
MaiLinhP: @joseharriaga I've attempt the upgrade on newest version instead [here](https://github.com/openai/openai-dotnet/pull/1044)! Will close this PR out!

PR #1039, opened 2026-03-19
Factor Responses into a separate package

christothes:


api/OpenAI.Responses.net10.0.cs
```diff
@@ -0,0 +1,2320 @@
...
+// </auto-generated>
+//------------------------------------------------------------------------------
+namespace OpenAI {
+    [Experimental("OPENAI001")]
+    public enum CreateResponseRequestAccept {
+        ApplicationJson = 0,
+        TextEventStream = 1
+    }
+    [Experimental("OPENAI001")]
+    public abstract class DotNetAssistantResponseFormat : IJsonModel<DotNetAssistantResponseFormat>, IPersistableModel<DotNetAssistantResponseFormat> {
```
joseharriaga: There are a lot of types here that should not be public, including (but not limited to) those prefixed by "DotNet".

OpenAI/src/Generated/Models/ApproximateLocation.cs
```diff
@@ -0,0 +1,34 @@
...
+#nullable disable
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace OpenAI
+{
+    [Experimental("OPENAI001")]
+    public partial class ApproximateLocation : Location
```
joseharriaga: There should not be any models in the `OpenAI` namespace.

api/OpenAI.net10.0.cs
```diff
@@ -7,6377 +7,214 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 namespace OpenAI {
-    public class OpenAIClient {
```
joseharriaga, line 9: Looks like a lot of stuff was unintentionall deleted. Instead, I expect the diff of the original API listings to only include all the OpenAI.Responses stuff deleted because it was moved to different files.

api/OpenAI.net10.0.cs
```diff
@@ -7,6377 +7,214 @@
...
-        protected OpenAIClient();
-        public OpenAIClient(ApiKeyCredential credential, OpenAIClientOptions options);
```
joseharriaga, line 11: The diffs for OpenAI.net8.0.cs and OpenAI.netstandard2.0.cs are missing.

OpenAI/OpenAI.sln
```diff
@@ -0,0 +1,48 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
joseharriaga: Is this file needed given the OpenAI.slnx file at the root of the repo? Or maybe it's the other way around and the OpenAI.slnx file at the root should go away, and this should be .slnx instead of .sln?

Responses/OpenAI.Responses.sln
```diff
@@ -0,0 +1,48 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
joseharriaga: Is this file needed given the OpenAI.slnx file at the root of the repo? Or maybe it's the other way around and the OpenAI.slnx file at the root should go away, and this should be .slnx instead of .sln?

Responses/src/Custom/OpenAIClient.cs
```diff
@@ -0,0 +1,9 @@
+using Microsoft.TypeSpec.Generator.Customizations;
+
+namespace OpenAI.Responses
+{
+    [CodeGenType("OpenAIClient")]
+    public partial class OpenAIClient
```
joseharriaga: There should not be an `OpenAIClient` class in the `OpenAI.Responses` namespace.

Responses/src/Custom/OpenAIClientOptions.cs
```diff
@@ -0,0 +1,9 @@
+using Microsoft.TypeSpec.Generator.Customizations;
+
+namespace OpenAI.Responses
+{
+    [CodeGenType("OpenAIClientOptions")]
+    public partial class OpenAIClientOptions
```
joseharriaga: There should not be an `OpenAIClientOptions` class in the `OpenAI.Responses` namespace.

Responses/src/Custom/OpenAIContext.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.ClientModel.Primitives;
+using System.Diagnostics.CodeAnalysis;
+
+namespace OpenAI.Responses;
+
+[Experimental("OPENAI001")]
+public partial class OpenAIContext : ModelReaderWriterContext
```
joseharriaga: There should not be an `OpenAIContext` class in the `OpenAI.Responses` namespace, but there should be an `OpenAIResponsesContext` class instead (which should be generated).

joseharriaga (changes requested):

OpenAI/OpenAI.sln
christothes: No, these sln files should go away.

Responses/OpenAI.Responses.sln
christothes: No, these sln files should go away.

api/OpenAI.Responses.net10.0.cs
christothes: Thanks for catching this. I ran into a bunch of issues with generator attributes not taking effect anymore once things were moved around. So there's definitely a class of them that I missed.

PR #1047, opened 2026-03-23
[Audio] Add SSE streaming support for speech generation

### LINKED ISSUE #914: [Audio] Add support for SSE streaming events for speech
Add client support for speech streaming events, enabling streaming from the Text-to-Speech API.

**Events to Support**
* speech.audio.delta
* speech.audio.done

Referecnce: 
* Spec: [CreateSpeechStreamingResponse](https://github.com/openai/openai-dotnet/blob/fcc535799be0b5b8bb9bdd32a060959ae77c2a95/specification/base/typespec/audio/models.tsp#L745-L755)
* Generated code: [SpeechAudioDeltaEvent](https://github.com/openai/openai-dotnet/blob/fcc535799be0b5b8bb9bdd32a060959ae77c2a95/src/Generated/Models/Audio/InternalDotNetSpeechAudioDeltaEvent.cs), [SpeechAudioDoneEvent](https://github.com/openai/openai-dotnet/blob/fcc535799be0b5b8bb9bdd32a060959ae77c2a95/src/Generated/Models/Audio/InternalDotNetSpeechAudioDoneEvent.cs)

ShivangiReja:
Adds SSE streaming support for the speech generation (Text-to-Speech) API, enabling consumers to receive audio data incrementally via `speech.audio.delta` and `speech.audio.done` server-sent events.

Resolves #914

### New public types
- **`StreamingSpeechUpdate`** — Base type for all speech streaming SSE events
- **`StreamingSpeechAudioDeltaUpdate`** — Represents `speech.audio.delta` events containing incremental audio data (`BinaryData Audio`)
- **`StreamingSpeechAudioDoneUpdate`** — Represents `speech.audio.done` events signaling completion, with token usage information
- **`SpeechTokenUsage`** — Token usage details (`InputTokenCount`, `OutputTokenCount`, `TotalTokenCount`)

### New client methods
- `AudioClient.GenerateSpeechStreamingAsync()` — Returns `AsyncCollectionResult<StreamingSpeechUpdate>`
- `AudioClient.GenerateSpeechStreaming()` — Returns `CollectionResult<StreamingSpeechUpdate>`

Both methods automatically set `stream_format=sse` on the request.

### Model validation
- Added `EnsureModelSupportsSpeechStreaming()` guard that throws `NotSupportedException` for `tts-1` and `tts-1-hd` (SSE streaming is not supported for these models per the API docs). Includes an `OPENAI_ENABLE_TTS_STREAMING` environment variable bypass.

### Tests
- **`StreamingSpeechWorks`** — Recorded integration test verifying delta/done events and token usage
- **`GenerateSpeechStreamingThrowsForUnsupportedModel`** — Mock test verifying the model guard for `tts-1` and `tts-1-hd`

---

### DIFF
--- a/src/Custom/Audio/AudioClient.cs
+++ b/src/Custom/Audio/AudioClient.cs
@@ -514,12 +514,12 @@ public partial class AudioClient
     {
         if (string.Equals(_model, "whisper-1", StringComparison.OrdinalIgnoreCase))
         {
-            string isEnabled = Environment.GetEnvironmentVariable("OPENAI_ENABLE_WHISPER_1_STREAMING");
+            string isEnabled = Environment.GetEnvironmentVariable("OPENAI_ENABLE_TRANSCRIPTION_SSE_STREAMING");
             if (!string.Equals(isEnabled, "true", StringComparison.OrdinalIgnoreCase))
             {
                 throw new NotSupportedException(
-                    "The selected model 'whisper-1' does not support streaming transcription. " +
-                    "Please use a compatible model or set the environment variable 'OPENAI_ENABLE_WHISPER_1_STREAMING=true' to bypass this check.");
+                    "The selected model 'whisper-1' does not support SSE streaming transcription. " +
+                    "Please use a compatible model or set the environment variable 'OPENAI_ENABLE_TRANSCRIPTION_SSE_STREAMING=true' to bypass this check.");
             }
         }
     }
@@ -529,12 +529,12 @@ public partial class AudioClient
         if (string.Equals(_model, "tts-1", StringComparison.OrdinalIgnoreCase)
             || string.Equals(_model, "tts-1-hd", StringComparison.OrdinalIgnoreCase))
         {
-            string isEnabled = Environment.GetEnvironmentVariable("OPENAI_ENABLE_TTS_STREAMING");
+            string isEnabled = Environment.GetEnvironmentVariable("OPENAI_ENABLE_TTS_SSE_STREAMING");
             if (!string.Equals(isEnabled, "true", StringComparison.OrdinalIgnoreCase))
             {
                 throw new NotSupportedException(
                     $"The selected model '{_model}' does not support SSE streaming for speech generation. "
-                    + "Please use a compatible model or set the environment variable 'OPENAI_ENABLE_TTS_STREAMING=true' to bypass this check.");
+                    + "Please use a compatible model or set the environment variable 'OPENAI_ENABLE_TTS_SSE_STREAMING=true' to bypass this check.");
             }
         }
     }
--- a/tests/Audio/GenerateSpeechMockTests.cs
+++ b/tests/Audio/GenerateSpeechMockTests.cs
@@ -41,6 +41,6 @@ internal class GenerateSpeechMockTests : ClientTestBase
             () => client.GenerateSpeechStreaming("text", GeneratedSpeechVoice.Alloy),
             Throws.InstanceOf<NotSupportedException>()
                 .With.Message.Contains(model)
-                .And.Message.Contains("OPENAI_ENABLE_TTS_STREAMING"));
+                .And.Message.Contains("OPENAI_ENABLE_TTS_SSE_STREAMING"));
     }
 }


tests/Utility/TestModel.cs
```diff
@@ -3,6 +3,7 @@ namespace OpenAI.Tests;
 internal static class TestModel
 {
     public const string Audio_TTS = "tts-1";
+    public const string Audio_TTS_Streaming = "gpt-4o-mini-tts";
```
joseharriaga: Would it make sense to simply update the model of `Audio_TTS` instead of adding a separate one?

api/OpenAI.net10.0.cs
```diff
@@ -1265,6 +1275,17 @@ public class StreamingAudioTranscriptionTextSegmentUpdate : StreamingAudioTransc
     [Experimental("OPENAI001")]
     public class StreamingAudioTranscriptionUpdate : IJsonModel<StreamingAudioTranscriptionUpdate>, IPersistableModel<StreamingAudioTranscriptionUpdate> {
     }
+    [Experimental("OPENAI001")]
+    public class StreamingSpeechGenerationAudioDeltaUpdate : StreamingSpeechGenerationUpdate, IJsonModel<StreamingSpeechGenerationAudioDeltaUpdate>, IPersistableModel<StreamingSpeechGenerationAudioDeltaUpdate> {
+        public BinaryData Audio { get; }
```
joseharriaga: Rename "Audio" to "AudioBytes".

api/OpenAI.net10.0.cs
```diff
@@ -1265,6 +1275,17 @@ public class StreamingAudioTranscriptionTextSegmentUpdate : StreamingAudioTransc
...
+    }
+    [Experimental("OPENAI001")]
+    public class StreamingSpeechGenerationAudioDoneUpdate : StreamingSpeechGenerationUpdate, IJsonModel<StreamingSpeechGenerationAudioDoneUpdate>, IPersistableModel<StreamingSpeechGenerationAudioDoneUpdate> {
+        public SpeechGenerationTokenUsage Usage { get; }
+    }
+    [Experimental("OPENAI001")]
+    public class StreamingSpeechGenerationUpdate : IJsonModel<StreamingSpeechGenerationUpdate>, IPersistableModel<StreamingSpeechGenerationUpdate> {
```
joseharriaga: We want to keep the names of these classes as close as possible to the name of the event they represent. As such, I recommend renaming these to:
- "StreamingSpeechUpdate"
- "StreamingSpeechAudioDeltaUpdate" (for "speech.audio.delta")
- "StreamingSpeechAudioDoneUpdate" (for "speech.audio.done")

api/OpenAI.net10.0.cs
```diff
@@ -1243,6 +1247,12 @@ public class SpeechGenerationOptions : IJsonModel<SpeechGenerationOptions>, IPer
         public float? SpeedRatio { get; set; }
     }
     [Experimental("OPENAI001")]
+    public class SpeechGenerationTokenUsage : IJsonModel<SpeechGenerationTokenUsage>, IPersistableModel<SpeechGenerationTokenUsage> {
```
joseharriaga: We usually skip the verb in these names (e.g. "ChatTokenUsage", "ImageTokenUsage", "ResponseTokenUsage"), so this should simply be "SpeechTokenUsage".

src/Custom/Audio/AudioClient.cs
```diff
@@ -476,6 +524,21 @@ private void EnsureModelSupportsStreaming()
         }
     }
 
+    private void EnsureModelSupportsSpeechStreaming()
```
joseharriaga, line 527: Question: If you pass the `"stream"` property to the "tts-1" or "tts-1-hd" models, do they simply ignore it (like "whisper-1" does)? Or does the service actually fail like one would expect?

src/Custom/Audio/Internal/GeneratorStubs.cs
```diff
@@ -28,10 +30,3 @@ namespace OpenAI.Audio;
 [CodeGenType("TranscriptTextDoneEventLogprobs")] internal partial class InternalTranscriptTextDoneEventLogprobs { }
 [CodeGenType("TranscriptTextSegmentEvent")] internal partial class InternalTranscriptTextSegmentEvent { }
```
joseharriaga: Unrelated to this PR, but I just happened to notice them: Can these go away now? (Maybe in a separate PR).

src/Custom/Audio/AudioClient.cs
ShivangiReja: I tested with both `tts-1` and `tts-1-hd` and it silently ignore `stream_format=sse`. The service returns HTTP 200 with normal audio output, no error and no SSE events. Same behavior as `whisper-1`.

src/Custom/Audio/Internal/GeneratorStubs.cs
ShivangiReja: Removed these types.

src/Custom/Audio/AudioClient.cs
```diff
@@ -476,6 +524,21 @@ private void EnsureModelSupportsStreaming()
...
+    {
+        if (string.Equals(_model, "tts-1", StringComparison.OrdinalIgnoreCase)
+            || string.Equals(_model, "tts-1-hd", StringComparison.OrdinalIgnoreCase))
+        {
+            string isEnabled = Environment.GetEnvironmentVariable("OPENAI_ENABLE_TTS_STREAMING");
```
christothes: Should we document this env var somewhere?

ShivangiReja: I don't think we need separate documentation for this. It's highly unlikely that someone would want to disable these checks. The error message itself names the env var and explains its purpose, so anyone who hits it already gets the context they need. The override is really just a forward-compatibility safety valve in case OpenAI adds streaming support to those models before we release an SDK update, which is very unlikely.

src/Custom/Audio/AudioClient.cs
joseharriaga: Sounds good. Thank you for confirming!

src/Custom/Audio/AudioClient.cs
```diff
@@ -476,6 +524,21 @@ private void EnsureModelSupportsStreaming()
...
+            if (!string.Equals(isEnabled, "true", StringComparison.OrdinalIgnoreCase))
+            {
+                throw new NotSupportedException(
+                    $"The selected model '{_model}' does not support SSE streaming for speech generation. "
+                    + "Please use a compatible model or set the environment variable 'OPENAI_ENABLE_TTS_STREAMING=true' to bypass this check.");
```
joseharriaga: 1. I think you made a good call in making a distinction between the different types of streaming by explicitly referring to it as "SSE streaming" in the exception message. 
2. I also think you made another good call in not including the name of the model in the environment variable name to keep it more generic, because technically there is no guarantee that all future models will support streaming too.

Then, with these two points in mind, I was wondering the following:

1. Would it be a good idea to also include the distinction between the different types of streaming in the environment variable name and explicitly call it "OPENAI_ENABLE_TTS_SSE_STREAMING".
2. Would it be a good idea to also rename the environment variable used for streaming transcriptions to apply these two improvements accordingly (landing on something like: "OPENAI_ENABLE_TRANSCRIPTION_SSE_STREAMING")? Streaming transcriptions are attributed as experimental, so it is fine to rename.

PR #1052, opened 2026-03-25
Bugfix: Handle OpenAPI tool type

### LINKED ISSUE #1049: [BUG] `openapi` tools deserialize as unknown tools in Assistants and Responses
## Describe the bug
Azure Foundry (possibly others?) agent/tool payloads that contain a tool with `"type": "openapi"` are not deserialized into a typed OpenAPI tool model in `openai-dotnet`. Instead, they fall back to the unknown-tool path (`UnknownAssistantToolDefinition` / `InternalUnknownTool` depending on API surface).

That fallback strips or obscures key OpenAPI metadata such as the tool `name` and `description`, which are required for effective tool use by the model.

This affects both API paths:
- Assistants tool definitions
- Responses tools (modern API)

Expected behavior:
`type: "openapi"` should deserialize into a first-class OpenAPI tool type, preserving the nested OpenAPI payload including `name`, `description`, `spec`, and `auth`.

Actual behavior:
The payload is treated as an unknown tool and essential metadata is lost.

## Steps to reproduce
1. Obtain an Azure Foundry agent definition or tool payload containing a tool shaped roughly like this:
   - top-level: `{"type":"openapi","openapi":{...}}`
   - nested fields including `name`, `description`, `spec`, and `auth`
2. Deserialize that payload through the Assistants or Responses model path in `openai-dotnet`.
3. Inspect the resulting tool instance.
4. Observe that it is deserialized as an unknown-tool type instead of a typed OpenAPI tool model.
5. Observe that `name` / `description` are not preserved in the typed surface.

## Code snippets
```csharp
using System.Text.Json;
using OpenAI;

string json = """
{
  "type": "openapi",
  "openapi": {
    "name": "weather_api",
    "description": "Gets weather data",
    "auth": { "type": "anonymous" },
    "spec": {
      "openapi": "3.0.0",
      "info": { "title": "Weather API", "version": "1.0.0" }
    }
  }
}
""";

// Repro shape only; exact entry point depends on whether the payload is flowing
// through the Assistants or Responses model path.
JsonDocument doc = JsonDocument.Parse(json);

// Current behavior: falls back to an unknown tool model.
// Expected behavior: a typed OpenAPI tool model with name/description/spec/auth preserved.
```

## OS
Windows

## .NET version
.NET SDK 10.0.201

## Library version
Observed on the 2.8.x line while validating Azure Foundry agent/tool payloads.

brandonh-msft:
## What Changed
 
 Adds first-class OpenAPI tool handling for Azure Foundry-style tool payloads in both Assistants and Responses.
 
 This change:
 - adds `openapi` discriminator support in the TypeSpec source for Assistants and Responses
 - generates typed OpenAPI tool models instead of falling back to unknown-tool handling
 - preserves OpenAPI tool metadata such as `name`, `description`, `spec`, and `auth`
 - adds convenience custom-code wrappers/factory helpers for the new OpenAPI tool types
 - adds smoke-test coverage for deserialization and round-trip preservation
 
 ## Why
 
 Azure Foundry agent definitions can return tools with `type: "openapi"`. Before this fix, those payloads deserialized as unknown tools, which scrubbed things like name & description which are essential for the model to know when to request invocation of it.
  
 ## Task Reference
 
 Closes #1049
 
 ## Testing
 
Aside from the unit tests, I built the nupkg into my project and validated that the openapi's tool was getting properly deserialized with name, description, auth, and spec.

## Notes

Ideally, I'd like this fix to also be applied to the v2.8 branch of the SDK but I wasn't readily seeing a hotfix process, so I applied it to HEAD.

### DIFF
--- a/api/OpenAI.net10.0.cs
+++ b/api/OpenAI.net10.0.cs
@@ -99,6 +99,24 @@ namespace OpenAI {
         public static IClientBuilder AddVectorStoreClient(this Microsoft.Extensions.Hosting.IHostApplicationBuilder builder, string sectionName);
         public static IClientBuilder AddVideoClient(this Microsoft.Extensions.Hosting.IHostApplicationBuilder builder, string sectionName);
     }
+    [Experimental("OPENAI001")]
+    public class OpenApiTool : ResponseTool, IJsonModel<OpenApiTool>, IPersistableModel<OpenApiTool> {
+        public OpenApiTool(OpenApiToolDetails openapi);
+        public BinaryData Authentication { get; set; }
+        public string Description { get; set; }
+        public string Name { get; set; }
+        public OpenApiToolDetails Openapi { get; set; }
+        public BinaryData Specification { get; set; }
+    }
+    [Experimental("OPENAI001")]
+    public class OpenApiToolDefinition : ToolDefinition, IJsonModel<OpenApiToolDefinition>, IPersistableModel<OpenApiToolDefinition> {
+        public OpenApiToolDefinition(OpenApiToolDefinitionDetails openapi);
+        public BinaryData Authentication { get; set; }
+        public string Description { get; set; }
+        public string Name { get; set; }
+        public OpenApiToolDefinitionDetails Openapi { get; set; }
+        public BinaryData Specification { get; set; }
+    }
 }
 namespace OpenAI.Assistants {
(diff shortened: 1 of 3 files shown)


brandonh-msft:
Closing, but going to leave this branch around in case this work is desired later

PR #1056, opened 2026-03-26
[Audio] Add Voices API support to AudioClient

ShivangiReja:
## Summary

Adds protocol-level support for the OpenAI Custom Voices API (`/audio/voices` and `/audio/voice_consents`) to `AudioClient`, ingested from the upstream TypeSpec specification.

### TypeSpec specification
 - Added 7 voice models to models.tsp: `VoiceResource`, `VoiceConsentResource`, `VoiceConsentListResource`, `VoiceConsentDeletedResource`, `CreateVoiceRequest`, `CreateVoiceConsentRequest`, `UpdateVoiceConsentRequest`
 - Added 6 voice operations to operations.tsp: `listVoiceConsents`, `createVoiceConsent`, `deleteVoiceConsent`, `getVoiceConsent`, `updateVoiceConsent`, `createVoice`
 - Updated audio.client.tsp with `@@clientLocation` for all new operations and `@@convenientAPI(false)` to expose protocol methods only

 ### Generated code
 - `AudioClient.cs` — 6 new protocol method pairs (sync + async): `CreateVoice`, `CreateVoiceConsent`, `DeleteVoiceConsent`, `GetVoiceConsent`, `UpdateVoiceConsent`, `GetVoiceConsents`
 - `AudioClient.RestClient.cs` — REST pipeline definitions for the 6 new endpoints
 - Updated API surface files (OpenAI.net10.0.cs, OpenAI.net8.0.cs, OpenAI.netstandard2.0.cs)

### Tests
 - Added VoiceTests.cs with a live test for voice consent creation and voice creation flow
- Test is marked `[Ignore]` because the Custom Voices API requires organization-level access that is not yet available

 ## Notes
 - Only protocol methods are exposed 
 - All operations are routed to `AudioClient` via `@@clientLocation`
- No `@@clientName` renames were needed — generated names already follow .NET conventions


tests/Audio/VoiceTests.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.IO;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace OpenAI.Tests.Audio;
+
+[Category("Audio")]
+public partial class VoiceTests : OpenAIRecordedTestBase
+{
+    public VoiceTests(bool isAsync) : base(isAsync)
```
christothes, line 16: It looks like we still need test coverage for a few more APIs (I don't see any of the collection operations, for example).

ShivangiReja: Currently we only have protocol methods for the voice APIs. For protocol methods, we usually cover the primary end-to-end scenario rather than per-operation coverage. Also, we don't have access to test the Voices API yet (hence the `[Ignore]`), so we can't test or record today. Will expand coverage once access is enabled.

PR #1067, opened 2026-04-01
Add ApplyPatchTool for Responses

joseharriaga:


src/Custom/Responses/Tools/ApplyPatchTool/ApplyPatchTool.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Microsoft.TypeSpec.Generator.Customizations;
+
+namespace OpenAI.Responses;
+
+// CUSTOM: Renamed.
+[CodeGenType("ApplyPatchTool")]
```
jsquire, line 6: Do the right thing!   Expose the tool constructors!

src/Custom/Responses/Items/ApplyPatch/ApplyPatchCallOutputItem.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Microsoft.TypeSpec.Generator.Customizations;
+
+namespace OpenAI.Responses;
+
+// CUSTOM: Renamed.
+[CodeGenType("ApplyPatchToolCallOutputItemResource")]
+public partial class ApplyPatchCallOutputItem
+{
+    // NOTE: Contrary to other tools, the `Status` property of `ApplyPatchCallOutputItem` is set by the client instead
```
jsquire: Who is this comment for?  Seems detached and floating in an unusual space.

src/Generated/Models/Responses/ApplyPatchCallStatus.cs
```diff
@@ -0,0 +1,48 @@
...
+    [Experimental("OPENAI001")]
+    public readonly partial struct ApplyPatchCallStatus : IEquatable<ApplyPatchCallStatus>
+    {
+        private readonly string _value;
+        private const string InProgressValue = "in_progress";
+        private const string CompletedValue = "completed";
+
+        public ApplyPatchCallStatus(string value)
+        {
+            Argument.AssertNotNull(value, nameof(value));
```
jsquire, line 21: Why are we allowing empty?

src/Custom/Responses/Tools/ApplyPatchTool/ApplyPatchTool.cs
joseharriaga: Yes, the constructor is public here: https://github.com/openai/openai-dotnet/blob/a9bcff7018aec93884482feccfb3f54da59cacfb/src/Generated/Models/Responses/ApplyPatchTool.cs#L13

src/Custom/Responses/Items/ApplyPatch/ApplyPatchCallOutputItem.cs
joseharriaga: All the items have a status property that is customized to be optional. For example:
https://github.com/openai/openai-dotnet/blob/a9bcff7018aec93884482feccfb3f54da59cacfb/src/Custom/Responses/Items/FunctionTool/FunctionCallOutputResponseItem.cs#L13

This one is an exception based on service behavior. I added this comment here in case someone looks at this file and wonders why it is the only one that is different from literally all the other ones, and to let them know that it is intentional. 

When we do a review over the whole TypeSpec for Responses, I'm hoping we can move this disclaimer over there.

src/Generated/Models/Responses/ApplyPatchCallStatus.cs
joseharriaga: That is how extensible enums are produced by the generator. Should we file a bug on the generator?

jsquire: Maybe?  I guess if it isn't causing problems, its fine until it isn't.   I just can't think of a valid case for empty.

src/Custom/Responses/Items/ApplyPatch/ApplyPatchCallOutputItem.cs
jsquire: >This one is an exception based on service behavior. I added this comment here in case someone looks at this file and wonders why it is the only one that is different from literally all the other ones, and to let them know that it is intentional.

But there are no members in this file.   That was my confusion.   I get that we can't comment on the generated code directly, but the member this is associated with doesn't live here.

I'm not pushing to block, just seems like it's more confusing to have a comment here that discusses a customization that isn't here.

joseharriaga: The problem is that it's not commenting on a customization but on the lack of one. Basically, I'm trying to avoid people seeing all the other items and then noticing that this is the only one missing the customization that all the other items have and thinking it's a bug.

I don't feel strongly, so I moved the comment to the client.tsp.

src/Generated/Models/Responses/ApplyPatchCallStatus.cs
joseharriaga: I do think it is incorrect. However, since all the extensible enums across all our libraries work like this, and it hasn't been an issue as far as we know, I'm tempted to just leave it rather than introduce a (niche) behavioral breaking change.

jsquire: Agreed.

src/Custom/Responses/Items/ApplyPatch/ApplyPatchCallOutputItem.cs
jsquire: ooooohhhhhh.   That was the context that was missing for me.   You were telling us why it wasn't here.

src/Generated/Models/Responses/ApplyPatchCallItem.Serialization.cs
```diff
@@ -0,0 +1,202 @@
...
+                writer.WritePropertyName("status"u8);
+                writer.WriteStringValue(Status.Value.ToString());
+            }
+            if (!Patch.Contains("$.operation"u8))
+            {
+                writer.WritePropertyName("operation"u8);
+                writer.WriteObjectValue(Operation, options);
+            }
+            if (Optional.IsDefined(CreatedBy) && !Patch.Contains("$.created_by"u8))
+            {
```
ShivangiReja: The Status property looks like optional.

If a user creates an `ApplyPatchCallItem` via `CreateApplyPatchCallItem(callId, operation)` and serializes it (which is valid), Status will be null and `Status.Value` will throw `InvalidOperationException`. Every other optional property guards with Optional.IsDefined() (e.g., CreatedBy). This one is missing the guard.

specification/client/responses.client.tsp
```diff
@@ -65,6 +65,12 @@ using TypeSpec.HttpClient.CSharp;
 @@visibility(ImageGenToolCallItemResource.status, Lifecycle.Read);
 @@clientName(ImageGenToolCallItemResource.result, "ImageResultBytes");
 
+@@visibility(ApplyPatchToolCallItemResource.status, Lifecycle.Read);
+
+// NOTE: Contrary to other tools, the "status" property of `ApplyPatchToolCallOutputItemResource`
+// is set by the client instead of being set by the server, and it is also required. The "status"
+// property is used to indicate if the client successfuly applied the patch or not.
```
ShivangiReja:
```suggestion
// property is used to indicate if the client successfully applied the patch or not.
```

tests/Responses/ResponsesToolTests.cs
```diff
@@ -1245,6 +1245,144 @@ in client.CreateResponseStreamingAsync(responseOptions))
         Assert.That(searchItemId, Is.Not.Null.And.Not.Empty);
     }
 
+    [RecordedTest]
+    [TestCase("CreateFile")]
+    [TestCase("UpdateFile")]
+    [TestCase("DeleteFile")]
+    public async Task ApplyPatchToolWorks(string testCase)
+    {
```
ShivangiReja, line 1389: Should we also add streaming test for this?

src/Generated/Models/Responses/ApplyPatchCallItem.Serialization.cs
joseharriaga: Woah, nice catch! Something about this code is tripping the generator causing it to not do the right thing. We actually have a visitor to fix that behavior, but it needs to be corrected on a case-by-case basis. Turns out we had missed this issue in the Code Interpreter item and in the Image Generation item too. I'm fixing those, as well as the Apply Patch items. We should investigate the underlying issue in the near future.

tests/Responses/ResponsesToolTests.cs
joseharriaga: There are no streaming events that are specific to the Apply Patch tool, so a streaming test shouldn't be necessary in this case.

codegen/generator/src/OpenAILibraryVisitor.cs
```diff
@@ -38,11 +38,15 @@ public class OpenAILibraryVisitor : ScmLibraryVisitor
             [
                 new("Id", "id", isCollection: false),
             ],
+        ["ApplyPatchCallItem"] = [_readonlyStatusReplacementInfo],
+        ["ApplyPatchCallOutputItem"] = [_readonlyStatusReplacementInfo],
```
ShivangiReja: Looks like Status property is not optional is `ApplyPatchCallOutputItem`, so do we need to add this here?

joseharriaga: No, I was just too eager to add it to everything. 😋 Fixed.

ShivangiReja (approved):

PR #1071, opened 2026-04-03
Expose data URIs in Responses as strings

joseharriaga:


src/Custom/Responses/ResponseContentPart.cs
```diff
@@ -59,9 +59,22 @@ public static ResponseContentPart CreateInputImagePart(string imageFileId, Respo
...
     {
+        Argument.AssertNotNull(imageUri, nameof(imageUri));
+
+        return new InternalItemContentInputImage()
+        {
+            ImageUri = imageUri.AbsoluteUri,
+            Detail = imageDetailLevel,
+        };
+    }
+    public static ResponseContentPart CreateInputImagePart(BinaryData imageBytes, ResponseImageDetailLevel? imageDetailLevel = null)
```
ShivangiReja, line 71:
```suggestion
    }
    
    public static ResponseContentPart CreateInputImagePart(BinaryData imageBytes, ResponseImageDetailLevel? imageDetailLevel = null)
```

PR #1077, opened 2026-04-03
[Skills] Add Skills API support

ShivangiReja:
Adds protocol-only support for the new OpenAI Skills API, enabling upload, versioning, and management of reusable skill bundles.

## Changes

 **TypeSpec / Code Generation**
- Added Skills spec (models.tsp, operations.tsp)
- Created skills.client.tsp with `@@clientLocation` for all 13 operations and `@@convenientAPI(false)` for protocol-only generation
 
 **Custom Client**
- Added `SkillClient` with standard constructor chain following the `ContainerClient` pattern
- Added `SkillClientSettings` for DI support
 - Added `GetSkillClient()` to `OpenAIClient`
- Added internal GeneratorStubs.cs for generated type namespace mapping
 
 **Generated Code**
- SkillClient.cs — 13 operations (sync + async = 26 methods)
- `SkillClient.RestClient.cs` — request pipeline helpers
 - `InternalUpdateSkillDefaultVersionRequestContentType` model
 
**Tests**
- Added SkillTests.cs with two live tests: `BasicSkillWorks` and `BasicSkillVersionWorks` 
- Session recordings included

## Notes
 Protocol-only (no convenience methods) — convenience layer can be added later


christothes (approved):

jsquire (approved):

specification/base/typespec/containers/models.tsp
```diff
@@ -128,3 +128,15 @@ model CreateContainerFileBody {
...
+@discriminator("type")
+model ContainerSkill {
+  type: ContainerSkillType;
+}
+
+union ContainerSkillType {
+  string,
+  "skill_reference",
+  "inline",
+}
```
joseharriaga, line 142: Was this change in the Container spec intentional?

api/OpenAI.net10.0.cs
```diff
@@ -6707,6 +6711,54 @@ public class WebSearchToolLocation : IJsonModel<WebSearchToolLocation>, IPersist
...
+        public SkillClient(AuthenticationPolicy authenticationPolicy);
+        protected internal SkillClient(ClientPipeline pipeline, OpenAIClientOptions options);
+        public SkillClient(string apiKey);
+        [Experimental("OPENAI001")]
+        public Uri Endpoint { get; }
+        public ClientPipeline Pipeline { get; }
+        public virtual ClientResult CreateSkillJson(BinaryContent content, RequestOptions options = null);
+        public virtual Task<ClientResult> CreateSkillJsonAsync(BinaryContent content, RequestOptions options = null);
+        public virtual ClientResult CreateSkillMultipart(BinaryContent content, string contentType, RequestOptions options = null);
+        public virtual Task<ClientResult> CreateSkillMultipartAsync(BinaryContent content, string contentType, RequestOptions options = null);
```
joseharriaga: A couple of questions:
1. Do you know if there are any functional differences between `CreateSkillJson` and `CreateSkillMultipart`? 
2. Is the `CreateSkillMultipart` protocol method a superset of the `CreateSkillJson` protocol method? In other words, given that `CreateSkillMultipart` has the `contentType` parameter, does that mean that I can pass "application/json" and get the behavior of `CreateSkillJson`? If so, does this mean we can remove `CreateSkillJson` entirely? We might also want to double-check with Krzysztof what is the correct approach in these cases.

Same for `CreateSkillVersionJson`/`CrateSkillVersionMultipart`.

api/OpenAI.net10.0.cs
```diff
@@ -6707,6 +6711,54 @@ public class WebSearchToolLocation : IJsonModel<WebSearchToolLocation>, IPersist
...
+        public virtual ClientResult CreateSkillVersionMultipart(string skillId, BinaryContent content, string contentType, RequestOptions options = null);
+        public virtual Task<ClientResult> CreateSkillVersionMultipartAsync(string skillId, BinaryContent content, string contentType, RequestOptions options = null);
+        public virtual ClientResult DeleteSkill(string skillId, RequestOptions options = null);
+        public virtual Task<ClientResult> DeleteSkillAsync(string skillId, RequestOptions options = null);
+        public virtual ClientResult DeleteSkillVersion(string skillId, string version, RequestOptions options = null);
+        public virtual Task<ClientResult> DeleteSkillVersionAsync(string skillId, string version, RequestOptions options = null);
+        public virtual ClientResult GetSkill(string skillId, RequestOptions options = null);
+        public virtual Task<ClientResult> GetSkillAsync(string skillId, RequestOptions options = null);
+        public virtual ClientResult GetSkillContent(string skillId, RequestOptions options = null);
+        public virtual Task<ClientResult> GetSkillContentAsync(string skillId, RequestOptions options = null);
```
joseharriaga, line 6908: It appears that `GetSkillContent` downloads the skill as a zip bundle? If so, I'm thinking we should rename this to `DownloadSkillContent` for consistency with other APIs that download files.

Same for `GetSkillVersionContent` --> `DownloadSkillVersionContent`

api/OpenAI.net10.0.cs
```diff
@@ -6707,6 +6711,54 @@ public class WebSearchToolLocation : IJsonModel<WebSearchToolLocation>, IPersist
...
+        public virtual ClientResult GetSkillVersion(string skillId, string version, RequestOptions options = null);
+        public virtual Task<ClientResult> GetSkillVersionAsync(string skillId, string version, RequestOptions options = null);
+        public virtual ClientResult GetSkillVersionContent(string skillId, string version, RequestOptions options = null);
+        public virtual Task<ClientResult> GetSkillVersionContentAsync(string skillId, string version, RequestOptions options = null);
+        public virtual ClientResult GetSkillVersions(string skillId, int? limit = null, string order = null, string after = null, RequestOptions options = null);
+        public virtual Task<ClientResult> GetSkillVersionsAsync(string skillId, int? limit = null, string order = null, string after = null, RequestOptions options = null);
+        public virtual ClientResult UpdateSkillDefaultVersion(string skillId, string contentType, BinaryContent content, RequestOptions options = null);
+        public virtual Task<ClientResult> UpdateSkillDefaultVersionAsync(string skillId, string contentType, BinaryContent content, RequestOptions options = null);
```
joseharriaga: A couple of questions:
1. While this method is only used today to update the default version of a skill, I wonder if it could evolve over time to update other parts of a skill. If so, should we call it simply `UpdateSkill`?
2. Why does this method take a `contentType` parameter, and why is the ordering different to the multipart form-data methods (before the `BinaryContent` parameter versus after the `BinaryContent parameter)?

api/OpenAI.net10.0.cs
ShivangiReja: Great questions! I investigated this thoroughly via the spec, live testing, and comparison with the Videos area.

**1. Functional differences between CreateSkillJson and CreateSkillMultipart**

Both hit the same endpoint (`POST /skills`). The only difference is the `Content-Type` header:
- `CreateSkillMultipart` — caller provides `contentType` (intended: `multipart/form-data`)
- `CreateSkillJson` — hardcodes `Content-Type: application/json`

The OpenAPI spec defines both content types for the same `CreateSkillBody` schema:
```yaml
/skills:
    post:
      tags:
        - Skills
      summary: Create a new skill.
      operationId: CreateSkill
      parameters: []
      requestBody:
        content:
          multipart/form-data:
            schema:
              $ref: '#/components/schemas/CreateSkillBody'
          application/json:
            schema:
              $ref: '#/components/schemas/CreateSkillBody'
```

However, in live testing, the `files` field only accepts actual file uploads (`Uploadable`). Passing string-encoded file data (base64 data URIs) to either method(CreateSkillMultipart or  CreateSkillJson) returns HTTP 400: `"Invalid type for 'files[0]': expected a file, but got a string instead."` So while the JSON content type is spec-defined, I haven't found a valid JSON input that the API accepts today.

**2. Is CreateSkillMultipart a superset?**

No. Passing `"application/json"` as the `contentType` to `CreateSkillMultipart` also returns HTTP 400 with the same error. The API validates the payload format strictly against the content type — you can't use multipart's `contentType` parameter to get JSON behavior.

**3. How Videos handles this**

For comparison, the Videos spec uses **separate schemas** per content type:
```yaml
/videos:
    post:
      tags:
        - Videos
      summary: >-
        Create a new video generation job from a prompt and optional reference
        assets.
      operationId: createVideo
      parameters: []
      requestBody:
        content:
          multipart/form-data:
            schema:
              $ref: '#/components/schemas/CreateVideoMultipartBody'
          application/json:
            schema:
              $ref: '#/components/schemas/CreateVideoJsonBody'
```

And the .NET SDK collapses them into a **single method** with a caller-provided `contentType`:
```csharp
public virtual ClientResult CreateVideo(BinaryContent content, string contentType, RequestOptions options = null);
```

Skills generates **two separate methods** because the TypeSpec uses `@sharedRoute` with two distinct `op` declarations sharing the same `CreateSkillBody`, rather than a single operation with multiple content types like Videos.

In practice, only multipart file upload works today.

specification/base/typespec/containers/models.tsp
ShivangiReja: Yes, it is used [here](https://github.com/ShivangiReja/openai-dotnet/blob/171a6e2141efc3c89987ef71b90323264b851694/specification/base/typespec/skills/models.tsp#L8).

api/OpenAI.net10.0.cs
ShivangiReja: 1. Renaming to `UpdateSkill` could mislead users into thinking they can update other skill properties (like name or description), when the body only accepts { default_version: string }. Since the API is currently experimental, we can keep the explicit name for now. If the endpoint evolves to support additional fields, we can broaden the name at that point.

2. For `@multipartBody` operations, the generator reorders to put content before contentType. For regular `@body` operations, it preserves the TypeSpec declaration order. This is a generator behavior, multipart operations have special handling that moves the body parameter ahead of the content-type header in the generated signature.

api/OpenAI.net10.0.cs
joseharriaga: Thank you for researching this. Given that only multipart works today, I think we should remove `CreateSkillJson` and leave only the multipart implementation with the `contentType` parameter and renamed to simply be `CreateSkill` (like in Videos). What do you think?

api/OpenAI.net10.0.cs
joseharriaga: 1. My concern is that this would be inconsistent with how we have been naming other update methods. For example:
- The `UpdateConversation` method in Conversations can only update the "metadata" property.
- The `UpdateVoiceConsent` method in Audio can only update the "name" property.
I think what is telling is that the HTTP method is `POST /skills/{skill_id}` which does not make any indication of only being able to update the "default_version" property.

2. This ordering inconsistency in the generator sounds like a bug. Could you check with them?

api/OpenAI.net10.0.cs
ShivangiReja: Yes, totally agree!

api/OpenAI.net10.0.cs
ShivangiReja: 1. Renamed
2. Created an issue: https://github.com/microsoft/typespec/issues/10344

joseharriaga (approved):

PR #1078, opened 2026-04-04
Expose unknown response stream updates for downstream streaming exten…

wmmihaa:
…sions

Make UnknownResponseStreamEvent public and expose its Kind value so consumers can safely detect and handle non-modeled stream events without reflection.


jsquire:
Hi @wmmihaa.   Thanks very much for your cotnribution!  As I mentioned on the assocaited issue, generally, we're in agreement that the ability to directly create/extend the unknown types is something that we'd like to provide. The exact design is not something that we've yet locked down. I've asked the owner of the feature to offer guide feedback here and offer thoughts on the issue.

PR #1082, opened 2026-04-06
Consolidate shared properties

### LINKED ISSUE #1069: Rationalize duplicate enums across multiple models and consolidate them
`ImageGenToolCallItemResource` and `ImageGenerationTool` share some duplicate properties that should be consolidated into some new shared types that are generic to image generation. ex. `ImageGenerationQuality`, `ImageGenerationSize`. (see [this comment](https://github.com/openai/openai-dotnet/pull/1060/changes/BASE..b8d040334a5bf943c33d10f619d9125f3c4b101c#r3030153861) for reference)

This problem also exists in WebSearch.

christothes:
fixes #1069


jsquire (approved):

api/OpenAI.net10.0.cs
```diff
@@ -5704,9 +5704,11 @@ public class ImageGenerationToolInputImageMask : IJsonModel<ImageGenerationToolI
     public readonly partial struct ImageGenerationToolQuality : IEquatable<ImageGenerationToolQuality> {
         public ImageGenerationToolQuality(string value);
         public static ImageGenerationToolQuality Auto { get; }
+        public static ImageGenerationToolQuality Hd { get; }
         public static ImageGenerationToolQuality High { get; }
         public static ImageGenerationToolQuality Low { get; }
         public static ImageGenerationToolQuality Medium { get; }
+        public static ImageGenerationToolQuality Standard { get; }
```
joseharriaga: The values added here (`Hd` and `Standard`) are not valid in the image generation tool. These values only work with the `dall-e-2` model, which is not supported by the image generation tool via the Responses API. This model is currently only available when using the Images API directly.

api/OpenAI.net10.0.cs
```diff
@@ -5722,7 +5724,11 @@ public class ImageGenerationToolInputImageMask : IJsonModel<ImageGenerationToolI
     public readonly partial struct ImageGenerationToolSize : IEquatable<ImageGenerationToolSize> {
         public static readonly ImageGenerationToolSize W1024xH1024;
         public static readonly ImageGenerationToolSize W1024xH1536;
+        public static readonly ImageGenerationToolSize W1024xH1792;
         public static readonly ImageGenerationToolSize W1536xH1024;
+        public static readonly ImageGenerationToolSize W1792xH1024;
+        public static readonly ImageGenerationToolSize W256xH256;
+        public static readonly ImageGenerationToolSize W512xH512;
```
joseharriaga: The values added here (`W1024xH1792`, `W1792xH1024`, `W256xH256`, and `W512xH512`) are not valid in the image generation tool. These values only work with the `dall-e-2` and/or `dall-e-3` models, which are not supported by the image generation tool via the Responses API. These models are currently only available when using the Images API directly.

specification/base/typespec/responses/custom/items.tsp
```diff
@@ -239,26 +239,10 @@ model ImageGenToolCallItemResource extends ItemResource {
...
-    | "1024x1536"
-    | "256x256"
-    | "512x512"
-    | "1792x1024"
-    | "1024x1792"
-    | null = "auto";
+  background: ImageGenerationBackground = ImageGenerationBackground.auto;
+  output_format?: ImageGenerationOutputFormat | null = ImageGenerationOutputFormat.png;
+  quality?: ImageGenerationQuality | null = ImageGenerationQuality.auto;
+  size?: ImageGenerationSize | null = ImageGenerationSize.auto;
```
joseharriaga: Output items like `ImageGenToolCallItemResource` rarely have properties with default values. Are these values actually returned by the service? For example, if we create an `ImageGenerationTool` but leave the `Background` as `null`, does the service return `Auto`, or does it return `null`? If the latter, I think we should remove the default values from the TypeSpec here.

specification/base/typespec/responses/custom/items.tsp
```diff
@@ -239,26 +239,10 @@ model ImageGenToolCallItemResource extends ItemResource {
...
-    | "auto"
-    | "1024x1024"
-    | "1536x1024"
-    | "1024x1536"
-    | "256x256"
-    | "512x512"
-    | "1792x1024"
-    | "1024x1792"
-    | null = "auto";
+  background: ImageGenerationBackground = ImageGenerationBackground.auto;
```
joseharriaga: The `Background` property in `ImageGenerationTool` is optional. Is that optionality preserved in the output here (i.e., the service returns this property as `null`)? Or does the service resolve it to a explicit default value when it returns the `ImageGenerationCall` item (i.e., this property is never `null` and the service throws if `null` is sent back to the service)?

specification/base/typespec/responses/custom/items.tsp
```diff
@@ -239,26 +239,10 @@ model ImageGenToolCallItemResource extends ItemResource {
   type: ItemType.image_generation_call;
   status: "in_progress" | "completed" | "generating" | "failed";
   action: ImageGenerationToolAction;
```
joseharriaga: The `Action` property in `ImageGenerationTool` is optional. Is that optionality preserved in the output here (i.e., the service returns this property as `null`)? Or does the service resolve it to a explicit default value when it returns the `ImageGenerationCall` item (i.e., this property is never `null` and the service throws if `null` is sent back to the service)?

src/Generated/Models/Chat/ChatWebSearchOptions.Serialization.cs
```diff
@@ -7,7 +7,7 @@
 using System.Text;
 using System.Text.Json;
 using OpenAI;
-using OpenAI.Internal;
+using OpenAI.Responses;
```
joseharriaga: `OpenAI.Chat` shouldn't take a dependency on `OpenAI.Responses`

joseharriaga (changes requested):

src/Generated/Models/Chat/ChatWebSearchOptions.Serialization.cs
christothes: This is due to WebSearchToolContextSize - we either need to re-duplicate it or put it in the root namespace.

specification/base/typespec/responses/custom/items.tsp
```diff
@@ -238,27 +238,11 @@ model ImageGenToolCallItemParam extends ItemParam {
...
-    | "256x256"
-    | "512x512"
-    | "1792x1024"
-    | "1024x1792"
-    | null = "auto";
+  action?: ImageGenerationToolAction;
+  background?: ImageGenerationBackground;
+  output_format?: ImageGenerationOutputFormat | null;
+  quality?: ImageGenerationQuality | null;
+  size?: ImageGenerationSize | null;
```
joseharriaga, line 245: Looking at the image generation tool, `output_format`, `quality`, and `size` are not nullable. Have you seen the service returning explicit `null`?

src/Custom/Chat/SearchContextSize.cs
```diff
@@ -0,0 +1,7 @@
+using Microsoft.TypeSpec.Generator.Customizations;
+
+namespace OpenAI.Chat
+{
+    [CodeGenType("SearchContextSize")]
+    public readonly partial struct SearchContextSize { }
```
joseharriaga: In `OpenAI.Chat`, everything is prefixed with "Chat". For consistency, rename this one to `ChatWebSearchContextSize`.

joseharriaga (approved):

specification/base/typespec/responses/custom/items.tsp
christothes: I didn't see any change in the response shape - do you have concerns with this?

joseharriaga: I think the public API will remain unchanged since these properties are also optional, but I would have expected the serialization/deserialization code to change.

PR #1105, opened 2026-04-16
Refactor common tsp types

### LINKED ISSUE #1096: Refactor common TypeSpec definitions into feature-area-specific definitions

The definitions under `./specification/base/typespec/common` should be refactored into separate definitions organized by feature area, instead of keeping them grouped in a shared common location.

This would improve ownership and maintainability by making it clearer which models belong to which API surface, and it would reduce coupling across unrelated feature areas.

### Additional context

Today, `./specification/base/typespec/common` mixes definitions that are used by different feature areas. Splitting these into feature-specific definition files/directories would create a cleaner separation of concerns and make future changes easier to reason about.

This issue is specifically about reorganizing the TypeSpec definitions so that shared/common definitions are minimized and feature-area boundaries are clearer.

christothes:
closes #1096


specification/base/typespec/containers/models.tsp
```diff
@@ -140,3 +154,69 @@ union ContainerSkillType {
...
+  ...ContainerCollectionOrderQueryParameter;
+}
+
+alias ContainerFileCollectionOrderQueryParameter = {
+  /**
+   * Sort order by the `created_at` timestamp of the objects. `asc` for ascending order and`desc`
+   * for descending order.
+   */
+  @query order?: ContainerCollectionOrder;
+};
```
ShivangiReja: Should the order type be `ContainerFileCollectionOrder`? The file‑collection alias is pointing at the container (non‑file) union, and `ContainerFileCollectionOrder` declared just below is unused.

specification/base/typespec/audio/models.tsp
```diff
@@ -1,13 +1,45 @@
...
+  type: string;
+}
+
+@error
+model AudioErrorResponse {
+  error: AudioError;
+}
+
+@access(Access.public)
+@usage(Usage.input | Usage.output)
```
joseharriaga: Since this is now referenced by the `CreateSpeechRequest` model, I believe it should be possible to remove the `@access` and `@usage` decorators.

specification/base/typespec/containers/models.tsp
```diff
@@ -140,3 +154,69 @@ union ContainerSkillType {
   "skill_reference",
   "inline",
 }
+
+alias ContainerCollectionOrderQueryParameter = {
```
joseharriaga: 1. We should remove this alias
2. And duplicate the `ContainerCollectionOrder` union to align with the two types of collections that we have in the Containers API: "containers" vs. "container files". In other words, we should have `ContainerCollectionOrder` and `ContainerFileCollectionOrder`.

specification/base/typespec/containers/operations.tsp
```diff
@@ -16,25 +15,38 @@ interface Containers {
...
+   * Sort order by the `created_at` timestamp of the objects. `asc` for ascending order and`desc`
+   * for descending order.
+   */
+  @query order?: ContainerCollectionOrder,
+    /**
+   * A cursor for use in pagination. `after` is an object ID that defines your place in the list.
+   * For instance, if you make a list request and receive 100 objects, ending with obj_foo, your
+   * subsequent call can include after=obj_foo in order to fetch the next page of the list.
+   */
+  @query @continuationToken after?: string,
```
joseharriaga: Can this be replace with `ContainerCollectionOptions`? You might have to spread it: `...ContainerCollectionOptions`.

specification/base/typespec/containers/operations.tsp
```diff
@@ -70,10 +82,23 @@ interface Containers {
...
+   * Sort order by the `created_at` timestamp of the objects. `asc` for ascending order and`desc`
+   * for descending order.
+   */
+  @query order?: ContainerFileCollectionOrder,
+    /**
+   * A cursor for use in pagination. `after` is an object ID that defines your place in the list.
+   * For instance, if you make a list request and receive 100 objects, ending with obj_foo, your
+   * subsequent call can include after=obj_foo in order to fetch the next page of the list.
+   */
+  @query @continuationToken after?: string,
```
joseharriaga: Similar to the above, can this be replace with `ContainerFileCollectionOptions`?

specification/base/typespec/conversations/models.tsp
```diff
@@ -1,8 +1,19 @@
...
+/**Set of 16 key-value pairs that can be attached to an object. This can be
+useful for storing additional information about the object in a structured
+format, and querying for objects via API or the dashboard.
+Keys are strings with a maximum length of 64 characters. Values are strings
+with a maximum length of 512 characters.*/
+#suppress "@azure-tools/typespec-azure-core/no-openapi" "Auto-suppressed warnings non-applicable rules during import."
+@extension("x-oaiTypeLabel", "map")
+model ConversationsMetadata {
+  ...Record<string>;
+}
```
joseharriaga: I think we can remove this level of indirection: We can remove this model and simply replace it with `Record<string>` wherever it is used.

specification/base/typespec/images/models.tsp
```diff
@@ -715,4 +714,18 @@ union CreateImageStreamingResponse {
   @extension("x-ms-sse-terminal-event", true)
   @terminalEvent
   `image_generation.completed`: ImageGenCompletedEvent,
+}
+
+// =====================================================================
+// Type moved from common/models.tsp — used exclusively by the images
+// feature area.
+// =====================================================================
```
joseharriaga: This comment can be removed.

tests/Utility/OpenAITestEnvironment.cs
```diff
@@ -181,6 +181,7 @@ public T GetTestClient<T>(string modelName = default, OpenAIClientOptions option
         OrganizationId = options?.OrganizationId,
         ProjectId = options?.ProjectId,
         UserAgentApplicationId = options?.UserAgentApplicationId,
+        Transport = options?.Transport,
```
joseharriaga: Shivangi has fixed this issue in main.

specification/base/typespec/vector-stores/operations.tsp
```diff
@@ -1,10 +1,19 @@
...
 import "@typespec/openapi";
-import "../common/models.tsp";
 import "./models.tsp";
 using TypeSpec.Http;
 using TypeSpec.OpenAPI;
 namespace OpenAI;
+
+#suppress "@azure-tools/typespec-azure-core/no-enum" "Auto-suppressed warnings non-applicable rules during import."
+#suppress "@azure-tools/typespec-azure-core/documentation-required" "Auto-suppressed warnings non-applicable rules during import."
+enum VectorStoresOrderEnum {
```
joseharriaga: 1. Could you turn this one into a union with a `string` component.
2. Can you move it to the models.tsp.
3. We should split it to align with the two types of collections that we have in the Vector Stores API: "vector stores" and "vector store files". In other words, we need `VectorStoreCollectionOrder` and `VectorStoreFileCollectionOrder`.

specification/base/typespec/vector-stores/models.tsp
```diff
@@ -1,7 +1,31 @@
...
+/**Set of 16 key-value pairs that can be attached to an object. This can be
+useful for storing additional information about the object in a structured
+format, and querying for objects via API or the dashboard.
+Keys are strings with a maximum length of 64 characters. Values are strings
+with a maximum length of 512 characters.*/
+#suppress "@azure-tools/typespec-azure-core/no-openapi" "Auto-suppressed warnings non-applicable rules during import."
+@extension("x-oaiTypeLabel", "map")
+model VectorStoresMetadata {
+  ...Record<string>;
+}
```
joseharriaga: I think we can remove this level of indirection: We can remove this model and simply replace it with `Record<string>` wherever it is used.

specification/base/typespec/skills/models.tsp
```diff
@@ -1,9 +1,31 @@
...
+}
+
+@error
+model SkillsErrorResponse {
+  error: SkillsError;
+}
+
+#suppress "@azure-tools/typespec-azure-core/no-enum" "Auto-suppressed warnings non-applicable rules during import."
+#suppress "@azure-tools/typespec-azure-core/documentation-required" "Auto-suppressed warnings non-applicable rules during import."
+enum SkillsOrderEnum {
```
joseharriaga: 1. Could you turn this one into a union with a `string` component.
2. We should split it to align with the two types of collections that we have in the Skills API: "skills" and "skill versions". In other words, we need `SkillCollectionOrder` and `SkillVersionCollectionOrder`.

specification/base/typespec/fine-tuning/models.tsp
```diff
@@ -529,3 +538,36 @@ model ListFineTuningCheckpointPermissionResponse {
   last_id?: string | null;
   has_more: boolean;
 }
+
+alias FineTuningCheckpointCollectionOrderQueryParameter = {
```
joseharriaga: This alias doesn't seem to be adding a lot of value, so I think we could remove it to have one less level of indirection.

specification/base/typespec/containers/operations.tsp
christothes: No, since the property names of the options type differ from the protocol, these have to remain this way.

specification/base/typespec/vector-stores/operations.tsp
```diff
@@ -1,10 +1,19 @@
...
+  #suppress "@azure-tools/typespec-azure-core/documentation-required" "Auto-suppressed warnings non-applicable rules during import."
+  asc,
+  #suppress "@azure-tools/typespec-azure-core/documentation-required" "Auto-suppressed warnings non-applicable rules during import."
+  desc,
+}
```
ShivangiReja: This isn’t being used. Can we remove it?

src/Custom/Evals/EvaluationClient.Protocol.cs
```diff
@@ -34,45 +34,45 @@ public partial class EvaluationClient
     public virtual ClientResult GetEvaluations(int? limit, string orderBy, string order, string after, RequestOptions options)
     {
         using PipelineMessage message = CreateGetEvalsRequest(after, limit, order, orderBy, options);
-        return ClientResult.FromResponse(Pipeline.ProcessMessage(message, options));
+        return ClientResult.FromResponse(Pipeline.ProcessMessage<EvaluationClient>(message, options));
```
joseharriaga, line 37: What does this do? I don't see it used in generated code. 🤔

joseharriaga (approved):

specification/base/typespec/containers/operations.tsp
joseharriaga: Good call. We should fix this separately. I filed an issue here:
🔗 https://github.com/openai/openai-dotnet/issues/1110

src/Custom/Evals/EvaluationClient.Protocol.cs
christothes: Are you referring to the GetEvauluations method?

joseharriaga: I'm referring to the diff that shows that `<EvaluationClient>` client was added to the call to `ProcessMessage`. In generated methods, the `ProcessMessage` calls don't have the client specified like that, so I'm trying to understand the context, and if one is better than the other, we should file an issue on the generator to include this too.

christothes: Ah - yes, this routes the error generation to the more specific `DeserializeInternal<Foo>ErrorResponse`. I don't know if it's worth incorporating this change into the generator, but I'll file an issue to track this.

christothes: https://github.com/openai/openai-dotnet/issues/1116

PR #1106, opened 2026-04-17
Fix recorded test failures for Responses client and align `GetRealtimeClient` with `GetResponsesClient`

ShivangiReja:
Recorded tests for `ResponsesClient` were failing because the test helpers (`CreateResponsesClientOptions`) only copied the 4 OpenAI-specific properties from `OpenAIClientOptions`, dropping base `ClientPipelineOptions` state including the test framework's recording `Transport`. Same shape of bug existed in `OpenAIClient.GetRealtimeClient()`.

**Changes:**
- `OpenAITestEnvironment` now copies base pipeline properties (`Transport`, `RetryPolicy`, `MessageLoggingPolicy`, `NetworkTimeout`, `ClientLoggingOptions`, `EnableDistributedTracing`) via a new `CopyPipelineOptions` helper so recorded/playback tests work.
- `OpenAIClient.GetRealtimeClient()` now reuses `Pipeline` directly, mirroring `GetResponsesClient()`.
- Removed `RealtimeClientOptions.FromClientOptions` (no longer needed).
- `RealtimeClientSettings.Options` changed to `RealtimeClientOptions` added config-binding ctor on `RealtimeClientOptions`.
- Added reflection-based guard tests in `RealtimeClientOptionsSmokeTests` and `ResponsesSmokeTests` to fail when new `ClientPipelineOptions` properties are added upstream, preventing silent regressions in the test helper.

### DIFF
--- a/src/Custom/Realtime/RealtimeClientOptions.cs
+++ b/src/Custom/Realtime/RealtimeClientOptions.cs
@@ -1,3 +1,4 @@
+using Microsoft.Extensions.Configuration;
 using System;
 using System.ClientModel.Primitives;
 
@@ -70,22 +71,35 @@ public class RealtimeClientOptions : ClientPipelineOptions
         }
     }
 
-    /// <summary>
-    /// Creates a new <see cref="RealtimeClientOptions"/> from the specified <see cref="OpenAIClientOptions"/>.
-    /// </summary>
-    internal static RealtimeClientOptions FromClientOptions(OpenAIClientOptions options)
+    public RealtimeClientOptions()
     {
-        if (options is null)
+    }
+
+    /// <summary>
+    /// Internal constructor for binding from a configuration section.
+    /// Used by ClientSettings classes to bind nested "Options" section.
+    /// </summary>
+    internal RealtimeClientOptions(IConfigurationSection section)
+    {
+        if (Uri.TryCreate(section[nameof(Endpoint)], UriKind.Absolute, out Uri endpoint))
         {
-            return new RealtimeClientOptions();
+            Endpoint = endpoint;
         }
 
-        return new RealtimeClientOptions
+        if (section[nameof(OrganizationId)] is string organizationId)
         {
-            Endpoint = options.Endpoint,
-            OrganizationId = options.OrganizationId,
-            ProjectId = options.ProjectId,
-            UserAgentApplicationId = options.UserAgentApplicationId,
-        };
+            OrganizationId = organizationId;
+        }
+
+        if (section[nameof(ProjectId)] is string projectId)
+        {
+            ProjectId = projectId;
+        }
+
+        if (section[nameof(UserAgentApplicationId)] is string userAgentApplicationId)
+        {
+            UserAgentApplicationId = userAgentApplicationId;
+        }
     }
+
 }
--- a/src/Custom/Realtime/RealtimeClientOptions.cs
+++ b/src/Custom/Realtime/RealtimeClientOptions.cs
@@ -1,6 +1,7 @@
 using Microsoft.Extensions.Configuration;
 using System;
 using System.ClientModel.Primitives;
+using System.Diagnostics.CodeAnalysis;
 
 namespace OpenAI.Realtime;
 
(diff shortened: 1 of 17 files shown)


src/Custom/Realtime/RealtimeClientOptions.cs
```diff
@@ -70,22 +71,35 @@ public string UserAgentApplicationId
...
+    {
+    }
+
     /// <summary>
-    /// Creates a new <see cref="RealtimeClientOptions"/> from the specified <see cref="OpenAIClientOptions"/>.
+    /// Internal constructor for binding from a configuration section.
+    /// Used by ClientSettings classes to bind nested "Options" section.
     /// </summary>
-    internal static RealtimeClientOptions FromClientOptions(OpenAIClientOptions options)
+    internal RealtimeClientOptions(IConfigurationSection section)
```
joseharriaga: Should this constructor forward to the [`ClientPipelineOptions` constructor](https://github.com/Azure/azure-sdk-for-net/blob/1ba2ecbb2defbf5849abe240b299543aab6476db/sdk/core/System.ClientModel/src/Options/ClientPipelineOptions.cs#L44)? 

```csharp
internal RealtimeClientOptions(IConfigurationSection section) : base(section)
```

tests/Responses/ResponsesSmokeTests.cs
```diff
@@ -634,4 +635,33 @@ public void CanSerializeCodeInterpreterCallLogsOutput()
...
+            .Where(p => p.GetSetMethod(nonPublic: false) is not null)
+            .Select(p => p.Name)
+            .OrderBy(n => n)
+            .ToArray();
+
+        Assert.That(actual, Is.EquivalentTo(expected),
+            "ClientPipelineOptions public settable properties changed. Audit " +
+            "OpenAITestEnvironment.CopyPipelineOptions and update this expected list.");
+    }
+
```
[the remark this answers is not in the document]
joseharriaga: You can move a single instance of this test to the ClientTests.cs file to avoid the redundancy.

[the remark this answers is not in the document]
ShivangiReja: I have updated the test to call `CreateResponsesClientOptions` method to verify it properly.

src/Custom/Realtime/RealtimeClientOptions.cs
ShivangiReja: I think yes.  I have updated all 3: `OpenAIClientOptions`, `RealtimeClientOptions`, and `ResponsesClientOptions`.

tests/Utility/OpenAITestEnvironment.cs
```diff
@@ -167,21 +167,46 @@ public T GetTestClient<T>(string modelName = default, OpenAIClientOptions option
...
+        var result = new ResponsesClientOptions
+        {
+            Endpoint = options?.Endpoint,
+            OrganizationId = options?.OrganizationId,
+            ProjectId = options?.ProjectId,
+            UserAgentApplicationId = options?.UserAgentApplicationId,
+        };
+        CopyPipelineOptions(options, result);
+        return result;
+    }
```
joseharriaga, line 174: Looking at this with fresh eyes, I'm thinking the following: We want to eventually create different Client Options for each of our clients, which means that we would have to write one of these methods for each of them too. At this point, would it be better to create separate `GetTestClient` methods that take the correct Client Options parameter?

In other words, instead of a single method like this:

```csharp
public T GetTestClient<T>(string modelName = default, OpenAIClientOptions options = default);
```

I think we might prefer:

```csharp
public ResponsesClient GetTestResponsesClient(ResponsesClientOptions options = default);
public RealtimeClient GetTestRealtimeClient(RealtimeClientOptions options = default);
// etc.
```

The benefits of the above include:

1. No need to clone/convert the `OpenAIClientOptions` to the specific type.
2. No need to test that the clone/conversion works.
3. Makes it clearer which clients take a model override and which ones don't (for example: Responses and Realtime don't).
4. It feels slightly "more correct".

What do you think? We can start by "splitting" only Responses and Realtime for now, and we can continue using the current `GetTestClient` for the others.

ShivangiReja: Great suggestion - done! Added `GetTestResponsesClient`/`GetTestRealtimeClient` (and proxied equivalents), removed both from the generic `GetTestClient<T>`, migrated all callers, and dropped the clone helpers.

joseharriaga: This ended up looking very clean! Nice!

joseharriaga (approved):

PR #1124, opened 2026-04-25
docs: use realtime function-call arguments in sample

### LINKED ISSUE #535: [FEATURE REQ] Please improve the Realtime sample

https://github.com/openai/openai-dotnet/blob/main/examples/Realtime/Example01_AudioFromFileWithToolsAsync.cs

Doesn't really show you how to use tools. It uses a hard coded response instead of the response that the AI actually returned for the tool. It should provide a sample for the real world that would use the AI's tool response output to create the RealtimeItem.

### Additional context

_No response_

Rohan5commit:
## Summary
- Updates the realtime tool-calling sample to use the function arguments returned by the model instead of a hard-coded location.
- Keeps the sample aligned with how a real tool invocation would be wired up.

## Related issue
- Fixes #535

## Guideline alignment
- Single-file example improvement
- No generated files touched

## Validation
- `git diff --check`
- Local `dotnet build` could not be run in this environment because the `dotnet` CLI is unavailable.

### DIFF
--- a/examples/Realtime/Example01_AudioFromFileWithToolsAsync.cs
+++ b/examples/Realtime/Example01_AudioFromFileWithToolsAsync.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.Json;
 using System.Threading.Tasks;
 
 namespace OpenAI.Examples;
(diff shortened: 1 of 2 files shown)


jsquire (commented):
Thank you for your contribution, @Rohan5commit, and your interest in improving the OpenAI developer experience.   I think there's merit in the approach that you're demonstrating here.  I find myself agreeing with the copilot comments and would ask that we address those before moving forward.

I'd also like to pull in @joseharriaga, @ShivangiReja, and @christothes for additional perspective here.

examples/Realtime/Example01_AudioFromFileWithToolsAsync.cs
```diff
@@ -209,9 +210,27 @@ public async Task Example01_AudioFromFileWithToolsAsync()
...
+                            using JsonDocument argumentsJson = JsonDocument.Parse(functionCallItem.FunctionArguments);
+                            if (!argumentsJson.RootElement.TryGetProperty("location", out JsonElement locationElement)
+                                || locationElement.ValueKind != JsonValueKind.String)
+                            {
+                                throw new InvalidOperationException("Tool call is missing a location.");
+                            }
+
+                            string location = locationElement.GetString()
+                                ?? throw new InvalidOperationException("Tool call is missing a location.");
+                            string unit = argumentsJson.RootElement.TryGetProperty("unit", out JsonElement unitElement)
```
jsquire: 🟡 Could we validate `unit` before passing it to the function? The model may return arguments that do not conform to the schema. A non-string value will cause `GetString()` to throw, while an unsupported string would be accepted by the sample.

Something along these lines would keep the validation consistent with `location`:

```csharp
string unit = "celsius";
if (argumentsJson.RootElement.TryGetProperty("unit", out JsonElement unitElement))
{
    if (unitElement.ValueKind != JsonValueKind.String
        || unitElement.GetString() is not string requestedUnit
        || (requestedUnit != "celsius" && requestedUnit != "fahrenheit"))
    {
        throw new InvalidOperationException("Tool call has an invalid unit.");
    }

    unit = requestedUnit;
}
```

---

:robot: jsquire-copilot

jsquire (approved):

PR #1126, opened 2026-04-27
examples: use tool call arguments in the realtime sample

### LINKED ISSUE #535: [FEATURE REQ] Please improve the Realtime sample

Rohan5commit:
## Summary
- parse the model-provided function arguments in the realtime tools example before constructing the tool output item
- use the requested `location` and `unit` values instead of always sending a hard-coded weather lookup

## Related issue
- Closes #535

## Guideline alignment
- keeps the change to a single manually maintained example file with no generated code changes
- makes the sample closer to a real tool-calling flow without changing library behavior

## Validation
- ran `git diff --check`
- did not run a local `dotnet` build because the `dotnet` CLI is unavailable in this environment

### DIFF
--- a/examples/Realtime/Example01_AudioFromFileWithToolsAsync.cs
+++ b/examples/Realtime/Example01_AudioFromFileWithToolsAsync.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.Json;
 using System.Threading.Tasks;
 
 namespace OpenAI.Examples;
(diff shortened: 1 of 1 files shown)


jsquire (changes requested):
This is a duplicate of #1124.   Closing this one out; please focus efforts there.

PR #1129, opened 2026-04-28
Split Responses

### LINKED ISSUE #1112: Reorganize src into OpenAI and OpenAI.Responses directories as a non-functional prep step

We want to preemptively reorganize the source layout to match the eventual split between the main OpenAI library and a future OpenAI.Responses library, without introducing any functional changes yet.

Under src, create two new directories:
- OpenAI
- OpenAI.Responses

For this step, keep a single project file at OpenAI.csproj. The helpers under Utility should remain unchanged.

The target structure is:

```
src
|-  OpenAI
   |-  Custom
       |-  Assistants
       |-  Audio
       |-  Batch
       |-  ...
   |-  Generated
       |-  Assistants
       |-  Audio
       |-  Batch
       |-  ...
|-  OpenAI.Responses
   |-  Custom
   |-  Generated
       |- Models
       |- ResponsesClient.cs
       |- ResponsesClient.RestClient.cs
|- Utility
|- OpenAI.csproj
```

This is intended to be a cosmetic repo-structure change only. No namespace, API, packaging, or runtime behavior changes should be introduced in this issue.

### Additional context

The key constraint is that generated files must not be moved manually as a one-off file operation. Because these files are produced by code generation, their output paths need to be changed at generation time so they land in the correct directory structure automatically.

For that reason, this issue should use a temporary visitor to redirect generated output into the new layout, similar to ModelDirectoryVisitor.cs. The visitor is not just an implementation detail here; it is the required mechanism for moving Generated files in this step.

Expected scope:
- Move non-Responses handwritten code under src/OpenAI
- Move Responses handwritten code under src/OpenAI.Responses
- Route generated non-Responses output into src/OpenAI/Generated via generator logic
- Route generated Responses output into src/OpenAI.Responses/Generated via generator logic
- Leave Utility unchanged
- Keep build behavior unchanged with the single existing project

The important outcome is that the repo layout changes, but generation remains authoritative for everything under Generated.

christothes:
closes #1112


joseharriaga (approved):

ericstj:
Were you planning to add type-forwards for all the types you pushed down?

PR #1133, opened 2026-04-29
Migrate ASP.NET Core samples to OpenAI Responses API

scottaddie:
## What changed
- Migrates the `examples/aspnet-core` sample app from the OpenAI **Chat Completions** API to the OpenAI **Responses** API.
- Uses latest stable `OpenAI` 2.10.0 package.
- Demonstrates the **configuration-driven** `builder.AddResponsesClient(...)` pattern.


examples/aspnet-core/v2.10.0/Program.cs
```diff
@@ -0,0 +1,26 @@
+using System.ClientModel;
+using OpenAI.Responses;
+
+var builder = WebApplication.CreateBuilder(args);
+builder.AddClient<ResponsesClient, ResponsesClientSettings>("Clients:ResponsesClient");
```
m-nash: Should we use the AddResponsesClient extension from OpenAI library?

examples/aspnet-core/v2.10.0/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+`ResponsesClientSettings` is annotated `[Experimental("SCME0002")]`. This warning is suppressed via `<NoWarn>SCME0002</NoWarn>` in the project file.
+
+### Dependency injection usage
+
+```csharp
+app.MapPost("/responses/create",
+    async (ResponsesRequest request, ResponsesClient client, IConfiguration configuration) =>
+{
+    string model = configuration["Clients:ResponsesClient:Model"]
```
m-nash: this is unfortunate, would be nice if the client had a `Model` property

examples/aspnet-core/v2.10.0/Program.cs
scottaddie: That approach does seem a bit nicer. Is it a safe assumption that we'll always have an `IHostApplicationBuilder` extension method to add each client?

m-nash: Yeah that is the plan, if its not there that should be considered a miss/bug.

m-nash (approved):

examples/aspnet-core/AspNetCoreSamples.slnx
```diff
@@ -0,0 +1,9 @@
+<Solution>
+  <Configurations>
+    <BuildType Name="Debug" />
+    <BuildType Name="Release" />
+    <BuildType Name="Unsigned" />
+  </Configurations>
+  <Project Path="v2.10.0/v2.10.0.csproj" />
+  <Project Path="v2.9.1/v2.9.1.csproj" />
```
jsquire: We have the same support path here that we do elsewhere - you're intended to be using the latest package.   I'd rather we didn't provide examples for older packages, as we don't want to imply that we're actively maintaining multiple version lines.

examples/aspnet-core/v2.10.0/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+    async (ResponsesRequest request, ResponsesClient client, IConfiguration configuration) =>
+{
+    string model = configuration["Clients:ResponsesClient:Model"]
+        ?? throw new InvalidOperationException("Model not configured at Clients:ResponsesClient:Model.");
+    ResponseResult response = await client.CreateResponseAsync(model, request.Message);
+    return new ResponsesResponse(response.GetOutputText());
+});
+app.Run();
+
+public class ResponsesRequest(string message)
```
jsquire: Let's convert these to `records`.    KC was on board with doing so.

jsquire (changes requested):

README.md
```diff
@@ -134,11 +134,19 @@ AudioClient whisperClient = client.GetAudioClient("whisper-1");
...
+### OpenAI v2.10.0 and later
 
-```C# Snippet:ReadMe_DependencyInjection_Register
+```C# Snippet:ReadMe_DependencyInjection_Register_New
+builder.AddChatClient("Clients:ChatClient");
+```
+
+> For a complete ASP.NET Core sample project, see the [dependency injection sample](examples/aspnet-core/README.md).
+
+### OpenAI v2.9.1 or earlier
```
scottaddie: @jsquire I removed the <= 2.9.1 sample app. Do you want this README section removed too?

scottaddie: Removed per offline discussion

jsquire (approved):

PR #1146, opened 2026-05-01
Follow-up cleanup to PR #1133

scottaddie:
Follow-up to https://github.com/openai/openai-dotnet/pull/1133.

- Removes the unused `ApplicationBuilder` class from `tests/Snippets/ReadMeSnippets.cs`.
- Introduces code snippets to ensure the sample app's README stays up-to-date.

### DIFF
--- a/examples/aspnet-core/Program.cs
+++ b/examples/aspnet-core/Program.cs
@@ -2,14 +2,9 @@ using OpenAI;
 using OpenAI.Responses;
 
 var builder = WebApplication.CreateBuilder(args);
-
-#region Snippet:Add_Responses_Client
 builder.AddResponsesClient("Clients:ResponsesClient");
-#endregion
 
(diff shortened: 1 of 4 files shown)


examples/aspnet-core/Program.cs
```diff
@@ -2,9 +2,14 @@
 using OpenAI.Responses;
 
 var builder = WebApplication.CreateBuilder(args);
+
+#region Snippet:Add_Responses_Client
```
jsquire: I'm torn on this one.  While I do appreciate the "avoids drift and stale code", it adds visual noise and complexity to an end-to-end example.

Can we push this into an `ExampleSnippets` file instead?   We're kind of in new territory here and don't have a good pattern for this established.

jsquire (approved):

PR #1152, opened 2026-05-01
Standardize pagination operations and collection option types

### LINKED ISSUE #1110: Use shared pagination option models instead of redeclaring pagination parameters in TypeSpec
We created shared models for pagination parameters, but many pagination operations still redeclare those query parameters inline instead of reusing the shared models.

For example, listVectorStores defines limit, order, after, and before directly even though VectorStoreCollectionOptions already represents the same inputs. This creates duplication and makes it easy for the shared models to drift from the REST API shape.

There is also a naming issue in some shared models. Parameters like pageSizeLimit, afterId, and beforeId should use the REST names limit, after, and before. Any .NET-specific renames should happen in client.tsp, not in the base REST-facing TypeSpec.

**Requested change**

1. Update pagination operations to spread the appropriate shared collection options model instead of redeclaring pagination parameters inline.
2. Rename shared pagination fields to match the REST API contract.
3. Keep any client-specific renames in client.tsp only.
4. Apply this consistently across all pagination operations, not just vector stores.

**Acceptance criteria**

1. Pagination operations no longer duplicate shared pagination parameter definitions.
2. Shared pagination models use the correct REST parameter names.
3. Client renames are handled only in the client customization layer.
4. Codegen is regenerated after the TypeSpec changes.

christothes:
closes #1110


codegen/generator/src/Visitors/PaginationVisitor.cs
```diff
@@ -18,20 +18,24 @@ namespace OpenAILibraryPlugin.Visitors;
...
+    {
+        "after", "afterId", "before", "beforeId", "limit", "pageSizeLimit", "order", "model", "metadata", "filter"
+    };
+    private static readonly Dictionary<string, string> _paramReplacementMap = new(StringComparer.OrdinalIgnoreCase)
     {
         { "after", "AfterId" },
         { "afterId", "AfterId" },
         { "before", "BeforeId" },
+        { "beforeId", "BeforeId" },
         { "limit", "PageSizeLimit" },
```
ShivangiReja: Now that `@@clientName` renames are applied before this visitor runs, the parameter names it sees on both the pagination methods and the *CollectionResult constructor calls are always the SDK-side names (`afterId`, `beforeId`, `pageSizeLimit`). The wire-name aliases (after, before, limit) in `_paginationParamsToReplace` and `_paramReplacementMap` are unreachable. So can we remove `after`, `before` and `limit`?

specification/client/vector-stores.client.tsp
```diff
@@ -43,3 +43,17 @@ using Azure.ClientGenerator.Core;
...
+
+@@clientName(VectorStoreFileCollectionOptions.limit, "PageSizeLimit");
+@@clientName(VectorStoreFileCollectionOptions.after, "afterId");
+@@clientName(VectorStoreFileCollectionOptions.before, "BeforeId");
+@@clientName(VectorStoreFileCollectionOptions.filter, "Filter");
+
+@@clientName(VectorStoreFileCollectionOptions.limit, "PageSizeLimit");
+@@clientName(VectorStoreFileCollectionOptions.after, "afterId");
+@@clientName(VectorStoreFileCollectionOptions.before, "BeforeId");
+@@clientName(VectorStoreFileCollectionOptions.filter, "Filter");
```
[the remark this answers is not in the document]
ShivangiReja: Yes, we need to remove the duplicates.

codegen/generator/src/Visitors/PaginationVisitor.cs
christothes: Yes, if we got them all we should be able to!

specification/client/vector-stores.client.tsp
```diff
@@ -43,3 +43,12 @@ using Azure.ClientGenerator.Core;
...
 @@clientLocation(searchVectorStore, "VectorStores");
+
+@@clientName(VectorStoreCollectionOptions.limit, "PageSizeLimit");
+@@clientName(VectorStoreCollectionOptions.after, "afterId");
+@@clientName(VectorStoreCollectionOptions.before, "BeforeId");
+
+@@clientName(VectorStoreFileCollectionOptions.limit, "PageSizeLimit");
+@@clientName(VectorStoreFileCollectionOptions.after, "afterId");
+@@clientName(VectorStoreFileCollectionOptions.before, "BeforeId");
+@@clientName(VectorStoreFileCollectionOptions.filter, "Filter");
```
ShivangiReja: nit: looks like we are not renaming filter. Can we remove it?

codegen/generator/src/Visitors/PaginationVisitor.cs
```diff
@@ -18,20 +18,21 @@ namespace OpenAILibraryPlugin.Visitors;
 public class PaginationVisitor : ScmLibraryVisitor
 {
 
-    private static readonly string[] _paginationParamsToReplace = ["after", "afterId", "before", "limit", "pageSizeLimit", "order", "model", "metadata", "filter"];
-    private static readonly Dictionary<string, string> _paramReplacementMap = new()
+    private static readonly HashSet<string> _paginationParamsToReplace = new(StringComparer.OrdinalIgnoreCase)
+    {
+        "after", "afterId", "before", "beforeId", "limit", "pageSizeLimit", "order", "model", "metadata", "filter"
+    };
```
ShivangiReja, line 24: nit: can we remove `after`, `before` and `limit` from `_paginationParamsToReplace` as well?

PR #1065, opened 2026-03-30
Add JsonSchemaSegment support for appsettings.json IntelliSense

m-nash:
## Summary

Ship a \ConfigurationSchema.json\ via the MSBuild \JsonSchemaSegment\ feature so that VS 2022 17.9+ provides IntelliSense and validation when configuring OpenAI clients via `appsettings.json\.

Closes #1023

## What this enables

When a customer references the OpenAI NuGet package and configures clients via `appsettings.json\, they get:

- **Client name suggestions** under the \Clients\ section (e.g. \ChatClient\, \AudioClient\, \EmbeddingClient\)
- **Per-client model suggestions** (e.g. chat models like \gpt-5.4\, \gpt-4.1\, \o3\; audio models like \gpt-4o-mini-tts\, \whisper-1\; image models like \gpt-image-1.5\)
- **Options IntelliSense** for \Endpoint\, \OrganizationId\, \ProjectId\, \UserAgentApplicationId\ — composed with System.ClientModel base options (\NetworkTimeout\, \EnableDistributedTracing\, \ClientLoggingOptions\)
- **Credential configuration** inherited from System.ClientModel's schema

## Changes

| File | Description |
|------|-------------|
| \schema/ConfigurationSchema.json\ | JSON Schema fragment with per-client model definitions and OpenAI-specific options |
| \ng/OpenAI.NuGet.targets\ | Registers the \JsonSchemaSegment\ for NuGet consumers |
| \Directory.Build.targets\ | Auto-packs the schema and targets into the NuGet package (conditioned on OpenAI project) |
| \Directory.Packages.props\ | Bumps System.ClientModel to 1.10.0 (ships its own ConfigurationSchema.json with credential/options definitions) |

## NuGet package layout

- \ConfigurationSchema.json\ at the package root
- `buildTransitive/netstandard2.0/OpenAI.targets\ for consumer-side schema registration

## Notes

- Model lists are per-client (chat, audio, embedding, image, moderation, responses) with descriptions noting they are not exhaustive
- VS sorts enum suggestions alphabetically (no schema-level override available)
- OpenAI client options compose with SCM base options via `allOf`n

PR #1158, opened 2026-05-05
Split ConfigurationSchema.json for the OpenAI.Responses package

ShivangiReja:
## Summary

Following the split of `OpenAI.Responses` into its own NuGet package, this PR gives `OpenAI.Responses` its own `ConfigurationSchema.json` (and matching `buildTransitive` targets) instead of relying on the schema shipped in the `OpenAI` package.

Previously, the `OpenAI.Responses` nupkg did not include `ConfigurationSchema.json` or a `buildTransitive/*.targets` file at all, meaning a consumer who installed `OpenAI.Responses` standalone would not get `appsettings.*.json` IntelliSense / validation for `ResponsesClient`. This PR fixes that by adding a Responses-scoped schema, packaged the same way as the existing one in the `OpenAI` package (introduced in #1065).

## Changes

- **Added `schema/Responses/ConfigurationSchema.json`** — Responses-only schema containing `Clients.ResponsesClient` and a dedicated `responsesClientOptions` definition (which mirrors the shape of [`ResponsesClientOptions`](OpenAI.Responses/src/Custom/ResponsesClientOptions.cs)). It is intentionally not a copy of the OpenAI schema — `ResponsesClient` has its own options class, so the schema is self-contained.
- **Removed the `ResponsesClient` block from `schema/ConfigurationSchema.json`** — Responses no longer ships in the `OpenAI` package, so its config entry doesn't belong there.
- **Added `eng/OpenAI.Responses.NuGet.targets`** — registers the schema as a `JsonSchemaSegment` for `appsettings.*.json`, identical in shape to `eng/OpenAI.NuGet.targets`.
- **Updated `Directory.Build.targets`** — added a second `<ItemGroup Condition="'$(MSBuildProjectName)' == 'OpenAI.Responses'">` that packs the Responses schema at the nupkg root and the `.targets` file into `buildTransitive/netstandard2.0/OpenAI.Responses.targets`.

## Verification

Packed both projects locally:

**`OpenAI.Responses.<version>.nupkg`** now contains:
- `ConfigurationSchema.json` (1,855 B — Responses-only)
- `buildTransitive/netstandard2.0/OpenAI.Responses.targets`

**`OpenAI.<version>.nupkg`** still contains its full schema (now 9,451 B — down from 9,768 B since `ResponsesClient` is removed) and `buildTransitive/netstandard2.0/OpenAI.targets`.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -99,4 +99,9 @@
     <None Include="$(RepositoryRoot)eng\OpenAI.NuGet.targets" Pack="true" PackagePath="buildTransitive\netstandard2.0\OpenAI.targets" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(MSBuildProjectName)' == 'OpenAI.Responses'">
+    <None Include="$(RepositoryRoot)schema\Responses\ConfigurationSchema.json" Pack="true" PackagePath="\" Link="schema\ConfigurationSchema.json" />
+    <None Include="$(RepositoryRoot)eng\OpenAI.Responses.NuGet.targets" Pack="true" PackagePath="buildTransitive\netstandard2.0\OpenAI.Responses.targets" />
+  </ItemGroup>
+
 </Project>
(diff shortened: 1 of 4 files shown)


jsquire (approved):
Looks good to me.   We should ask @m-nash to review as well, as he's likely to catch something we've overlooked.

jsquire:
@m-nash:  Would you be so kind as to review?

joseharriaga:
Closing this one for now. We can re-open it later when needed.

PR #1159, opened 2026-05-05
Add dedicated README and CHANGELOG for OpenAI.Responses package

ShivangiReja:
## Summary

Following the split of OpenAI.Responses into its own NuGet package, this PR adds a self-contained README and CHANGELOG so the package no longer ships with the parent OpenAI package's docs.

### Changes

- **README.md** (new): Adapted from the main OpenAI README, scoped to Responses-only content. Examples cover `ResponsesClient` basics, async usage, custom base URL, dependency injection, streaming + reasoning, file search, web search, and Azure OpenAI.
- **CHANGELOG.md** (new): Standard skeleton starting at version `2.11.0 (Unreleased)`.
- **OpenAI.Responses.csproj**: Updated `PackageReadmeFile` / changelog paths to point at the new package-local files instead of the repo-root ones.
- **ReadMeSnippets.cs**: Added test methods/regions for the new snippet markers used by the Responses README.

  Existing Responses snippets (`ReadMe_ResponsesStreaming`, `ReadMe_ResponsesFileSearch`, `ReadMe_ResponsesWebSearch`, `ReadMe_AzureOpenAI`) are reused.

### Validation

- Update-Snippets.ps1 runs cleanly and populates all markers in the new README.
- `dotnet build tests/OpenAI.Tests.csproj` succeeds with 0 warnings / 0 errors.
- `dotnet pack` produces an OpenAI.Responses nupkg containing the new README and CHANGELOG.

### DIFF
--- a/OpenAI.Responses/README.md
+++ b/OpenAI.Responses/README.md
@@ -1,0 +1,290 @@
+# OpenAI.Responses .NET API library
+
+[![NuGet stable version](https://img.shields.io/nuget/v/openai.responses.svg)](https://www.nuget.org/packages/OpenAI.Responses)
+
+The `OpenAI.Responses` .NET library provides convenient access to the [OpenAI Responses API](https://platform.openai.com/docs/api-reference/responses) from .NET applications.
+
+It is generated from the [OpenAPI specification](https://github.com/openai/openai-openapi) in collaboration with Microsoft.
+
+## Table of Contents
+
+- [Getting started](#getting-started)
+  - [Prerequisites](#prerequisites)
+  - [Install the NuGet package](#install-the-nuget-package)
+- [Using the client library](#using-the-client-library)
+  - [Using the async API](#using-the-async-api)
+  - [Using a custom base URL and API key](#using-a-custom-base-url-and-api-key)
+- [How to use dependency injection](#how-to-use-dependency-injection)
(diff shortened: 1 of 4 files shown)


OpenAI.Responses/README.md
```diff
@@ -0,0 +1,290 @@
...
+The OpenAI .NET library has been designed to support mocking, providing key features such as:
+
+- Client methods made virtual to allow overriding.
+- Model factories to assist in instantiating API output models that lack public constructors.
+
+To support mocking output models such as `ResponseResult` (which lacks public constructors), use the `OpenAIResponsesModelFactory` static class to build instances for return values from your mocks.
+
+### Automatically retrying errors
+
+By default, the `ResponsesClient` will automatically retry the following errors up to three additional times using exponential backoff:
```
jsquire, line 279: I'm not sure that we want the explicit retry conditions here; that'll be difficult to keep accurate over time.   It feels like we should consider extending this with a snippet that demonstrates configuring the retry policy, so that folks know how to tune to their needs.

This would also be a great area to demonstrate how to override the network timeout, since that comes up with respect to "my queries are taking up to 10 minutes because I'm using high reasoning."

joseharriaga:
Closing this one for now. We can re-open it later when needed.

PR #1166, opened 2026-05-13
Agentic workflow to fix potential errors in weekly generator upgrade

MaiLinhP:
Addressed comments from https://github.com/openai/openai-dotnet/pull/1156 

Making new PR to bypass push restriction to workflow files from local.


.github/workflows/fix-codegen-errors.md
```diff
@@ -0,0 +1,279 @@
...
+### Step 5: Check for changes
+
+```bash
+git status
+git diff --stat
+```
+
+If there are **no changes** (codegen and API export produced no modifications to tracked files),
+exit gracefully without creating a PR — the new generator version may be fully
+backwards-compatible.
```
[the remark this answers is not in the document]
MaiLinhP: @copilot apply changes based on [this feedback](https://github.com/openai/openai-dotnet/pull/1166#discussion_r3243200593)

.github/skills/fixing-codegen-errors/prohibited-namespace.md
```diff
@@ -0,0 +1,42 @@
...
+   If the target `GeneratorStubs.cs` file does not yet exist for an area, create it with the
+   appropriate namespace. Match the namespace and file header of an existing `GeneratorStubs.cs`
+   in a sibling area.
+
+```csharp
+// Internal example
+[CodeGenType("SomeNewInternalType")] internal partial class InternalSomeNewInternalType { }
+
+// Public example
+[CodeGenType("SomeNewPublicType")] public partial class SomeNewPublicType { }
```
joseharriaga, line 28: I am noticing a pattern when we write new skills, which is that we're taking a very deep dependency on the current repo structure. For example, here we have many explicit mentions to specific path references. The problem with this is that, if anything moves around, these path references become stale and need to be updated. Not only does this represent extra work, but it is also very easy to miss stale path references. The biggest risk with this is that stale path references won't cause things to fail explicitly, but they will silently compromise the effectiveness of the skill by confusing the agent.

We should brainstorm ways to simplify this. For example, the `ingesting-spec` skill has a table of file locations:
🔗 https://github.com/openai/openai-dotnet/blob/main/.github/skills/ingesting-spec/file-locations.md
This looks like something that we should make accessible to the agent **always** regardless of the skill that it is using (perhaps via an AGENTS.md file?). And then, we should avoid mentioning specific path references in our skills as much as possible, and instead refer to things in more generic terms. For example, saying something like "these should go under the `Custom` folders" should be enough for the agent to figure out that we're refering to "OpenAI/src/Custom", "OpenAI.Responses/src/Custom", and any other "Custom" folder that we create in the future if we split another part of the library.

What do you think?

MaiLinhP: Fixed!

PR #1184, opened 2026-05-28
fix: serialize realtime truncate audio end as integer

### LINKED ISSUE #1173: [BUG] RealtimeClientCommandConversationItemTruncate serializes audio_end_ms as a floating-point value
### Describe the bug

Package: OpenAI 2.10.0
Type: RealtimeClientCommandConversationItemTruncate

The Realtime API specifies conversation.item.truncate.audio_end_ms as an Integer value in milliseconds.

However, the generated SDK implementation serializes the value using:

    writer.WriteNumberValue(AudioEndTime.TotalMilliseconds);

TimeSpan.TotalMilliseconds is a double. As a result, audio_end_ms can be serialized as a non-integer JSON number, for example:

    "audio_end_ms": 1234.567

This does not match the Realtime API schema.

Expected behavior:
audio_end_ms should be serialized as an integer number of milliseconds.

Possible fix:
Serialize an integral millisecond value, for example:

    writer.WriteNumberValue(AudioEndTime.Ticks / TimeSpan.TicksPerMillisecond);

### Steps to reproduce

1. Call RealtimeSessionClient.TruncateItemAsync() with a non-zero audioDuration.

### Code snippets

```C#

```

### OS

OS-independent

### .NET version

.NET 10

### Library version

2.10.0

kiwigitops:
## Summary
- serialize realtime `conversation.item.truncate` `audio_end_ms` from `TimeSpan` ticks as an integer millisecond value
- add a serialization regression that rejects fractional JSON output for truncate commands

Fixes #1173.

## Testing
- `dotnet test tests/OpenAI.Tests.csproj -f net8.0 --filter FullyQualifiedName~RealtimeSerializationSmokeTests`
- `dotnet test tests/OpenAI.Tests.csproj -f net10.0 --filter FullyQualifiedName~RealtimeSerializationSmokeTests`
- `git diff --check`

### DIFF
--- a/OpenAI/src/Generated/Models/Realtime/RealtimeClientCommandConversationItemTruncate.Serialization.cs
+++ b/OpenAI/src/Generated/Models/Realtime/RealtimeClientCommandConversationItemTruncate.Serialization.cs
@@ -91,7 +91,7 @@ namespace OpenAI.Realtime
             if (!Patch.Contains("$.audio_end_ms"u8))
             {
                 writer.WritePropertyName("audio_end_ms"u8);
-                writer.WriteNumberValue(AudioEndTime.TotalMilliseconds);
+                writer.WriteNumberValue(AudioEndTime.Ticks / TimeSpan.TicksPerMillisecond);
             }
 
             Patch.WriteTo(writer);
--- a/tests/Realtime/RealtimeSerializationSmokeTests.cs
+++ b/tests/Realtime/RealtimeSerializationSmokeTests.cs
@@ -1,0 +1,31 @@
+using NUnit.Framework;
+using OpenAI.Realtime;
+using System;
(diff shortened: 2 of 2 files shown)


jsquire (changes requested):
Hi @kiwigitops. Thanks for your contribution and your interest in improving the OpenAI developer experience. Unfortunately, this is not a change that we're able to accept.

This change modifies generated code and would be overwritten the next time the library is generated. Any changes to address this issue would need to be made upstream.  We're in conversations with the folks who own the generator about an enhancement that would allow  the format of the TypeSpec `duration` type to be influenced for serialization.

In the meantime, one of the OpenAI SDK core team members already has a short-term work around in flight.  Thank you, again, for your time and contribution. I'm going to close this out.

PR #1185, opened 2026-05-28
fix: avoid underscores in multipart boundaries

### LINKED ISSUE #322: UploadFile exception error: boundary
### Service

OpenAI

### Describe the bug

I'm woring on a Unity project and trying to upload image using UploadFile, the image path is valid, but get the error: boundary
Weixin Screenshot_20250113150205

### Steps to reproduce

N/A

### Code snippets

```C#
private void UploadImage(string imagePath, OpenAIFileClient fileClient, AssistantClient assistantClient)
{
    if (!File.Exists(imagePath))
    {
        Console.WriteLine("File not found: " + imagePath);
        return;
    }
    try
    {
        pictureOfUploadingFile = fileClient.UploadFile(imagePath, FileUploadPurpose.Vision);

        if (pictureOfUploadingFile != null && !string.IsNullOrEmpty(pictureOfUploadingFile.Id))
        {
            Console.WriteLine($"File uploaded successfully! File ID: {pictureOfUploadingFile.Id}");
            Console.WriteLine($"File Status: {pictureOfUploadingFile.Status}");
            Console.WriteLine($"Assistant has recieved the image");
            ImageAssistant(pictureOfUploadingFile, assistantClient);
        }
        else
        {
            Console.WriteLine("File upload failed. The returned file object is null or incomplete.");
        }
    }
    catch (Exception ex) 
    {
        Debug.LogError($"An error occurred while uploading the file: {ex.Message}");
    }
}
```

### OS

winOS11

### .NET version

9.0.101

### Library version

2.1.0

kiwigitops:
## Summary
- avoid `_` in generated multipart boundaries for older `System.Net.Http` implementations
- keep the boundary alphabet at 64 characters for the existing mask-based selection
- add a regression that generated boundaries contain no underscores and are accepted by `MultipartFormDataContent`

Fixes #322.

## Testing
- `dotnet test tests/OpenAI.Tests.csproj -f net8.0 --filter FullyQualifiedName~MultipartContentTests`
- `dotnet test tests/OpenAI.Tests.csproj -f net10.0 --filter FullyQualifiedName~MultipartContentTests`
- `git diff --check`

### DIFF
--- a/OpenAI/src/Generated/Internal/MultiPartFormDataBinaryContent.cs
+++ b/OpenAI/src/Generated/Internal/MultiPartFormDataBinaryContent.cs
@@ -17,7 +17,7 @@ namespace OpenAI
     {
         private readonly MultipartFormDataContent _multipartContent;
         private static readonly Random _random = new Random();
-        private static readonly char[] _boundaryValues = "0123456789=ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz".ToCharArray();
+        private static readonly char[] _boundaryValues = "0123456789=ABCDEFGHIJKLMNOPQRSTUVWXYZ?abcdefghijklmnopqrstuvwxyz".ToCharArray();
 
         public MultiPartFormDataBinaryContent()
         {
--- a/tests/Files/MultipartContentTests.cs
+++ b/tests/Files/MultipartContentTests.cs
@@ -1,0 +1,21 @@
+using NUnit.Framework;
+using OpenAI;
+using System.Net.Http;
+
+namespace OpenAI.Tests.Files;
+
+[Category("Smoke")]
+public class MultipartContentTests
+{
+    [Test]
+    public void MultipartBoundaryAvoidsUnderscore()
+    {
+        for (int i = 0; i < 256; i++)
+        {
+            string boundary = MultiPartFormDataBinaryContent.CreateBoundary();
+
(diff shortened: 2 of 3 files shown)


jsquire (changes requested):
Hi @kiwigitops.  Thanks for your contribution and your interest in improving the OpenAI developer experience.  Unfortunately, this is not a change that we're able to accept.

- This change removes a character that is legal in a boundary, per [RFC 1341 (section 7.2.1)](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html).   While we understand that there is an incompatibility in older versions of the Mono runtime, we do not believe the proposed approach is the right path is to address a bug in another product.

- This change modifies generated code and would be overwritten the next time the library is generated.  Any changes to address this issue would need to be made upstream.

- We've been working with the folks who own the generator and `System.ClientModel` library that underpins the OpenAI SDK.  A change that fixes this scenario landed yesterday in [commit d023aa0](https://github.com/Azure/azure-sdk-for-net/commit/d023aa0faa3f947115f089f1b62d7bc89a60289e#diff-9d76fd6ea99dbfad117f653e935793b08632edf434a6ffa7712e8203f7109dbd).   This changes the approach from random generation to using a new GUID, which also decreases the odds of collisions between the boundary and data in the payload.   This will be adopted in the next SCM version.


Thank you, again, for your time and contribution.   I'm going to close this out.

PR #1186, opened 2026-05-28
Use basename when uploading files by path

### LINKED ISSUE #308: UploadFile/Async with filePath sets whole path as filename

kiwigitops:
Fixes #308.

The file path overloads opened the local file and then forwarded the full `filePath` as the multipart `filename` value. This changes those overloads to pass `Path.GetFileName(filePath)` into the stream overloads, so uploads use the basename while still reading the requested path.

Tested:
- `dotnet test tests\OpenAI.Tests.csproj --framework net8.0 --filter Name~UploadFile`


jsquire (approved):
Thank you for your contribution, @kiwigitops, and for your interest in improving the OpenAI developer experience.   There is some duplication of efforts with #951, but that PR stalled for non-technical reasons due to concerns over this being a behavioral breaking change.  Your submission restarted conversation on that topic.  

The implementation is correct according to RFC 2138, which in section 2.3 calls out that the receiver should "not respect any directory path information that is present in the filename parameter."  That said, my interpretation is that the RFC clearly intends filename only.   

Since the client cannot influence the OpenAI service behavior and force it to ignore the path and the path adds no value, this is a change that should be made and not one that we should view as breaking.

I've updated the associated recorded test and recordings so that the suite now passes fully end-to-end for Files.   I'll get this merged in.

Thank you, again, for your submission.

PR #1188, opened 2026-05-29
Serialize AudioEndTime and AudioStartTime as integer to match REST API

MaiLinhP:
Problem
Addresses https://github.com/openai/openai-dotnet/issues/1173 - Problem
`RealtimeClientCommandConversationItemTruncate` serializes audio_end_ms using `TimeSpan.TotalMilliseconds`, which returns a double. When the `TimeSpan` has sub-millisecond precision (e.g. from tick-level arithmetic), the JSON output contains a fractional value like 1234.5678 instead of the integer 1234 that the REST API expects.

Fix
Add a custom serialization hook via [CodeGenSerialization] that writes audio_end_ms as `(long)(AudioEndTime.Ticks / TimeSpan.TicksPerMillisecond)`, truncating to whole milliseconds.

This is a temporary workaround until the upstream generator bug [microsoft/typespec#10831](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is resolved.


OpenAI/src/Custom/Realtime/GA/ClientCommand/RealtimeClientCommandConversationItemTruncate.Serialization.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace OpenAI.Realtime;
+
+// CUSTOM: Temporary fix to serialize AudioEndTime as an integer number of milliseconds instead of a double to match the REST API.
+// Remove once generator bug https://github.com/microsoft/typespec/issues/10831 is addressed
+[CodeGenSerialization(nameof(AudioEndTime), SerializationValueHook = nameof(SerializeAudioEndTimeValue))]
+public partial class RealtimeClientCommandConversationItemTruncate
+{
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private void SerializeAudioEndTimeValue(Utf8JsonWriter writer, ModelReaderWriterOptions options)
+        => writer.WriteNumberValue((long)(AudioEndTime.Ticks / TimeSpan.TicksPerMillisecond));
```
jsquire: Why not just do `Math.Truncate(AudioEndTime.TotalMilliseconds)`?   If we're really worried about dropping the fractional bits, then `Math.Ceiling(AudioEndTime.TotalMilliseconds)` would ensure that we bump up.

jsquire (dismissed):

tests/Realtime/RealtimeDurationSerializationTests.cs
```diff
@@ -0,0 +1,32 @@
...
+using System;
+using System.ClientModel.Primitives;
+using System.Text.Json;
+
+namespace OpenAI.Tests.Realtime;
+
+#pragma warning disable OPENAI002
+
+[Category("Smoke")]
+public class RealtimeDurationSerializationTests
```
joseharriaga: nit: Could you do the following?
1. Rename the existing `RealtimeClientOptionsSmokeTests` class to `RealtimeUnitTests` (rename the file too).
2. Move this test over there instead of having it in its own individual file.

OpenAI/src/Custom/Realtime/GA/ClientCommand/RealtimeClientCommandConversationItemTruncate.Serialization.cs
joseharriaga: I think `Math.Truncate` sounds good.

joseharriaga (approved):

PR #1192, opened 2026-06-02
Add multiple reference image support to image edits

### LINKED ISSUE #432: There doesn't seem to be a way to pass multiple reference images to a single generation
### Confirm this is a feature request for the .NET library and not the underlying OpenAI API

- [x] This is a feature request for the .NET library

The [official python API example](url) has an example where they reference multiple images (body lotion, bath bomb, etc).

This seems not possible yet with the .NET SDK, as we seem to only be able to pass in a single image via `imageFilePath`

### Additional context

_No response_

adityasingh2400:
## Summary

The image edit endpoint accepts an array of reference images for the GPT image models, sent as repeated `image[]` multipart fields. The TypeSpec model already describes this with `image: HttpPart<bytes | bytes[]>`, and the platform documentation shows the multi-image curl form. However `ImageClient` only exposed single-image edit overloads, so there was no supported way to pass more than one reference image from .NET. This is the gap reported in #432.

This change adds convenience overloads on `ImageClient` that accept multiple reference images:

- `GenerateImageEdits(IEnumerable<Stream> images, IEnumerable<string> imageFilenames, string prompt, ImageEditOptions options = null, CancellationToken cancellationToken = default)`
- `GenerateImageEdits(IEnumerable<string> imageFilePaths, string prompt, ImageEditOptions options = null, CancellationToken cancellationToken = default)`
- the matching `...Async` variants

## Root cause and fix

The hand-written `ImageEditOptions.ToMultipartContent` always wrote a single part named `image`. I added a sibling overload that writes one part per image under the `image[]` field name, which is what the service expects when several images are provided. The existing single-image path is unchanged and still uses the `image` field, so `dall-e-2` and current callers are unaffected. The shared option fields (prompt, model, size, quality, and so on) were factored into a small private helper that both overloads call, to avoid duplicating that block.

The new client overloads validate that the image and filename collections are non-null, non-empty, and the same length before sending, and the file-path overload opens and disposes the file streams deterministically.

All of the changes are in hand-written customization code under `OpenAI/src/Custom/Images`. No generated files were modified. The public API listings under `api/` were updated to include the new overloads.

## Verification

- Built `OpenAI/src/OpenAI.csproj` across netstandard2.0, net8.0, and net10.0 with no warnings or errors.
- Added tests to `ImagesMockTests` and ran them: a serialization test that captures the outgoing multipart body and asserts both reference images are emitted under `image[]`, deserialization tests for the stream and file-path overloads, an argument-validation test, and a cancellation-token test. All pass.
- Ran `dotnet format` on the changed files with no further changes needed.

Fixes #432


api/OpenAI.net10.0.cs
```diff
@@ -3338,11 +3338,15 @@ public class ImageClient {
...
         public virtual ClientResult<GeneratedImageCollection> GenerateImageEdits(string imageFilePath, string prompt, int imageCount, ImageEditOptions options = null);
         public virtual ClientResult<GeneratedImageCollection> GenerateImageEdits(string imageFilePath, string prompt, string maskFilePath, int imageCount, ImageEditOptions options = null);
         public virtual Task<ClientResult> GenerateImageEditsAsync(BinaryContent content, string contentType, RequestOptions options = null);
+        public virtual Task<ClientResult<GeneratedImageCollection>> GenerateImageEditsAsync(IEnumerable<Stream> images, IEnumerable<string> imageFilenames, string prompt, ImageEditOptions options = null, CancellationToken cancellationToken = default);
+        public virtual Task<ClientResult<GeneratedImageCollection>> GenerateImageEditsAsync(IEnumerable<string> imageFilePaths, string prompt, ImageEditOptions options = null, CancellationToken cancellationToken = default);
```
joseharriaga: Since `ImageClient` is already stable, please add the `[Experimental("OPENAI001")]` attribute to the four new methods.

OpenAI/src/Custom/Images/ImageClient.cs
```diff
@@ -866,6 +977,81 @@ private void CreateImageEditOptions(Stream image, string imageFilename, string p
...
+            }
+
+            if (string.IsNullOrEmpty(filenameList[i]))
+            {
+                throw new ArgumentException("The collection of image filenames cannot contain a null or empty entry.", nameof(imageFilenames));
+            }
+        }
+
+        imageFilenameList = filenameList;
+        return imageList;
```
joseharriaga: I don't think this method needs to return anything nor have `out` parameters.

api/OpenAI.net10.0.cs
adityasingh2400: Done. All four new methods now carry `[Experimental("OPENAI001")]`, and the regenerated API listings show the attribute on each of them.

OpenAI/src/Custom/Images/ImageClient.cs
adityasingh2400: Agreed, that was doing too much. `ValidateImages` is now `void` and takes the already-materialized lists, so it only validates. The callers do the null checks and the `ToList()` themselves, which also makes it obvious that the collections are enumerated exactly once. I dropped the `out` parameter from `OpenImageFiles` the same way, since the caller can materialize the path list before handing it over.

adityasingh2400:
Rebased onto main and addressed both review points.

The rebase needed real work because the API listings were split by target framework and namespace in #1237 and #1268, so my edits to the old `api/OpenAI.net*.cs` files no longer applied. I dropped those and regenerated with `scripts/Export-Api.ps1`. The regeneration touches only the three `OpenAI.Images.*` listings and adds exactly the four new overloads, each with the `[Experimental("OPENAI001")]` attribute.

Verified locally: the library builds clean with 0 warnings and 0 errors across netstandard2.0, net8.0, and net10.0, and the Images tests pass 64 of 64.

OpenAI/src/Custom/Images/ImageClient.cs
```diff
@@ -678,6 +679,134 @@ public virtual ClientResult<GeneratedImageCollection> GenerateImageEdits(string
...
+        Argument.AssertNotNull(imageFilenames, nameof(imageFilenames));
+        Argument.AssertNotNullOrEmpty(prompt, nameof(prompt));
+
+        List<Stream> imageList = images.ToList();
+        List<string> imageFilenameList = imageFilenames.ToList();
+        ValidateImages(imageList, imageFilenameList);
+
+        options ??= new();
+        options.Prompt = prompt;
+        options.Model = _model;
```
jsquire, line 710: 🔴 Please clear `options.N` when preparing a multi-reference edit. An existing `GenerateImageEditsAsync` overload stores its `imageCount` in the supplied options. If a caller uses those options to request three outputs and then reuses them with this overload, the second request still sends `n=3`, even though it specifies no output count. The existing single-output edit path clears `N` in this situation.

---
> [!NOTE]

tests/Images/ImagesMockTests.cs
```diff
@@ -339,6 +340,115 @@ public void GenerateImageEditsFromStreamWithMaskRespectsTheCancellationToken()
...
+        await client.GenerateImageEditsAsync([firstStream, secondStream], ["first.png", "second.png"], "prompt");
+
+        string[] imageContentDispositions = requestBody
+            .Split(["\r\n", "\n"], StringSplitOptions.None)
+            .Where(line =>
+                line.StartsWith("Content-Disposition: form-data;", StringComparison.Ordinal) &&
+                (line.Contains("name=image[]") || line.Contains("name=\"image[]\"")))
+            .ToArray();
+
+        Assert.That(imageContentDispositions, Has.Length.EqualTo(2));
```
jsquire, line 420: 🟡 Please use distinct, nonempty image payloads and verify that both bodies arrive in order, along with the model field. This test checks the `image[]` headers and filenames, but its empty streams do not verify that both images are transmitted correctly.

---
> [!NOTE]

Issue #1111, opened 2026-04-22
Add dedicated ClientOptions types for each client


joseharriaga:
## Summary

Most clients currently use `OpenAIClientOptions` directly in constructors and `ClientSettings`. We should introduce a dedicated `*ClientOptions` type for each client, following the existing `ResponsesClient` / `ResponsesClientOptions` pattern.

## Motivation

Using a shared `OpenAIClientOptions` across most clients makes the API less consistent and gives each client no clear place for client-specific configuration. We already have a better pattern in `ResponsesClient`, and the rest of the library should align with it.

## Requirements

- Add a dedicated `*ClientOptions` type for each client.
- For `Experimental` clients, replace constructor and settings dependencies on `OpenAIClientOptions` with the dedicated type.
- For non-`Experimental` clients that are already GA, keep existing `OpenAIClientOptions` support side-by-side with the new dedicated type to avoid breaking stable APIs.

## Scope

This should include:

- public constructors
- `ClientSettings.Options`
- configuration binding paths
- any other dependencies that currently tie a client to `OpenAIClientOptions`

## Acceptance criteria

- Every client has a dedicated `*ClientOptions` type.
- Experimental clients no longer depend on `OpenAIClientOptions` as their client-specific options type.
- Stable clients keep existing `OpenAIClientOptions` entry points for compatibility while also supporting the new dedicated options type.
- API shape is consistent with the `ResponsesClient` pattern.

PR #1208, opened 2026-06-17
Add dedicated ClientOptions types for each client

MaiLinhP:
Addresses https://github.com/openai/openai-dotnet/issues/1111

GA Clients: AudioClient, ChatClient, EmbeddingClient, ImageClient, ModerationClient, OpenAIFileClient, OpenAIModelClient - keep both OpenAIClientOptions entry points along side new dedicated ClientOptions type.

Experimental Clients: AssistantClient, BatchClient, ContainerClient, ConversationClient, EvaluationClient, FineTuningClient, GraderClient, SkillClient, VectorStoreClient, VideoClient - fully replaced OpenAIClientOptions with dedicated ClientOptions types.

### DIFF
--- a/OpenAI/src/Custom/Assistants/AssistantClient.cs
+++ b/OpenAI/src/Custom/Assistants/AssistantClient.cs
@@ -32,7 +32,7 @@ public partial class AssistantClient
     /// <summary> Initializes a new instance of <see cref="AssistantClient"/>. </summary>
     /// <param name="apiKey"> The API key to authenticate with the service. </param>
     /// <exception cref="ArgumentNullException"> <paramref name="apiKey"/> is null. </exception>
-    public AssistantClient(string apiKey) : this(new ApiKeyCredential(apiKey), new OpenAIClientOptions())
+    public AssistantClient(string apiKey) : this(new ApiKeyCredential(apiKey), new AssistantClientOptions())
     {
     }
 
@@ -42,7 +42,7 @@ public partial class AssistantClient
     /// <summary> Initializes a new instance of <see cref="AssistantClient"/>. </summary>
     /// <param name="credential"> The <see cref="ApiKeyCredential"/> to authenticate with the service. </param>
     /// <exception cref="ArgumentNullException"> <paramref name="credential"/> is null. </exception>
-    public AssistantClient(ApiKeyCredential credential) : this(credential, new OpenAIClientOptions())
+    public AssistantClient(ApiKeyCredential credential) : this(credential, new AssistantClientOptions())
     {
     }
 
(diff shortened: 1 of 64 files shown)


OpenAI/src/Custom/Assistants/AssistantClient.cs
```diff
@@ -53,15 +53,15 @@ public partial class AssistantClient
...
+    public AssistantClient(ApiKeyCredential credential, AssistantClientOptions options) : this(OpenAIClient.CreateApiKeyAuthenticationPolicy(credential), options)
     {
     }
 
     // CUSTOM: Added as a convenience.
     /// <summary> Initializes a new instance of <see cref="AssistantClient"/>. </summary>
     /// <param name="authenticationPolicy"> The authentication policy used to authenticate with the service. </param>
     /// <exception cref="ArgumentNullException"> <paramref name="authenticationPolicy"/> is null. </exception>
-    public AssistantClient(AuthenticationPolicy authenticationPolicy) : this(authenticationPolicy, new OpenAIClientOptions())
+    public AssistantClient(AuthenticationPolicy authenticationPolicy) : this(authenticationPolicy, new AssistantClientOptions())
```
jsquire, line 64: @joseharriaga:  With Assistants due for retirement, it seems like we should leave this alone, as we wouldn't split it into its own package.    Thoughts?

OpenAI/src/Custom/Assistants/AssistantClient.cs
```diff
@@ -70,16 +70,17 @@ public AssistantClient(ApiKeyCredential credential, OpenAIClientOptions options)
...
-        options ??= new OpenAIClientOptions();
+        options ??= new AssistantClientOptions();
 
-        Pipeline = OpenAIClient.CreatePipeline(authenticationPolicy, options);
-        _endpoint = OpenAIClient.GetEndpoint(options);
-        _messageSubClient = new(Pipeline, options);
-        _runSubClient = new(Pipeline, options);
-        _threadSubClient = new(Pipeline, options);
+        Pipeline = OpenAIClientUtilities.CreatePipeline(authenticationPolicy, options, options.UserAgentApplicationId, options.OrganizationId, options.ProjectId);
+        _endpoint = OpenAIClientUtilities.GetEndpoint(options.Endpoint);
```
jsquire, line 79: Not a fan of having to allocate the shared options to create the sub clients.   Do we actually need them or is there a different path that we could take?

MaiLinhP: Addressed! Switched sub clients to using the dedicated client options.

PR #1219, opened 2026-06-19
Reset release VersionSuffix and clobber re-uploaded assets

achandmsft:
Two small release-pipeline fixes. Tag-derived versioning was split out into a separate, lower-priority PR (#1220).

### 1. Reset `VersionSuffix` (`Directory.Build.props`)
`VersionSuffix` was left as `oidc.test1` from the OIDC publish test. Reset to empty so the next release publishes as stable `2.11.0` instead of a `-oidc.test1` prerelease.

### 2. `--clobber` on release asset upload (`.github/workflows/release.yml`)
`gh release upload` fails if an asset of the same name already exists, which breaks re-runs. `--clobber` makes re-publishing the same release idempotent.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -169,6 +169,7 @@ jobs:
       - name: Upload release asset
         run: gh release upload ${{ github.event.release.tag_name }}
           ${{ github.workspace }}/build-artifacts/packages/*.*nupkg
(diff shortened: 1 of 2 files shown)


achandmsft:
@benbp

joseharriaga (approved):

PR #1226, opened 2026-06-26
Defer streaming request content creation until enumeration

fallintoplace:
## Summary

- snapshot streaming JSON request payloads when the API is called
- create and dispose `BinaryContent` inside the deferred send when enumeration begins
- pass the snapshot through lightweight `SseUpdateCollection` and `AsyncSseUpdateCollection` adapter constructors
- cover chat, assistants, and speech streaming with request-serialization regressions

## Problem

Several streaming APIs created disposable request content before returning a lazy collection. The HTTP send does not begin until that collection is enumerated, so the request body could be disposed before the transport attempted to serialize it.

The affected paths were:

- chat completion streaming
- assistants thread-and-run streaming
- assistants submit-tool-outputs streaming
- speech generation streaming

## Approach

Each affected method serializes its request once to `BinaryData` at call time. This preserves the original call-time snapshot behavior while allowing a fresh `BinaryContent` instance to be created inside the deferred send and disposed only after that send completes.

The new collection constructors are adapters over the existing delegate-based constructors. They keep the collection internals on a single dispatch path while threading `BinaryData` explicitly at call sites.

Multipart transcription streaming is intentionally unchanged because its request content wraps a live stream and has different lifetime requirements.

## Tests

- added sync and async mock coverage for chat, assistants, and speech streaming that serializes request content only when enumeration starts
- added a negative chat regression demonstrating that stream-backed `BinaryContent` created outside a deferred send becomes unusable after disposal
- verified the transport callback ran before parsing the captured chat request
- ran the focused mock suite on .NET 10: 60 passed, 2 existing mode-specific skips

### DIFF
--- a/OpenAI/src/Custom/Assistants/AssistantClient.cs
+++ b/OpenAI/src/Custom/Assistants/AssistantClient.cs
@@ -690,10 +690,13 @@ public partial class AssistantClient
 
         runOptions ??= new();
         runOptions.Stream = true;
-        BinaryContent protocolContent = CreateThreadAndRunProtocolContent(assistantId, threadOptions, runOptions);
-
+        BinaryData protocolData = CreateThreadAndRunProtocolData(assistantId, threadOptions, runOptions);
         return new AsyncSseUpdateCollection<StreamingUpdate>(
-            async () => await CreateThreadAndRunAsync(protocolContent, cancellationToken.ToRequestOptions(streaming: true)).ConfigureAwait(false),
+            async () =>
+            {
+                using BinaryContent protocolContent = BinaryContent.Create(protocolData);
+                return await CreateThreadAndRunAsync(protocolContent, cancellationToken.ToRequestOptions(streaming: true)).ConfigureAwait(false);
+            },
             StreamingUpdate.FromSseItem,
             cancellationToken);
     }
@@ -715,10 +718,13 @@ public partial class AssistantClient
 
         runOptions ??= new();
         runOptions.Stream = true;
-        BinaryContent protocolContent = CreateThreadAndRunProtocolContent(assistantId, threadOptions, runOptions);
-
+        BinaryData protocolData = CreateThreadAndRunProtocolData(assistantId, threadOptions, runOptions);
         return new SseUpdateCollection<StreamingUpdate>(
-            () => CreateThreadAndRun(protocolContent, cancellationToken.ToRequestOptions(streaming: true)),
+            () =>
+            {
+                using BinaryContent protocolContent = BinaryContent.Create(protocolData);
+                return CreateThreadAndRun(protocolContent, cancellationToken.ToRequestOptions(streaming: true));
+            },
             StreamingUpdate.FromSseItem,
             cancellationToken);
     }
@@ -837,12 +843,15 @@ public partial class AssistantClient
     {
         Argument.AssertNotNullOrEmpty(threadId, nameof(threadId));
         Argument.AssertNotNullOrEmpty(runId, nameof(runId));
-
         var submitToolOutputsRunRequest = new InternalSubmitToolOutputsRunRequest(toolOutputs.ToList(), stream: true, null);
-        using BinaryContent content = BinaryContent.Create(submitToolOutputsRunRequest, ModelSerializationExtensions.WireOptions);
+        BinaryData requestData = ModelReaderWriter.Write(submitToolOutputsRunRequest, ModelSerializationExtensions.WireOptions, OpenAIContext.Default);
 
         return new AsyncSseUpdateCollection<StreamingUpdate>(
-            async () => await SubmitToolOutputsToRunAsync(threadId, runId, content, cancellationToken.ToRequestOptions(streaming: true)).ConfigureAwait(false),
+            async () =>
+            {
+                using BinaryContent content = BinaryContent.Create(requestData);
+                return await SubmitToolOutputsToRunAsync(threadId, runId, content, cancellationToken.ToRequestOptions(streaming: true)).ConfigureAwait(false);
+            },
             StreamingUpdate.FromSseItem,
             cancellationToken);
     }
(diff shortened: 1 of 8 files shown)


OpenAI/src/Custom/Assistants/AssistantClient.cs
```diff
@@ -690,10 +690,13 @@ public virtual AsyncCollectionResult<StreamingUpdate> CreateThreadAndRunStreamin
...
         runOptions ??= new();
         runOptions.Stream = true;
-        BinaryContent protocolContent = CreateThreadAndRunProtocolContent(assistantId, threadOptions, runOptions);
-
+        BinaryData protocolData = CreateThreadAndRunProtocolData(assistantId, threadOptions, runOptions);
         return new AsyncSseUpdateCollection<StreamingUpdate>(
-            async () => await CreateThreadAndRunAsync(protocolContent, cancellationToken.ToRequestOptions(streaming: true)).ConfigureAwait(false),
+            async () =>
+            {
+                using BinaryContent protocolContent = BinaryContent.Create(protocolData);
```
jsquire, line 698: We should elide the closure scope and pass `protocolData` into the lambda as an argument for these.   It's a somewhat minor efficiency, but aggregated, it does make a difference.

jsquire (commented):
Thank you for digging into this and putting together a clear, well-structured fix. The root cause analysis in the PR description is accurate and the approach is sound. We really appreciate the contribution.

Before we merge, we'd like to ask for one design adjustment to the infrastructure layer that we think will result in a cleaner, simpler outcome.

**On the `AsyncSseUpdateCollection` / `SseUpdateCollection` changes**

While reviewing, we looked more closely at how the broader codebase handles this same deferred-send pattern. `CreateRunStreaming` and its siblings already use a safe approach, calling `ToBinaryContent()` inside the lambda so the content is created fresh at enumeration time rather than at call time. That pattern avoids the disposal bug without any infrastructure changes.

We would like to take the spirit of your explicit-argument approach (which we like, because it avoids closure capture and makes data flow clear) and express it through an adapter constructor rather than adding new private fields and conditional dispatch to the base class. Something along these lines would do it:

```csharp
public AsyncSseUpdateCollection(
    Func<BinaryData, Task<ClientResult>> sendRequestAsync,
    BinaryData requestData,
    Func<SseItem<byte[]>, IEnumerable<T>> eventDeserializerFunc,
    CancellationToken cancellationToken)
    : this(() => sendRequestAsync(requestData), eventDeserializerFunc, cancellationToken)
{
    Argument.AssertNotNull(sendRequestAsync, nameof(sendRequestAsync));
    Argument.AssertNotNull(requestData, nameof(requestData));
}
```

This achieves the same thing at the call site, a `BinaryData` snapshot at call time, explicit argument threading, no closures, while keeping the base class internals unchanged with no new nullable fields and no conditional in `GetRawPagesAsync`. The same pattern applies to `SseUpdateCollection`, and the call sites themselves would look identical to what you have today.

We think this is the right balance. It is correct, performant (serializes once at call time and defers allocation of `BinaryContent` to enumeration time), and readable without the added weight of a dual-path dispatch.

**On `TranscribeAudioStreaming[Async]`**

While reviewing we noticed that `TranscribeAudioStreaming[Async]` has a similar structure, where `MultiPartFormDataBinaryContent` is created outside the lambda before being captured. It is a different content type (multipart wrapping a live stream) so it is a separate design problem and we are not asking you to address it in this PR. We just wanted to flag it so you are aware. We will track it on our end, but if you would like to take a pass at it here that would be very welcome.

**On `// CUSTOM:` comments**

Could you add a brief `// CUSTOM:` comment on each of the modified streaming methods explaining the intent? Specifically that `BinaryContent` is intentionally deferred to the lambda body to ensure it is not disposed before enumeration begins. This is a non-obvious inversion of the naive pattern and the explanation will help future readers.

tests/Chat/ChatMockTests.cs
```diff
@@ -290,6 +292,44 @@ public async Task CompleteChatStreamingClosesNetworkStream()
...
+            }
+        }
+        else
+        {
+            foreach (StreamingChatCompletionUpdate _ in client.CompleteChatStreaming(s_messages))
+            {
+            }
+        }
+
+        using JsonDocument requestDocument = JsonDocument.Parse(requestBody);
```
jsquire, line 329: Minor nit: `requestBody` is initialized to `null` and only assigned inside the transport callback. If the callback never fires due to a test setup issue, `JsonDocument.Parse(requestBody)` will throw a non-descriptive `ArgumentNullException`. Adding `Assert.That(requestBody, Is.Not.Null, "Transport callback was never invoked")` before the `Parse` call would give a much clearer failure message in that scenario.

Also, would you be willing to add a negative test here as well, one that constructs the collection with a `using BinaryContent` created outside the lambda (the original broken pattern) and confirms it produces an `ObjectDisposedException` or an empty/unusable request body at enumeration time? This would make the regression suite self-documenting so that someone reading the tests later can see both what the bug looked like and why the fix prevents it.

tests/Audio/GenerateSpeechMockTests.cs
```diff
@@ -43,4 +47,52 @@ public void GenerateSpeechStreamingThrowsForUnsupportedModel(string model)
...
+            }
+        }
+        else
+        {
+            foreach (StreamingSpeechUpdate _ in client.GenerateSpeechStreaming("text", GeneratedSpeechVoice.Alloy))
+            {
+            }
+        }
+
+        using JsonDocument requestDocument = JsonDocument.Parse(requestBody);
```
jsquire, line 84: Minor nit: `requestBody` is initialized to `null` and only assigned inside the transport callback. If the callback never fires due to a test setup issue, `JsonDocument.Parse(requestBody)` will throw a non-descriptive `ArgumentNullException`. Adding `Assert.That(requestBody, Is.Not.Null, "Transport callback was never invoked")` before the `Parse` call would give a much clearer failure message in that scenario.

Also, would you be willing to add a negative test here as well, one that constructs the collection with a `using BinaryContent` created outside the lambda (the original broken pattern) and confirms it produces an `ObjectDisposedException` or an empty/unusable request body at enumeration time? This would make the regression suite self-documenting so that someone reading the tests later can see both what the bug looked like and why the fix prevents it.

tests/Assistants/AssistantsMockTests.cs
```diff
@@ -127,4 +130,53 @@ public async Task StreamingRunIgnoresBenignUnknownEvent()
...
+            }
+        }
+        else
+        {
+            foreach (StreamingUpdate _ in client.CreateThreadAndRunStreaming("asst_abc"))
+            {
+            }
+        }
+
+        using JsonDocument requestDocument = JsonDocument.Parse(requestBody);
```
jsquire, line 168: Minor nit: `requestBody` is initialized to `null` and only assigned inside the transport callback. If the callback never fires due to a test setup issue, `JsonDocument.Parse(requestBody)` will throw a non-descriptive `ArgumentNullException`. Adding `Assert.That(requestBody, Is.Not.Null, "Transport callback was never invoked")` before the `Parse` call would give a much clearer failure message in that scenario.

Also, would you be willing to add a negative test here as well, one that constructs the collection with a `using BinaryContent` created outside the lambda (the original broken pattern) and confirms it produces an `ObjectDisposedException` or an empty/unusable request body at enumeration time? This would make the regression suite self-documenting so that someone reading the tests later can see both what the bug looked like and why the fix prevents it.

One additional item: `SubmitToolOutputsToRunStreaming` and its async variant are also fixed in this PR but do not yet have corresponding regression tests. Could you add coverage here, including the negative test, to match the treatment of the other three call sites?

PR #1227, opened 2026-06-26
Implement synchronous realtime receive updates

fallintoplace:
## Summary
- implement the public sync realtime receive APIs instead of throwing `NotImplementedException`
- reuse the existing async websocket receive pipeline through a blocking sync wrapper, matching the current sync patterns used for realtime session start and command send
- add realtime unit coverage with a fake websocket for both raw protocol updates and typed server updates

## Why
The SDK already exposes public sync receive methods on `RealtimeSessionClient`, but today they only fail at runtime with `NotImplementedException`. That is a rough surprise for callers, especially in a client that already offers sync wrappers for other realtime operations.

## Impact
Sync callers can now enumerate both protocol-level and typed realtime updates without dropping down to the async API.

## Validation
- manual code review of the sync wrapper and fake-websocket tests
- added focused unit tests for `ReceiveUpdates(RequestOptions)` and `ReceiveUpdates(CancellationToken)`
- local `dotnet test` was not run here because this machine does not currently have a `dotnet` SDK on PATH

### DIFF
--- a/OpenAI/src/Custom/Realtime/RealtimeSessionClient.Protocol.cs
+++ b/OpenAI/src/Custom/Realtime/RealtimeSessionClient.Protocol.cs
@@ -116,7 +116,35 @@ public partial class RealtimeSessionClient
 
     public virtual IEnumerable<ClientResult> ReceiveUpdates(RequestOptions options)
     {
-        throw new NotImplementedException();
+        lock (_singleReceiveLock)
+        {
+            _receiveCollectionResult ??= new(WebSocket, options?.CancellationToken ?? default);
+        }
+
+        return EnumerateSync();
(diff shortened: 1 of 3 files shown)


jsquire (changes requested):
Hi @fallintoplace.   Thank you for your contribution and your interest in improving the OpenAI developer experience.   I'm not sure that I agree with having a synchronous surface for `Realtime`.    Streaming APIs are generally better expressed as asynchronous only.   

The current stub exists due to an early design that is now outdated and expected to undergo a full redesign in the short term.   As a result, this is not a submission that we'll be moving forward with at this time.

Thank you, again, for your contribution.

PR #1228, opened 2026-06-26
Fix ResponseContentPart kind setter

fallintoplace:
## Summary
- fix `ResponseContentPart.Kind` to use the assigned `value` when updating `InternalType`
- add a smoke regression that exercises the private setter and verifies the backing type is updated

## Notes
- the current `ResponseContentPart` JSON deserialization path appears to set `InternalType` directly through concrete internal content types, so this looks like a latent correctness fix rather than a confirmed live deserialization failure


joseharriaga (approved):

joseharriaga:
Thank you for your contribution, @fallintoplace !

PR #1230, opened 2026-06-27
Fix data URI TryParse contract

fallintoplace:
## Summary
- catch invalid base64 payloads in `DataEncodingHelpers.TryParseDataUri` and return `false` instead of leaking `FormatException`
- add a chat smoke test that deserializes an invalid file data URI and verifies it is rejected with `ArgumentException`

## Notes
- this keeps the existing file-content validation behavior intact while fixing the `Try*` helper contract underneath it
- I intentionally did not change the separate `image_url` fallback behavior in this PR
- I was not able to run the test suite locally because the `dotnet` SDK is not installed in this environment


joseharriaga (approved):

joseharriaga:
Thank you for your contribution, @fallintoplace !

PR #1236, opened 2026-07-01
Fix release publishing user configuration

achandmsft:
Updates the publishing user in the release workflow to fix the failing package publish step.


.github/workflows/release.yml
```diff
@@ -191,7 +191,7 @@ jobs:
         uses: NuGet/login@ebc737b6fc418a6ca0073cf116ec8dc156d8b81e # v1
         id: nuget-login
         with:
-          user: openai-publisher
+          user: bebroder
```
jsquire, line 194: Seems like we should make this not a real human person at some point.

jsquire (approved):

Issue #1232, opened 2026-06-29
[FEATURE REQ] Move model JSON usage setting out of the emitter


MaiLinhP:
### Describe the feature or improvement you are requesting

By default, we want to set JSON usage for multipart form models so that we also add the correct properties for JSON serialization during generation. We currently set that in the emitter, but this should be moved to a visitor to preserve the pattern of having visitors for OpenAI customization.

PR #1243, opened 2026-07-07
Move model JSON usage setting out of the emitter

MaiLinhP:
Addresses https://github.com/openai/openai-dotnet/issues/1232

With the generator update to add InputModelType Update function, we can now move model JSON usage setting from the emitter to visitor. This ensures consistency of keeping all OpenAI customization in visitors.


Trustinbtc999-hue (commented):
[config]()

joseharriaga (approved):

PR #1253, opened 2026-07-23
[CODEOWNERS] Update for team changes

jsquire:
The focus of these changes is to retire Chris and Shivangi and to onboard Jair and Linh.


jsquire:
Jair will be flagged as invalid until he's officially added to the organization, but I'd rather add him now since it causes no issues.

PR #1255, opened 2026-07-24
Fix conversation include query serialization

### LINKED ISSUE #1254: [BUG] Include parameter in GetConversationItemsAsync causes a 400 response
### Describe the bug

Calling ConversationClient.GetConversationItemsAsync with 1 to many items in the `include` parameter causes an exception.
The OpenAI API responds with a 400 - Bad Request:
```bash
System.ClientModel.ClientResultException: HTTP 400 (invalid_request_error: invalid_type)
Parameter: include

Invalid type for 'include': expected an array of one of 'fil...lts', 'web...ces', 'mes...url', 'com...url', 'cod...uts', 'rea...ent', or 'mes...obs', but got a string instead.
   at OpenAI.ClientPipelineExtensions.ProcessMessageAsync(ClientPipeline pipeline, PipelineMessage message, RequestOptions options)
   at OpenAI.Conversations.ConversationClientGetConversationItemsAsyncCollectionResult.GetNextResponseAsync(PipelineMessage message)
   at OpenAI.Conversations.ConversationClientGetConversationItemsAsyncCollectionResult.GetRawPagesAsync()+MoveNext()
   at OpenAI.Conversations.ConversationClientGetConversationItemsAsyncCollectionResult.GetRawPagesAsync()+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
```

### Steps to reproduce

Create a conversation client:
`builder.Services.AddSingleton<ConversationClient>(_ => new ConversationClient(openAiKey));`

Call `GetConversationItemsAsync`
```csharp
conversationClient.GetConversationItemsAsync(conversationId, limit: 50, 
            include: [IncludedConversationItemProperty.MessageInputImageUri, IncludedConversationItemProperty.WebSearchCallActionSources], order: "asc");
```

### Code snippets

```C#

```

### OS

Windows

### .NET version

10

### Library version

2.12.0

hogeheer499-commits:
Fixes #1254.

Conversation item endpoints currently serialize multiple `include` values as one comma-delimited query value. The API then sees a string instead of an array and returns a 400 response.

This updates the Conversations TypeSpec to use exploded query serialization for `include` on list, create, and retrieve operations, then regenerates the client. Regression tests verify that each endpoint sends one query parameter per included property.

Tests:
- `./scripts/Invoke-CodeGen.ps1 -Clean`
- `dotnet test ./tests/OpenAI.Tests.csproj --configuration Release`
- `dotnet test codegen/generator/test/ --configuration Release`
- `dotnet pack --configuration Release`

### DIFF
--- a/OpenAI/src/Generated/ConversationClient.RestClient.cs
+++ b/OpenAI/src/Generated/ConversationClient.RestClient.cs
@@ -37,7 +37,10 @@ namespace OpenAI.Conversations
             }
             if (include != null && !(include is ChangeTrackingList<IncludedConversationItemProperty> changeTrackingList && changeTrackingList.IsUndefined))
             {
-                uri.AppendQueryDelimited("include", include, ",", escape: true);
+                foreach (var @param in include)
+                {
+                    uri.AppendQuery("include", @param.ToString(), true);
+                }
             }
             PipelineMessage message = Pipeline.CreateMessage(uri.ToUri(), "GET", PipelineMessageClassifier200);
             PipelineRequest request = message.Request;
@@ -56,7 +59,10 @@ namespace OpenAI.Conversations
             uri.AppendPath("/items", false);
             if (include != null && !(include is ChangeTrackingList<IncludedConversationItemProperty> changeTrackingList && changeTrackingList.IsUndefined))
             {
-                uri.AppendQueryDelimited("include", include, ",", escape: true);
+                foreach (var @param in include)
+                {
+                    uri.AppendQuery("include", @param.ToString(), true);
+                }
             }
             PipelineMessage message = Pipeline.CreateMessage(uri.ToUri(), "POST", PipelineMessageClassifier200);
             PipelineRequest request = message.Request;
@@ -94,7 +100,10 @@ namespace OpenAI.Conversations
             uri.AppendPath(itemId, true);
             if (include != null && !(include is ChangeTrackingList<IncludedConversationItemProperty> changeTrackingList && changeTrackingList.IsUndefined))
             {
-                uri.AppendQueryDelimited("include", include, ",", escape: true);
+                foreach (var @param in include)
+                {
+                    uri.AppendQuery("include", @param.ToString(), true);
+                }
             }
             PipelineMessage message = Pipeline.CreateMessage(uri.ToUri(), "GET", PipelineMessageClassifier200);
             PipelineRequest request = message.Request;
--- a/OpenAI/src/Generated/ConversationClient.RestClient.cs
+++ b/OpenAI/src/Generated/ConversationClient.RestClient.cs
@@ -39,7 +39,7 @@ namespace OpenAI.Conversations
             {
                 foreach (var @param in include)
                 {
-                    uri.AppendQuery("include", @param.ToString(), true);
+                    uri.AppendQuery("include[]", @param.ToString(), true);
                 }
             }
             PipelineMessage message = Pipeline.CreateMessage(uri.ToUri(), "GET", PipelineMessageClassifier200);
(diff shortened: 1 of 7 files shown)


specification/base/typespec/conversations/operations.tsp
```diff
@@ -31,7 +31,7 @@ op createConversationItems(
   /** Additional fields to include in the response. See the `include`
     * parameter for [listing Conversation items above](https://platform.openai.com/docs/api-reference/conversations/list-items#conversations_list_items-include)
     * for more information. */
-  @query
+  @query(#{ explode: true })
```
joseharriaga: I was looking at other places in the TypeSpec where the `include` query parameter is used, and I noticed that in addition to `explode: true`, we also need `name: "include[]"` in order for the service to correctly interpret this as an array instead of a string. Here's an existing example:

https://github.com/openai/openai-dotnet/blob/383860cbcf7e51de24fb11bf390267be4077dc98/specification/base/typespec/responses/operations.tsp#L41-L42

In other words, the service expects something like this:

```text
include[]=message.input_image.image_url&include[]=web_search_call.action.sources
```

To fully solve the issue, could you please also add `name: "include[]"` in the three different places where you added `explode: true`?

tests/Conversations/ConversationMockTests.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.ClientModel.Primitives;
+using System.Linq;
+
+namespace OpenAI.Tests.Conversations;
+
+#pragma warning disable OPENAI001
+
+[Category("Conversations")]
+[Category("Smoke")]
+public class ConversationMockTests
```
joseharriaga, line 15: In addition to these mock tests, could you also add recorded tests that validate that the service is accepting the query parameter? 

For context, here's more information about our recorded tests:
🔗 https://github.com/openai/openai-dotnet/blob/main/CONTRIBUTING.md#running-tests

joseharriaga (changes requested):

joseharriaga:
Hi, @hogeheer499-commits : Thank you for your contribution and interest in improving the OpenAI library for .NET! Your changes look good, and I just left two small comments to ensure that your fix fully address the root issue.

specification/base/typespec/conversations/operations.tsp
hogeheer499-commits: Updated all three TypeSpec query parameters to use `name: "include[]"` and regenerated the client. The wire format is now `include[]=...` for list, create, and retrieve.

tests/Conversations/ConversationMockTests.cs
hogeheer499-commits: Thanks for the guidance. I've added a recorded test covering the list, create, and retrieve conversation-item endpoints, while keeping the focused mock coverage.

I was able to capture the two session recordings locally after all. Both fixture variants now pass in Playback mode, and the recordings are included in the latest commit.

tests/Conversations/ConversationTests.cs
```diff
@@ -76,6 +84,64 @@ public async Task CanCreateGetUpdateAndDeleteConversation()
         }
     }
 
+    [RecordedTest]
+    public async Task ConversationItemEndpointsAcceptIncludeQuery()
```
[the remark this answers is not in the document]
hogeheer499-commits: Added both Playback recordings in caee1f1. The synchronous and asynchronous fixture variants now pass with automatic recording disabled.

OpenAI/src/Generated/ConversationClient.RestClient.cs
```diff
@@ -37,7 +37,10 @@ internal virtual PipelineMessage CreateGetConversationItemsRequest(string conver
             }
             if (include != null && !(include is ChangeTrackingList<IncludedConversationItemProperty> changeTrackingList && changeTrackingList.IsUndefined))
             {
-                uri.AppendQueryDelimited("include", include, ",", escape: true);
+                foreach (var @param in include)
+                {
+                    uri.AppendQuery("include[]", @param.ToString(), true);
+                }
```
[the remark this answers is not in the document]
joseharriaga: This is an issue in our code generator, and it's not a blocker for this PR.

joseharriaga (approved):

joseharriaga:
Thank you, @hogeheer499-commits. Approved and merged! 🙂

PR #1260, opened 2026-07-29
[Diagnostics] Add Stainless telemetry headers

### LINKED ISSUE #1252: [FEATURE REQ] Add structured SDK platform metadata headers to outgoing requests

Add structured SDK platform metadata headers to every request made by the .NET client, following the same public header contract used by the other official OpenAI SDKs.

The minimum header set should be:

| Header | Proposed .NET value |
|---|---|
| `X-Stainless-Lang` | Stable language identifier, preferably `csharp` |
| `X-Stainless-Package-Version` | The `OpenAI` NuGet package version |
| `X-Stainless-Runtime` | Stable runtime identifier, preferably `dotnet` |
| `X-Stainless-Runtime-Version` | The running .NET version, e.g. from `Environment.Version` |

For full parity with the other official SDKs, also include:

| Header | Proposed .NET value |
|---|---|
| `X-Stainless-OS` | A normalized OS family such as `Windows`, `Linux`, `MacOS`, or `Other:<value>` |
| `X-Stainless-Arch` | A normalized process architecture such as `x64`, `x86`, `arm64`, `arm`, or `other:<value>` |

These should be default, non-secret request headers. They must not contain usernames, machine names, paths, host identifiers, project identifiers, or other user-specific data.

A natural integration point appears to be `OpenAIClientUtilities.CreateAddCustomHeadersPolicy`, alongside the existing `User-Agent`, organization, and project header handling. The implementation should be shared by all service clients that use the standard client pipeline.

#### Expected behavior

- Add the platform metadata consistently to all outgoing OpenAI requests.
- Add the headers once per request, including retried requests.
- Preserve explicitly supplied caller values instead of overwriting them.
- Keep the existing `User-Agent`, organization, and project behavior unchanged.
- Derive the package version from assembly/package metadata rather than maintaining a second hard-coded version.
- Use stable, documented normalization for OS, architecture, language, and runtime values.
- Ensure all values are non-empty for supported environments; use a stable `unknown`/`other:<value>` fallback when detection is unavailable.
- Keep detection local and deterministic—no network lookup or persistent machine identifier.

#### Tests

Please add request-pipeline tests covering:

1. A default request contains the four required headers.
2. Package version matches the shipped assembly/NuGet package version.
3. Runtime and runtime-version values are non-empty on supported target frameworks.
4. OS and architecture values are normalized if the parity headers are included.
5. Explicit custom header values are respected case-insensitively.
6. The metadata is present across representative service clients that share the pipeline.
7. Retry behavior retains the headers without producing duplicate values.
8. Existing `User-Agent`, organization, project, authentication, and custom-header tests continue to pass.

#### Acceptance criteria

- [ ] All standard .NET client requests announce language, package version, runtime, and runtime version through structured headers.
- [ ] OS and architecture are included, or a follow-up is recorded if they are intentionally deferred.
- [ ] Header values follow one documented normalization contract.
- [ ] Caller overrides are preserved.
- [ ] No user-specific or machine-identifying data is added.
- [ ] Automated tests cover default values, overrides, retries, and shared-pipeline behavior.
- [ ] The change is documented in the release notes/changelog.

### Additional context

The current client sets a package-oriented `User-Agent`, but that string does not provide the same structured language/runtime metadata as the other official SDKs. Structured fields avoid fragile `User-Agent` parsing and bring the .NET client into parity with the public behavior of the Python, JavaScript/TypeScript, Go, Java, and Ruby clients.

Reference for the common header contract: https://www.stainless.com/docs/sdks/configure/client

jsquire:
# Summary

Outgoing requests now carry six structured headers describing the SDK and the platform that issued them, bringing the .NET library to parity with the other official OpenAI SDKs:

| Header | Example |
| --- | --- |
| `X-Stainless-Lang` | `csharp` |
| `X-Stainless-Package-Version` | `2.12.0` |
| `X-Stainless-Runtime` | `dotnet` |
| `X-Stainless-Runtime-Version` | `8.0.11` |
| `X-Stainless-OS` | `Windows` |
| `X-Stainless-Arch` | `x64` |

These restate, in a machine-parseable form, information already present in the `User-Agent`, plus the process CPU architecture, so consumers no longer have to parse the user agent string.

## Why not reuse `System.ClientModel`?

`UserAgentPolicy` exposes only the composed string; the structured inputs behind it (`RuntimeInformationWrapper`, `GenerateUserAgentString`) are `internal`. Parsing that string back apart would reintroduce exactly the fragility this change exists to remove, so the values are derived directly instead.

## Opt-out

Added an opt-out via the `OpenAI.TelemetryDisabled` `AppContext` switch or the `OPENAI_TELEMETRY_DISABLED` environment variable:

```csharp
AppContext.SetSwitch("OpenAI.TelemetryDisabled", true);
```

When set, the library sends neither these headers nor the `User-Agent` it would otherwise add. A `User-Agent` supplied by the caller is still honored, as are the `Authorization`, `OpenAI-Organization`, and `OpenAI-Project` headers. The setting is read once when a pipeline is built, matching how the experimental OpenTelemetry switch behaves.

## Implementation notes

- Headers are applied by the single shared per-call policy that every client is built on, so no client can be constructed without them.
- Values are computed once per process and are never allowed to fail a request; every path degrades to `unknown` rather than throwing.
- Platform detection is `netstandard2.0`-safe. `OSPlatform.FreeBSD` and the mobile/browser platforms do not exist in that reference assembly, so they are created by name. Android and the Apple mobile platforms are evaluated *before* Linux and macOS, because Mono and legacy Xamarin report them as their base platform.
- Free-form platform text is normalized to printable ASCII and bounded in length. **The package version is not** — it is reproduced verbatim so that it always matches the `User-Agent`, and falls back to `unknown` if it could not be transmitted unaltered, rather than being rewritten into a value that would silently disagree.
- Realtime WebSocket handshakes deliberately omit these headers, matching the other official OpenAI SDKs. Regular HTTP requests from `RealtimeClient` include them.

## Recordings

**Existing session recordings are unaffected and none were re-recorded.** A `RemoveHeaderSanitizer` in `OpenAIRecordedTestBase` strips the headers before the proxy matches a playback request and before a recording is written, so existing recordings continue to match and future ones will not capture machine-specific values.

## Testing

80 new mock-based tests covering:

- Normalization of every OS and architecture branch, including the Android-before-Linux and iOS-before-macOS ordering rules
- Never-throw behavior under a runtime that fails on every property
- Header application, case-insensitive caller override, retry de-duplication, and application across multiple client types
- Opt-out precedence and blast radius
- A drift guard asserting `X-Stainless-Package-Version` never diverges from the `User-Agent` version token

| Check | Result |
| --- | --- |
| Library build (`net10.0`, `net8.0`, `netstandard2.0`) | Clean, 0 warnings |
| `PlatformTelemetry` tests (`net10.0`, `net9.0`, `net8.0`) | 80 passed, 0 failed |
| Full Playback suite | 1525 passed, 0 failed, 87 skipped |

## Performance

Measured in Release with the baseline subtracted: **~590 ns**, ~912 bytes allocated, and **183 wire bytes** per request — under 0.01% of a typical call's wall-clock time.

## Compatibility

No public API change and no new dependencies.

### DIFF
--- a/OpenAI/src/Utility/PlatformTelemetry.cs
+++ b/OpenAI/src/Utility/PlatformTelemetry.cs
@@ -61,11 +61,6 @@ internal static class PlatformTelemetry
     private static readonly OSPlatform s_macCatalyst = OSPlatform.Create("MACCATALYST");
     private static readonly OSPlatform s_browser = OSPlatform.Create("BROWSER");
 
-    private static readonly string s_packageVersion = GetPackageVersion(typeof(PlatformTelemetry).Assembly);
-    private static readonly string s_runtimeVersion = GetRuntimeVersion(new TelemetryDetails.RuntimeInformationWrapper());
-    private static readonly string s_operatingSystem = GetOperatingSystem(new TelemetryDetails.RuntimeInformationWrapper());
-    private static readonly string s_architecture = GetArchitecture(new TelemetryDetails.RuntimeInformationWrapper());
-
     /// <summary>
     /// Determines whether the caller has opted out of SDK telemetry, which suppresses both these headers and
     /// the <c>User-Agent</c> header that the library would otherwise add.
@@ -93,11 +88,29 @@ internal static class PlatformTelemetry
         }
 
         SetIfAbsent(request, LangHeaderName, LangValue);
-        SetIfAbsent(request, PackageVersionHeaderName, s_packageVersion);
+        SetIfAbsent(request, PackageVersionHeaderName, Values.PackageVersion);
         SetIfAbsent(request, RuntimeHeaderName, RuntimeValue);
-        SetIfAbsent(request, RuntimeVersionHeaderName, s_runtimeVersion);
-        SetIfAbsent(request, OSHeaderName, s_operatingSystem);
-        SetIfAbsent(request, ArchHeaderName, s_architecture);
+        SetIfAbsent(request, RuntimeVersionHeaderName, Values.RuntimeVersion);
+        SetIfAbsent(request, OSHeaderName, Values.OperatingSystem);
+        SetIfAbsent(request, ArchHeaderName, Values.Architecture);
+    }
+
+    /// <summary>
+    /// The values that are determined once and reused for the lifetime of the process.
+    /// </summary>
+    /// <remarks>
+    /// These are held in a nested type so that they are computed on first use rather than when
+    /// <see cref="PlatformTelemetry"/> is first touched. A process that has opted out calls
+    /// <see cref="IsTelemetryDisabled"/> but never <see cref="ApplyTo"/>, and so never pays to determine them.
+    /// </remarks>
+    private static class Values
+    {
+        private static readonly TelemetryDetails.RuntimeInformationWrapper s_runtimeInformation = new();
+
+        public static readonly string PackageVersion = GetPackageVersion(typeof(PlatformTelemetry).Assembly);
+        public static readonly string RuntimeVersion = GetRuntimeVersion(s_runtimeInformation);
+        public static readonly string OperatingSystem = GetOperatingSystem(s_runtimeInformation);
+        public static readonly string Architecture = GetArchitecture(s_runtimeInformation);
     }
 
     private static void SetIfAbsent(PipelineRequest request, string name, string value)
--- a/OpenAI/src/Utility/PlatformTelemetry.cs
+++ b/OpenAI/src/Utility/PlatformTelemetry.cs
@@ -1,0 +1,322 @@
+using System;
+using System.ClientModel.Primitives;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Text;
+
+#nullable enable
+
+namespace OpenAI;
+
+/// <summary>
+/// Structured, individually-parseable SDK platform metadata that is sent alongside the
+/// <c>User-Agent</c> header so that consumers do not have to parse the user agent string.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Every value is process-invariant, so the set is computed once and cached.
+/// </para>
+/// <para>
+/// Values are derived from the same sources used to build the <c>User-Agent</c> (see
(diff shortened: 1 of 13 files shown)


jsquire:
Fixes #1252

OpenAI/src/Utility/PlatformTelemetry.cs
```diff
@@ -0,0 +1,342 @@
...
+internal static class PlatformTelemetry
+{
+    public const string LangHeaderName = "X-Stainless-Lang";
+    public const string PackageVersionHeaderName = "X-Stainless-Package-Version";
+    public const string RuntimeHeaderName = "X-Stainless-Runtime";
+    public const string RuntimeVersionHeaderName = "X-Stainless-Runtime-Version";
+    public const string OSHeaderName = "X-Stainless-OS";
+    public const string ArchHeaderName = "X-Stainless-Arch";
+
+    public const string TelemetryDisabledSwitchName = "OpenAI.TelemetryDisabled";
```
joseharriaga: Should we rename this switch: `OpenAI.TelemetryDisabled` --> `OpenAI.DisableTelemetry`? From my experience, switch are typically verbs. This would also align with other switch that we have: `OpenAI.Experimental.EnableOpenTelemetry`. Same for the environment variable name: `OPENAI_TELEMETRY_DISABLED` --> `OPENAI_DISABLED_TELEMETRY`. And then, the docs would need to be updated.

joseharriaga (approved):

OpenAI/src/Utility/PlatformTelemetry.cs
jsquire: Let me double-check.  This was a refactor of an existing switch, so the name should have been stable.

jsquire: Turns out this was named based on prior art in Azure.Core.   I think your name makes more sense; upodating.

PR #1263, opened 2026-07-30
Add mutual TLS transport example

jbeckwith-oai:
## Summary

- document API-key + HTTP mTLS using the SDK's existing transport override
- add a runnable .NET example with PowerShell setup for global and EU mTLS endpoints
- add deterministic end-to-end TLS tests for full chains, rejected credentials, API-key auth, request/response handling, and redirect safety

## Validation

- `dotnet test tests/OpenAI.Tests.csproj --framework net10.0 --configuration Release --no-restore`
  - 1,451 passed; 87 skipped
- `dotnet test tests/OpenAI.Tests.csproj --framework net10.0 --configuration Release --no-build --no-restore --filter "TestCategory=MutualTls"`
  - 6 passed
- `dotnet build tests/OpenAI.Tests.csproj --configuration Release --no-restore`
  - net8.0/net9.0/net10.0, 0 warnings
- `dotnet build examples/OpenAI.Examples.csproj --framework net10.0 --configuration Release --no-restore`
  - 0 warnings
- formatting, generated snippets, and `git diff --check`

## Live mTLS validation

Using the dedicated smoke-test project and `~/.openai` fixture kit:

- enrolled direct client certificate: passed
- enrolled leaf + intermediate chain: passed
- unregistered CA certificate: rejected as expected
- chained leaf without intermediate: rejected as expected

### DIFF
--- a/README.md
+++ b/README.md
@@ -991,9 +991,11 @@ Console.WriteLine(response.Value.GetOutputText());
 
 ### Using mutual TLS
 
-The API mTLS beta combines your API key with a client certificate. For how request authentication works, certificate requirements, activation, and supported API operations, see the [OpenAI Mutual TLS Beta Program](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program). To manage uploaded certificates programmatically, see the [Certificates API reference](https://developers.openai.com/api/reference/resources/admin/subresources/organization/subresources/certificates/).
+The API mTLS beta combines your API key with a client certificate. For how request authentication works, certificate requirements, activation, and supported API operations, see the [OpenAI Mutual TLS Beta Program](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program). To upload and activate the issuing CA programmatically, see the [Certificates API reference](https://developers.openai.com/api/reference/resources/admin/subresources/organization/subresources/certificates/).
 
-Configure the client certificate chain on a `SocketsHttpHandler`, then provide its `HttpClient` to the SDK through `HttpClientPipelineTransport`. The PFX/PKCS#12 file must contain exactly one certificate with a private key—the leaf client certificate—along with any intermediate certificates required to build its chain. This example uses `X509CertificateLoader`, available in .NET 9 and later.
+Upload and activate the CA certificate that signs the client leaf; keep the leaf certificate and its private key in the PFX used by the application. Certificate-chain support is available by request and must be enabled for your organization. Without it, the leaf must be signed directly by the uploaded CA. When chain support is enabled, the PFX/PKCS#12 file must contain exactly one certificate with a private key—the leaf—and every intermediate certificate needed to build a chain to the uploaded CA.
+
+Configure that bundle on a `SocketsHttpHandler`, then provide its `HttpClient` to the SDK through `HttpClientPipelineTransport`. This example uses `X509CertificateLoader`, available in .NET 9 and later.
 
 Choose the base URL that matches your data residency:
 
@@ -1034,7 +1036,11 @@ try
     };
     handler.SslOptions.ClientCertificateContext = certificateContext;
 
-    using HttpClient httpClient = new(handler);
+    using HttpClient httpClient = new(handler)
+    {
+        // Let the SDK pipeline enforce OpenAIClientOptions.NetworkTimeout.
+        Timeout = System.Threading.Timeout.InfiniteTimeSpan,
+    };
     OpenAIClientOptions options = new()
     {
         Endpoint = new Uri("https://mtls.api.openai.com/v1"),
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ It is generated from our [OpenAPI specification](https://github.com/openai/opena
 - [How to use assistants with streaming and vision](#how-to-use-assistants-with-streaming-and-vision)
 - [How to work with Azure OpenAI](#how-to-work-with-azure-openai)
 - [Advanced scenarios](#advanced-scenarios)
+  - [Using mutual TLS](#using-mutual-tls)
   - [Using protocol methods](#using-protocol-methods)
   - [Mock a client for testing](#mock-a-client-for-testing)
   - [Automatically retrying errors](#automatically-retrying-errors)
@@ -987,6 +988,79 @@ Console.WriteLine(response.Value.GetOutputText());
 - Drop‑in model switching: Swap "gpt-5-mini" or any other model as long as the Azure model deployment has the same name as the model.
 
 ## Advanced scenarios
+
+### Using mutual TLS
+
+The API mTLS beta combines your API key with a client certificate. For how request authentication works, certificate requirements, activation, and supported API operations, see the [OpenAI Mutual TLS Beta Program](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program). To manage uploaded certificates programmatically, see the [Certificates API reference](https://developers.openai.com/api/reference/resources/admin/subresources/organization/subresources/certificates/).
+
+Configure the client certificate chain on a `SocketsHttpHandler`, then provide its `HttpClient` to the SDK through `HttpClientPipelineTransport`. The PFX/PKCS#12 file must contain exactly one certificate with a private key—the leaf client certificate—along with any intermediate certificates required to build its chain. This example uses `X509CertificateLoader`, available in .NET 9 and later.
+
+Choose the base URL that matches your data residency:
+
+| Scope | Base URL |
+| --- | --- |
+| Global | `https://mtls.api.openai.com/v1` |
+| EU Data Residency | `https://mtls-eu.api.openai.com/v1` |
+
+```C# Snippet:ReadMe_MutualTls
+string pfxPath = Environment.GetEnvironmentVariable("OPENAI_CLIENT_PFX_PATH")
+    ?? throw new InvalidOperationException("OPENAI_CLIENT_PFX_PATH is required.");
+string apiKey = Environment.GetEnvironmentVariable("OPENAI_API_KEY")
+    ?? throw new InvalidOperationException("OPENAI_API_KEY is required.");
+
+X509Certificate2Collection certificates =
+    X509CertificateLoader.LoadPkcs12CollectionFromFile(
+        pfxPath,
+        Environment.GetEnvironmentVariable("OPENAI_CLIENT_PFX_PASSWORD"));
+
+try
+{
+    X509Certificate2 clientCertificate =
+        certificates.Single(certificate => certificate.HasPrivateKey);
+    X509Certificate2Collection intermediateCertificates = new(
+        certificates
+            .Where(certificate => !certificate.HasPrivateKey)
+            .ToArray());
+    SslStreamCertificateContext certificateContext =
+        SslStreamCertificateContext.Create(
+            clientCertificate,
+            intermediateCertificates,
+            offline: true);
+
+    SocketsHttpHandler handler = new()
+    {
+        // Do not automatically follow a redirect with a certificate-bearing handler.
+        AllowAutoRedirect = false,
+    };
+    handler.SslOptions.ClientCertificateContext = certificateContext;
+
+    using HttpClient httpClient = new(handler);
+    OpenAIClientOptions options = new()
+    {
+        Endpoint = new Uri("https://mtls.api.openai.com/v1"),
+        Transport = new HttpClientPipelineTransport(httpClient),
+    };
+
+    OpenAIClient client = new(
+        new ApiKeyCredential(apiKey),
+        options);
+
+    ChatCompletion completion = await client
+        .GetChatClient("gpt-4o-mini")
+        .CompleteChatAsync("Reply with exactly: mTLS request succeeded.");
+}
+finally
+{
+    foreach (X509Certificate2 certificate in certificates)
+    {
+        certificate.Dispose();
+    }
+}
+```
+
+Keep the `HttpClient` alive for as long as the OpenAI client uses it, and create a new handler, transport, and client when rotating the certificate so new TLS connections use the replacement credentials. This transport recipe applies to HTTP clients created from `OpenAIClient`; it does not configure Realtime WebSocket connections.
+
+See the [complete mTLS example](examples/MutualTls/README.md) for setup and a runnable end-to-end request.
 
 ### Using protocol methods
 
--- a/README.md
+++ b/README.md
@@ -1027,6 +1027,7 @@ try
         SslStreamCertificateContext.Create(
             clientCertificate,
             intermediateCertificates,
+            // Build the local chain without revocation network lookups.
             offline: true);
 
     SocketsHttpHandler handler = new()
--- a/examples/MutualTls/Example01_MutualTlsAsync.cs
+++ b/examples/MutualTls/Example01_MutualTlsAsync.cs
@@ -1,0 +1,101 @@
+// Demonstrates API-key + HTTP mTLS with OpenAI by configuring .NET's native
+// SocketsHttpHandler and passing its HttpClient through the SDK's existing
+// transport seam. Keeping certificate handling in the HTTP transport preserves
+// native .NET support for certificate stores, proxies, and credential rotation.
+
+using NUnit.Framework;
+using OpenAI.Chat;
+using System;
+using System.ClientModel;
+using System.ClientModel.Primitives;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading.Tasks;
+
+namespace OpenAI.Examples;
+
+public partial class MutualTlsExamples
+{
+    [Test]
+    [Explicit("Requires an API key and a client certificate enrolled in the OpenAI mTLS beta.")]
+    public async Task Example01_MutualTlsAsync()
+    {
+        string apiKey = GetRequiredEnvironmentVariable("OPENAI_API_KEY");
+        string pfxPath = GetRequiredEnvironmentVariable("OPENAI_CLIENT_PFX_PATH");
+        Uri endpoint = GetEndpoint();
+
+        X509Certificate2Collection certificates =
+            X509CertificateLoader.LoadPkcs12CollectionFromFile(
+                pfxPath,
+                Environment.GetEnvironmentVariable("OPENAI_CLIENT_PFX_PASSWORD"));
+
+        try
+        {
+            X509Certificate2 clientCertificate =
+                certificates.Single(certificate => certificate.HasPrivateKey);
+            X509Certificate2Collection intermediateCertificates = new(
+                certificates
+                    .Where(certificate => !certificate.HasPrivateKey)
+                    .ToArray());
+            SslStreamCertificateContext certificateContext =
+                SslStreamCertificateContext.Create(
+                    clientCertificate,
+                    intermediateCertificates,
+                    offline: true);
+
+            SocketsHttpHandler handler = new()
+            {
+                AllowAutoRedirect = false,
+            };
+            handler.SslOptions.ClientCertificateContext = certificateContext;
+
+            using HttpClient httpClient = new(handler);
+            OpenAIClientOptions options = new()
+            {
+                Endpoint = endpoint,
+                Transport = new HttpClientPipelineTransport(httpClient),
+            };
+
+            OpenAIClient client = new(new ApiKeyCredential(apiKey), options);
+            ChatCompletion completion = await client
+                .GetChatClient("gpt-4o-mini")
+                .CompleteChatAsync("Reply with exactly: mTLS request succeeded.");
+
+            Console.WriteLine(completion.Content[0].Text);
+        }
+        finally
+        {
+            foreach (X509Certificate2 certificate in certificates)
+            {
+                certificate.Dispose();
+            }
+        }
+    }
+
+    private static string GetRequiredEnvironmentVariable(string name)
+    {
+        string value = Environment.GetEnvironmentVariable(name);
+        return !string.IsNullOrWhiteSpace(value)
+            ? value
+            : throw new InvalidOperationException($"{name} is required.");
+    }
+
+    private static Uri GetEndpoint()
+    {
+        // Choose the global endpoint or set OPENAI_BASE_URL to the EU endpoint:
+        // https://mtls-eu.api.openai.com/v1
+        string value = Environment.GetEnvironmentVariable("OPENAI_BASE_URL")
+            ?? "https://mtls.api.openai.com/v1";
+
+        if (!Uri.TryCreate(value, UriKind.Absolute, out Uri endpoint)
+            || endpoint.Scheme != Uri.UriSchemeHttps)
+        {
+            throw new InvalidOperationException(
+                "OPENAI_BASE_URL must be an absolute HTTPS URI.");
+        }
+
+        return endpoint;
+    }
+}
--- a/examples/MutualTls/Example01_MutualTlsAsync.cs
+++ b/examples/MutualTls/Example01_MutualTlsAsync.cs
@@ -46,6 +46,7 @@ public partial class MutualTlsExamples
                 SslStreamCertificateContext.Create(
                     clientCertificate,
                     intermediateCertificates,
+                    // Build the local chain without revocation network lookups.
                     offline: true);
 
             SocketsHttpHandler handler = new()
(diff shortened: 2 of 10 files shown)


examples/MutualTls/Example01_MutualTlsAsync.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    private static Uri GetEndpoint()
+    {
+        // Choose the global endpoint or set OPENAI_BASE_URL to the EU endpoint:
+        // https://mtls-eu.api.openai.com/v1
+        string value = Environment.GetEnvironmentVariable("OPENAI_BASE_URL")
+            ?? "https://mtls.api.openai.com/v1";
+
+        if (!Uri.TryCreate(value, UriKind.Absolute, out Uri endpoint)
+            || endpoint.Scheme != Uri.UriSchemeHttps)
```
HAYDEN-OAI: **[P1] Restrict certificate-bearing requests to the supported mTLS origins.** `GetEndpoint()` accepts every absolute HTTPS origin, so setting `OPENAI_BASE_URL=https://attacker.example/v1` sends the OpenAI bearer token to that server and also presents the client certificate if it requests one. `AllowAutoRedirect = false` protects later redirect hops but does nothing to constrain this initial destination. Since this example only supports the global and EU OpenAI mTLS endpoints, validate the host against `mtls.api.openai.com` / `mtls-eu.api.openai.com` (and reject unexpected user info/ports, or require an explicitly trusted allowlist) before using the certificate-bearing transport.

README.md
```diff
@@ -988,6 +989,79 @@ Console.WriteLine(response.Value.GetOutputText());
...
+            offline: true);
+
+    SocketsHttpHandler handler = new()
+    {
+        // Do not automatically follow a redirect with a certificate-bearing handler.
+        AllowAutoRedirect = false,
+    };
+    handler.SslOptions.ClientCertificateContext = certificateContext;
+
+    using HttpClient httpClient = new(handler);
```
HAYDEN-OAI: **[P2] Preserve the SDK pipeline's configured network timeout.** A newly constructed `HttpClient` retains its 100-second default timeout, while `HttpClientPipelineTransport`'s built-in client deliberately sets `Timeout = Timeout.InfiniteTimeSpan` because the pipeline owns timeout handling. Customers following this recipe who set `OpenAIClientOptions.NetworkTimeout` above 100 seconds will still have slow requests canceled at 100 seconds. Set the custom client's timeout to `System.Threading.Timeout.InfiniteTimeSpan` here and in the runnable example/snippet so the SDK's configured timeout remains authoritative.

examples/MutualTls/README.md
```diff
@@ -0,0 +1,73 @@
+# API-key + HTTP mTLS example
+
+This example configures mTLS on .NET's native `SocketsHttpHandler` and passes the resulting `HttpClient` to the OpenAI SDK through `HttpClientPipelineTransport`. Keeping certificate handling in the HTTP transport preserves .NET's native certificate, proxy, and rotation capabilities without adding an SDK-specific mTLS option.
+
+Before running it, follow the [OpenAI Mutual TLS Beta Program](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program) instructions to understand request authentication and enroll, upload, and activate a client certificate for your organization or project. To manage uploaded certificates programmatically, see the [Certificates API reference](https://developers.openai.com/api/reference/resources/admin/subresources/organization/subresources/certificates/).
```
HAYDEN-OAI: **[P2] Instruct customers to upload the issuing CA, not the client leaf.** OpenAI's [mTLS program requirements](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program) require uploading and activating an X.509 **CA certificate**; the client leaf/private key stay in the PFX and are only presented during the TLS handshake. This sentence instead says to upload and activate a client certificate, which directs users to submit the leaf and fails the CA/basic-constraints requirements. Say to upload/activate the issuing CA (the direct signer, or an enabled chain's trust anchor) and use the corresponding leaf certificate in `OPENAI_CLIENT_PFX_PATH`.

README.md
```diff
@@ -988,6 +989,79 @@ Console.WriteLine(response.Value.GetOutputText());
 
 ## Advanced scenarios
 
+### Using mutual TLS
+
+The API mTLS beta combines your API key with a client certificate. For how request authentication works, certificate requirements, activation, and supported API operations, see the [OpenAI Mutual TLS Beta Program](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program). To manage uploaded certificates programmatically, see the [Certificates API reference](https://developers.openai.com/api/reference/resources/admin/subresources/organization/subresources/certificates/).
+
+Configure the client certificate chain on a `SocketsHttpHandler`, then provide its `HttpClient` to the SDK through `HttpClientPipelineTransport`. The PFX/PKCS#12 file must contain exactly one certificate with a private key—the leaf client certificate—along with any intermediate certificates required to build its chain. This example uses `X509CertificateLoader`, available in .NET 9 and later.
```
HAYDEN-OAI: **[P2] Document that intermediate-chain support requires separate enablement.** The linked [current mTLS documentation](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program) says certificate-chain support is available **by request** and must be enabled for the organization; otherwise the client certificate must be directly signed by an uploaded, activated CA. Presenting this intermediate-chain recipe as universally supported therefore produces handshake failures for customers who enabled standard mTLS but do not have the additional chain feature. Call out that prerequisite prominently in both READMEs and explain the directly-signed fallback.

tests/MutualTls/MutualTlsExamplesTests.cs
```diff
@@ -0,0 +1,840 @@
...
+                    notAfter,
+                    CreateSerialNumber());
+            X509Certificate2 intermediateCertificateWithPrivateKey =
+                intermediatePublicCertificate.CopyWithPrivateKey(intermediateKey);
+
+            using RSA clientKey = RSA.Create(2048);
+            CertificateRequest clientRequest = CreateCertificateRequest(
+                $"CN=OpenAI mTLS Test Client {certificateSetId}",
+                clientKey);
+            AddEndEntityExtensions(clientRequest, ClientAuthenticationOid);
```
HAYDEN-OAI: **[P2] Make the hermetic certificates satisfy the actual OpenAI mTLS requirements.** Every successful loopback request currently uses a client certificate that the [published service requirements](https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program) reject: `AddEndEntityExtensions` supplies only `DigitalSignature` and a subject-key identifier, but the client also needs `KeyEncipherment`, an authority-key identifier, and a subject-alternative-name extension. The generated CA likewise lacks an authority-key identifier and expires after one hour even though uploaded CAs must remain valid for more than one day. Because the loopback validator checks only chain trust/EKU, it reports success for a chain the real enrollment/TLS flow cannot accept. Generate service-conformant CA/leaf extensions and validity, and have the local validator enforce those same constraints.

HAYDEN-OAI (commented):
Reviewed exact head `9e9d7a9fadec33c0adbcbe15cab36156fb40d78d`. The transport approach is reasonable, but the inline comments identify one credential-scoping issue plus concrete timeout, certificate-enrollment, chain-enablement, and certificate-fixture mismatches. Existing bot feedback was intentionally not duplicated.

tests/Snippets/MutualTlsReadMeSnippets.cs
```diff
@@ -0,0 +1,88 @@
...
+            X509CertificateLoader.LoadPkcs12CollectionFromFile(
+                pfxPath,
+                Environment.GetEnvironmentVariable("OPENAI_CLIENT_PFX_PASSWORD"));
+#else
+        X509Certificate2Collection certificates = new();
+        certificates.Import(
+            pfxPath,
+            Environment.GetEnvironmentVariable("OPENAI_CLIENT_PFX_PASSWORD"),
+            X509KeyStorageFlags.DefaultKeySet);
+#endif
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d8203bc9. The snippet file is now included when SNIPPET or NET9_0_OR_GREATER is defined, so snippet generation still sees it and the net9/net10 test targets compile the exact X509CertificateLoader path. The obsolete net8 fallback was removed because this example documents .NET 9 and later.

tests/MutualTls/MutualTlsExamplesTests.cs
```diff
@@ -0,0 +1,840 @@
...
+                    client.GetStream(),
+                    leaveInnerStreamOpen: false,
+                    ValidateClientCertificate);
+                SslServerAuthenticationOptions authenticationOptions = new()
+                {
+                    CertificateRevocationCheckMode = X509RevocationMode.NoCheck,
+                    ClientCertificateRequired = true,
+                    EnabledSslProtocols = SslProtocols.Tls12,
+                    ServerCertificate = _serverCertificate,
+                };
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d8203bc9. The loopback TLS server now enables both TLS 1.2 and TLS 1.3.

tests/MutualTls/MutualTlsExamplesTests.cs
```diff
@@ -0,0 +1,840 @@
...
+        private static async Task<ReceivedRequest> ReadRequestAsync(
+            Stream stream,
+            CancellationToken cancellationToken)
+        {
+            using StreamReader reader = new(
+                stream,
+                Encoding.ASCII,
+                detectEncodingFromByteOrderMarks: false,
+                bufferSize: 1024,
+                leaveOpen: true);
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d8203bc9. The server now parses the header terminator as bytes, reads exactly Content-Length bytes from the stream, and only then decodes UTF-8. I also added a non-ASCII request-body regression test.

tests/MutualTls/MutualTlsExamplesTests.cs
```diff
@@ -0,0 +1,840 @@
...
+                    pfx,
+                    password,
+                    X509KeyStorageFlags.DefaultKeySet);
+#else
+            X509Certificate2Collection certificates = new();
+#pragma warning disable SYSLIB0057
+            certificates.Import(
+                pfx,
+                password,
+                X509KeyStorageFlags.DefaultKeySet);
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d8203bc9. The PFX round-trip test is now net9+ and exclusively uses X509CertificateLoader.LoadPkcs12Collection; the obsolete Import branch and warning suppressions were removed.

tests/Snippets/MutualTlsReadMeSnippets.cs
```diff
@@ -0,0 +1,88 @@
...
+        X509Certificate2Collection certificates =
+            X509CertificateLoader.LoadPkcs12CollectionFromFile(
+                pfxPath,
+                Environment.GetEnvironmentVariable("OPENAI_CLIENT_PFX_PASSWORD"));
+#else
+        X509Certificate2Collection certificates = new();
+        certificates.Import(
+            pfxPath,
+            Environment.GetEnvironmentVariable("OPENAI_CLIENT_PFX_PASSWORD"),
+            X509KeyStorageFlags.DefaultKeySet);
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d8203bc9. The compile-time snippet now uses X509CertificateLoader.LoadPkcs12CollectionFromFile directly on its documented net9+ targets, with no obsolete Import fallback.

examples/MutualTls/Example01_MutualTlsAsync.cs
```diff
@@ -0,0 +1,101 @@
...
+            OpenAIClient client = new(new ApiKeyCredential(apiKey), options);
+            ChatCompletion completion = await client
+                .GetChatClient("gpt-4o-mini")
+                .CompleteChatAsync("Reply with exactly: mTLS request succeeded.");
+
+            Console.WriteLine(completion.Content[0].Text);
+        }
+        finally
+        {
+            foreach (X509Certificate2 certificate in certificates)
```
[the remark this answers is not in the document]
jbeckwith-oai: Reviewed in d8203bc9. X509Certificate2Collection does not implement IDisposable, so the suggested using declaration would not compile. The finally loop is intentionally retained to dispose every loader-created X509Certificate2 after the request completes; HttpClient already owns and disposes its SocketsHttpHandler.

tests/Snippets/MutualTlsReadMeSnippets.cs
```diff
@@ -0,0 +1,88 @@
...
+                new ApiKeyCredential(apiKey),
+                options);
+
+            ChatCompletion completion = await client
+                .GetChatClient("gpt-4o-mini")
+                .CompleteChatAsync("Reply with exactly: mTLS request succeeded.");
+        }
+        finally
+        {
+            foreach (X509Certificate2 certificate in certificates)
```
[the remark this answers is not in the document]
jbeckwith-oai: Reviewed in d8203bc9. X509Certificate2Collection does not implement IDisposable, so a using declaration for it would not compile. The generated snippet intentionally keeps the finally loop so every certificate object returned by the collection loader is deterministically disposed after the request.

tests/MutualTls/MutualTlsExamplesTests.cs
```diff
@@ -0,0 +1,840 @@
...
+            _cancellationSource.Cancel();
+            _listener.Stop();
+
+            try
+            {
+                await _serveTask.ConfigureAwait(false);
+            }
+            catch (Exception) when (_cancellationSource.IsCancellationRequested)
+            {
+            }
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d8203bc9. ServeOnceAsync now handles cancellation centrally and records only unexpected exceptions; DisposeAsync can await the task directly with no empty catch block.

examples/MutualTls/Example01_MutualTlsAsync.cs
jbeckwith-oai: Fixed in d8203bc9. The example now accepts only the global and EU OpenAI mTLS hosts over HTTPS at the default port and the /v1 base path, with no credentials, query, or fragment. Eleven allowlist tests cover supported URLs and credential-exfiltration variants.

README.md
jbeckwith-oai: Fixed in d8203bc9. The README snippet, runnable example, and test transport now set HttpClient.Timeout to InfiniteTimeSpan so OpenAIClientOptions.NetworkTimeout remains authoritative.

examples/MutualTls/README.md
jbeckwith-oai: Fixed in d8203bc9. Both READMEs now tell customers to upload and activate the issuing CA while keeping the leaf certificate and private key in the application PFX, and retain the developers.openai.com Certificates API link.

README.md
jbeckwith-oai: Fixed in d8203bc9. Both READMEs now call out that chain support is available by request and must be enabled, and explain that the client leaf must be directly signed by the uploaded CA otherwise.

tests/MutualTls/MutualTlsExamplesTests.cs
jbeckwith-oai: Fixed in d8203bc9. The fixture now generates seven-day CA and leaf certificates with SKI, key-identifier AKI, required key usage, EKU, and SAN extensions. The loopback validator enforces the published CA size, validity, basic-constraints, key-usage, identifier, client EKU, and SAN requirements. The support code was also split into focused files.

tests/MutualTls/MutualTlsExamplesTests.cs
```diff
@@ -0,0 +1,378 @@
...
+                new X509Certificate2[]
+                {
+                    certificates.ClientCertificate,
+                    certificates.IntermediateCertificate,
+                }),
+            certificates.ServerCertificate);
+
+        using HttpResponseMessage response = await httpClient.PostAsync(
+            new Uri(server.Endpoint, "unicode"),
+            new StringContent(requestBody, Encoding.UTF8));
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in a9a9869b. The request content now has an explicit using declaration and is passed to PostAsync, so it is deterministically disposed after the request.

tests/MutualTls/MutualTlsServer.cs
```diff
@@ -0,0 +1,433 @@
...
+                    _cancellationSource.Token)
+                .ConfigureAwait(false);
+        }
+        catch (Exception exception)
+        {
+            if (!_cancellationSource.IsCancellationRequested)
+            {
+                Failure = exception;
+            }
+        }
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in a9a9869b. ServeOnceAsync now handles cancellation, listener disposal, and cancellation-related socket closure explicitly; it records only expected non-cancellation socket, I/O, TLS authentication, and cryptographic failures, while unexpected programming exceptions are allowed to fault the task.

HAYDEN-OAI (approved):
Re-reviewed exact head `a9a9869b2aec1d29852cb6a164bc8c054767f3b7` and both follow-up commits after `9e9d7a9fadec33c0adbcbe15cab36156fb40d78d`. The prior credential/client-certificate origin exposure, 100-second HttpClient timeout, CA-enrollment guidance, explicit chain-enablement requirement, and nonconformant certificate fixtures are resolved. Also rechecked URI/origin restrictions, redirect safety, TLS and certificate-chain validation, byte-based request parsing, transport/certificate lifetimes, and scoped exception handling. No substantive correctness or security findings remain.

examples/MutualTls/Example01_MutualTlsAsync.cs
```diff
@@ -0,0 +1,151 @@
...
+                certificates.Single(certificate => certificate.HasPrivateKey);
+            X509Certificate2Collection intermediateCertificates = new(
+                certificates
+                    .Where(certificate => !certificate.HasPrivateKey)
+                    .ToArray());
+            SslStreamCertificateContext certificateContext =
+                SslStreamCertificateContext.Create(
+                    clientCertificate,
+                    intermediateCertificates,
+                    offline: true);
```
jsquire, line 50: The `offline: true` flag disables CRL and OCSP revocation checks during the TLS handshake, which keeps connection setup predictable and avoids a network dependency on a certificate distribution point. Worth a brief comment here so readers understand the intent rather than wondering whether it was overlooked.

examples/MutualTls/Example01_MutualTlsAsync.cs
```diff
@@ -0,0 +1,151 @@
...
+            || endpoint.IdnHost.Equals(
+                "mtls-eu.api.openai.com",
+                StringComparison.OrdinalIgnoreCase));
+    }
+
+    [TestCase(GlobalEndpoint)]
+    [TestCase(EuropeanUnionEndpoint)]
+    [TestCase("https://mtls.api.openai.com:443/v1")]
+    [TestCase("https://mtls.api.openai.com/v1/")]
+    public void GetEndpointAcceptsSupportedMtlsBaseUrls(string value)
```
jsquire: Could you move these tests out of the example file? The intent for this file is to be a self-contained runnable sample that developers can read straight through to understand the pattern. Having test methods mixed in with the example code makes it harder to follow as documentation. The endpoint validation tests would be a good fit in `MutualTlsExamplesTests.cs`.

examples/MutualTls/Example01_MutualTlsAsync.cs
jbeckwith-oai: Fixed in 72cf733b. The runnable example now explains that `offline: true` builds the local chain without revocation network lookups, and the same rationale is included in the generated README snippet so the two examples stay synchronized.

examples/MutualTls/Example01_MutualTlsAsync.cs
jbeckwith-oai: Fixed in 72cf733b. The endpoint-policy tests now live in a dedicated `examples/MutualTls/MutualTlsExamplesTests.cs` partial-class file, leaving the runnable example linear and self-contained while preserving exact coverage of its private validation methods without exposing test hooks or adding a cross-project dependency.

HAYDEN-OAI (approved):
Fresh final-head review of exact commit 72cf733ba854664a5a1e3f8c3b21eed87797f325: all five prior credential/certificate-origin, timeout-ownership, CA-enrollment, chain-enablement, and fixture-conformance findings remain resolved. Verified offline certificate-chain construction, endpoint-policy separation, origin allowlisting, redirect safety, transport and certificate ownership, and SDK bearer-key authentication. No substantive findings.

tests/MutualTls/MutualTlsEndpointTests.cs
```diff
@@ -0,0 +1,30 @@
+// Keeps endpoint security-policy tests separate from the runnable mTLS example.
+
+using NUnit.Framework;
+using System;
+
+namespace OpenAI.Examples;
+
+public partial class MutualTlsExamples
```
jsquire, line 11: This isn't a pattern that we generally use for examples, and my fear is that including it here is confusing.

If we really feel that the example needs tests beyond the example itself, I think what we'd want to do is move the tests partial to the tests project and pull in the example partial as a linked file.

Something like:
```xml
<ItemGroup>
  <Compile Include="..\Examples\Example01_MutualTlsAsync.cs" LinkBase="Shared\Examples" />
</ItemGroup>
```

I'd appreciate thoughts from @joseharriaga on this as well.

jbeckwith-oai: Fixed in 976efb77. The endpoint-test partial now lives at `tests/MutualTls/MutualTlsEndpointTests.cs`, and `OpenAI.Tests.csproj` links the runnable example source under `Shared\Examples` on .NET 9+ (the sample uses `X509CertificateLoader`). This keeps `examples/` free of the nonstandard test partial while preserving exact coverage of the sample’s private endpoint policy without adding public test hooks.

tests/MutualTls/MutualTlsExamplesTests.cs
```diff
@@ -0,0 +1,380 @@
...
+        if (clientCertificates.Count > 0)
+        {
+            X509Certificate2 clientCertificate =
+                clientCertificates.Single(certificate => certificate.HasPrivateKey);
+            X509Certificate2Collection intermediateCertificates = new(
+                clientCertificates
+                    .Where(certificate => !certificate.HasPrivateKey)
+                    .ToArray());
+            handler.SslOptions.ClientCertificateContext =
+                SslStreamCertificateContext.Create(
```
HAYDEN-OAI, line 322: **[P2] Remove generated intermediates from Windows certificate stores.** On Windows, `SslStreamCertificateContext.Create(clientCertificate, intermediateCertificates, offline: true)` is not in-memory: when these freshly generated intermediates are absent, .NET opens the LocalMachine Intermediate Certification Authorities store (falling back to CurrentUser) and permanently adds the intermediate while constructing the context. Every successful-chain, redirect, and authorization test therefore leaves a new fixture CA in machine/user state; disposing the handler and `CertificateFixture` does not remove the store entry. The behavior is explicit in the [.NET 10 implementation](https://github.com/dotnet/runtime/blob/v10.0.0/src/libraries/System.Net.Security/src/System/Net/Security/SslStreamCertificateContext.Windows.cs#L24-L104) and also exists in .NET 8/9. Please isolate this Windows-specific path or track and remove the inserted intermediate certificates during teardown so the advertised hermetic tests cannot persistently pollute developer or CI certificate stores.

HAYDEN-OAI (commented):
Independently re-reviewed exact pushed head `976efb77d40d0a7cf546e754501879022c2e5553`, all 31 prior reviews, all 36 inline comments, and all 18 resolved review threads. The prior origin-allowlisting, timeout-ownership, issuing-CA enrollment, chain-enablement, and certificate-fixture findings remain fixed. Endpoint-test relocation, .NET 9+ sample linkage, redirects, malformed requests, certificate/transport lifetimes, and offline validation were rechecked. One substantive Windows-specific hermetic-isolation issue remains inline; submitting COMMENT without requesting changes.

jbeckwith-oai:
@jsquire 👋  I would love to get more feedback, or a stamp from someone on your team before merging this!

jsquire:
> @jsquire 👋 I would love to get more feedback, or a stamp from someone on your team before merging this!

Yes.  Please do not merge without approval from one of the .NET core team.   At the moment, we're waiting on @joseharriaga to take a look, which he's told me that he plans to do today.

joseharriaga (approved):

joseharriaga:
Thank you, @jbeckwith-oai ! Approved and merged.

Issue #1244, opened 2026-07-08
[Architecture] Audit enum use in the public API


jsquire:
In some cases, such as [WebSearchToolCallStatus](https://github.com/openai/openai-dotnet/blob/main/OpenAI.Responses/src/Custom/Items/WebSearchTool/WebSearchCallStatus.cs#L10), input and output data are modeled as `enum`, resulting in a closed set. While this does offer some convenience around strong typing, it is a source of friction as the service evolves.   

We should look at all OpenAI sets of data, both input and output, as unstable and strongly consider modeling them as extensible enums.

PR #1267, opened 2026-08-03
Convert Experimental enums to extensible enums

MaiLinhP:
Addresses https://github.com/openai/openai-dotnet/issues/1244

Converted all experimental fixed C# enums in the Responses, Assistants, VectorStores, and Moderations areas into extensible enums (readonly partial struct) so the SDK gracefully handles unknown API values without data loss or exceptions. This involved creating named TypeSpec unions with `@@alternateType` and `@@dynamicModel` for code-generated types, plus manual struct conversions for types like `MessageImageDetail` that lack TypeSpec backing. `StreamingUpdateReason` was intentionally kept as a plain enum since it's an SDK-internal routing concept for the deprecated Assistants API with no extensibility benefit.


specification/client/models/assistants.models.tsp
```diff
@@ -0,0 +1,27 @@
+import "../../base/typespec/runs/main.tsp";
```
joseharriaga: Instead of adding these unions in these separate files, could you amend the base TypeSpec? That way the base TypeSpec is correct and consistent. Additionally, you won't need to use the `@access` and `@usage` decorators to expose, and you won't need to swap the types via `@alternateType` either.

api/OpenAI.net10.0.cs
```diff
@@ -460,19 +460,41 @@ public class MessageFailureDetails : IJsonModel<MessageFailureDetails>, IPersist
         public override readonly string ToString();
     }
     [Experimental("OPENAI001")]
-    public enum MessageImageDetail {
```
joseharriaga: Please sync to the latest and re-generate the API listings, which are now organized by namespace instead of it being one single giant file. 🙂

specification/client/models/assistants.models.tsp
MaiLinhP: Fixed!

api/OpenAI.net10.0.cs
MaiLinhP: Fixed!

api/net10.0/OpenAI.Responses.net10.0.cs
```diff
@@ -121,12 +121,21 @@ public class CodeInterpreterCallResponseItem : ResponseItem, IJsonModel<CodeInte
         public CodeInterpreterCallStatus? Status { get; set; }
     }
     [Experimental("OPENAI001")]
-    public enum CodeInterpreterCallStatus {
-        InProgress = 0,
-        Interpreting = 1,
-        Completed = 2,
-        Incomplete = 3,
-        Failed = 4
```
joseharriaga, line 129: It appears that two values were accidentally deleted: `Incomplete` and `Failed`.

OpenAI/src/Custom/Assistants/MessageImageDetail.Serialization.cs
```diff
@@ -4,19 +4,7 @@ namespace OpenAI.Assistants;
 
 internal static partial class MessageImageDetailExtensions
 {
-    public static string ToSerialString(this MessageImageDetail value) => value switch
```
joseharriaga: Now that this is an extensible enum, these extension methods are not needed anymore. The entire file can be deleted.

api/net10.0/OpenAI.Assistants.net10.0.cs
```diff
@@ -363,19 +363,41 @@ public class MessageFailureDetails : IJsonModel<MessageFailureDetails>, IPersist
         public override readonly string ToString();
     }
     [Experimental("OPENAI001")]
-    public enum MessageImageDetail {
```
joseharriaga, line 366: Since Assistants will be deprecated soon, I recommend reverting the changes to it for now, and we can re-evaluate it later.

OpenAI.Responses/src/Custom/ResponseTextFormatKind.cs
```diff
@@ -1,16 +1,42 @@
...
 namespace OpenAI.Responses;
 
 // CUSTOM:
 // - Added Experimental attribute.
 // - Renamed.
-// - Plain enum type, with Unknown, to convert from an underlying extensible enum
+// - Converted to extensible enum.
 [Experimental("OPENAI001")]
-public enum ResponseTextFormatKind
+public readonly partial struct ResponseTextFormatKind : IEquatable<ResponseTextFormatKind>
```
joseharriaga, line 12: There are two types for the same definition: `ResponseTextFormatKind` and `InternalResponsesTextFormatType`. This was necessary to be able to store the underlying value of "Unknown" somewhere, but with your changes here, that is no longer needed because an extensible enum can hold any value by itself. 

Then, we should reconcile these separate types into one single type that is is auto-generated instead of being written in custom code.

1. Delete this custom type.
2. Remove `InternalResponsesTextFormatType` from GeneratorStubs.cs and put it in its own file.
3. Change it from internal to public.
4. Rename it to `ResponseTextFormatKind`.
5. In Custom/ResponseTextFormat.cs, remove the `InternalType` property of type `InternalResponsesTextFormatType` and move its `[CodeGenMember]` attribute to the `Kind` property. You should be able to remove its custom implementations of `get` and `set` too.

OpenAI.Responses/src/Custom/ResponseContentPartKind.cs
```diff
@@ -1,18 +1,46 @@
...
 namespace OpenAI.Responses;
 
 // CUSTOM:
 // - Added Experimental attribute.
 // - Renamed.
-// - Plain enum type, with Unknown, to convert from an underlying extensible enum
+// - Converted to extensible enum.
 [Experimental("OPENAI001")]
-public enum ResponseContentPartKind
+public readonly partial struct ResponseContentPartKind : IEquatable<ResponseContentPartKind>
```
joseharriaga, line 12: This is the same case as `ResponseTextFormatKind`.

There are two types for the same definition: `ResponseContentPartKind` and `InternalItemContentType`. With your changes, this separation is no longer necessary, so we should reconcile them into one single type that is auto-generated instead of being written in custom code.

1. Delete this custom type.
2. Remove `InternalItemContentType` from GeneratorStubs.cs and put it in its own file.
3. Change it from internal to public.
4. Rename it to `ResponseContentPartKind`.
5. In Custom/ResponseContentPart.cs, remove the `InternalType` property of type `InternalItemContentType` and move its `[CodeGenMember]` attribute to the `Kind` property. You should be able to remove its custom implementations of `get` and `set` too.

OpenAI.Responses/src/Custom/Items/Message/MessageRole.cs
```diff
@@ -1,17 +1,44 @@
...
 namespace OpenAI.Responses;
 
 // CUSTOM:
 // - Added Experimental attribute.
 // - Renamed.
-// - Plain enum type, with Unknown, to convert from an underlying extensible enum.
+// - Converted to extensible enum.
 [Experimental("OPENAI001")]
-public enum MessageRole
+public readonly partial struct MessageRole : IEquatable<MessageRole>
```
joseharriaga, line 12: This is the same case as `ResponseTextFormatKind`.

There are two types for the same definition: `MessageRole` and `InternalResponsesMessageRole`. With your changes, this separation is no longer necessary, so we should reconcile them into one single type that is auto-generated instead of being written in custom code.

1. Delete this custom type.
2. Remove `InternalResponsesMessageRole` from GeneratorStubs.cs and put it in its own file.
3. Change it from internal to public.
4. Rename it to `MessageRole`.
5. In Custom/Items/Message/MessageResponseItem.cs, remove the `InternalType` property of type `InternalResponsesMessageRole` and move its `[CodeGenMember]` attribute to the `Role` property. You should be able to remove its custom implementations of `get` and `set` too.

OpenAI.Responses/src/Custom/Tools/ResponseToolChoiceKind.cs
```diff
@@ -1,19 +1,50 @@
...
 
 namespace OpenAI.Responses;
 
 // CUSTOM:
 // - Added Experimental attribute.
 // - Renamed.
+// - Converted to extensible enum.
 [Experimental("OPENAI001")]
-public enum ResponseToolChoiceKind
+public readonly partial struct ResponseToolChoiceKind : IEquatable<ResponseToolChoiceKind>
```
joseharriaga, line 12: This is the same case as `ResponseTextFormatKind`.

There are two types for the same definition: `ResponseToolChoiceKind` and `InternalToolChoiceObjectType `. With your changes, this separation is no longer necessary, so we should reconcile them into one single type that is auto-generated instead of being written in custom code.

OpenAI.Responses/src/Custom/Tools/Internal/InternalToolChoiceObjectType.Serialization.cs
```diff
@@ -4,13 +4,6 @@ internal static partial class InternalToolChoiceObjectTypeExtensions
...
     {
-        return objectType.ToString() switch
-        {
-            "file_search" => ResponseToolChoiceKind.FileSearch,
-            "function" => ResponseToolChoiceKind.Function,
-            "computer" => ResponseToolChoiceKind.Computer,
-            "web_search" => ResponseToolChoiceKind.WebSearch,
-            _ => ResponseToolChoiceKind.Unknown,
-        };
+        return new ResponseToolChoiceKind(objectType.ToString());
```
joseharriaga, line 7: Since this is now an extensible enum, these extension methods are no longer needed. The entire file can be deleted.

OpenAI.Responses/src/Custom/Tools/Internal/InternalToolChoiceOptions.Serialization.cs
```diff
@@ -4,12 +4,6 @@ internal static partial class InternalToolChoiceOptionsExtensions
...
     internal static ResponseToolChoiceKind ToResponseToolChoiceKind(this InternalToolChoiceOptions options)
     {
-        return options.ToString() switch
-        {
-            "none" => ResponseToolChoiceKind.None,
-            "auto" => ResponseToolChoiceKind.Auto,
-            "required" => ResponseToolChoiceKind.Required,
-            _ => ResponseToolChoiceKind.Unknown,
-        };
+        return new ResponseToolChoiceKind(options.ToString());
```
joseharriaga, line 7: Since this is now an extensible enum, these extension methods are no longer needed. The entire file can be deleted.

joseharriaga (changes requested):

OpenAI.Responses/src/Custom/Tools/ResponseToolChoiceKind.cs
joseharriaga: Talked offline. `ToolChoice` will be resolved separately and can be excluded from this PR. Filed an issue to track it here: https://github.com/openai/openai-dotnet/issues/1361

OpenAI.Responses/src/Custom/Tools/Internal/InternalToolChoiceObjectType.Serialization.cs
joseharriaga: Talked offline. `ToolChoice` will be resolved separately and can be excluded from this PR. Filed an issue to track it here: https://github.com/openai/openai-dotnet/issues/1361

PR #1272, opened 2026-08-05
Forward the cancellation token in the synchronous ModifyAssistant

adityasingh2400:
The synchronous `AssistantClient.ModifyAssistant` overload passed a literal `null` to the protocol method instead of the caller's cancellation token, so the token was dropped before the request was sent and the call could not be cancelled. Every other convenience method on this client already forwards its token through `ToRequestOptions`, including `ModifyAssistant`'s own asynchronous twin one method above it, so this was an isolated gap rather than a deliberate difference. The change makes the synchronous overload match its sibling and honor the behavior its own documentation comment describes. Callers who pass a default token see no change at all, because `ToRequestOptions` returns `null` for `default`, which is exactly what was being passed before. The added mock transport regression test exercises both overloads and fails only on the synchronous one without this fix.


tests/Assistants/AssistantsMockTests.cs
```diff
@@ -21,6 +24,41 @@ public AssistantsMockTests(bool isAsync) : base(isAsync)
...
+
+        AssistantClient client = CreateProxyFromClient(new AssistantClient(s_fakeCredential, options));
+
+        using CancellationTokenSource cancellationSource = new();
+        cancellationSource.Cancel();
+
+        // Without the token the mock transport answers 200 and the call succeeds, so
+        // reaching this assertion at all depends on the token being forwarded.
+        Assert.That(
+            async () => await client.ModifyAssistantAsync("asst_abc", new AssistantModificationOptions(), cancellationSource.Token),
```
jsquire, line 56: Could we please exercise the synchronous overload in this test as well? The only call here is `ModifyAssistantAsync`, so this regression test never reaches the synchronous `ModifyAssistant` implementation changed in `AssistantClient.cs`. As written, a future regression that drops the token from the synchronous overload would still leave this test green. Please branch on `IsAsync` and call `ModifyAssistant` for the synchronous case, matching the pattern in `StreamingRunSurfacesErrorEventAsException`.

---
:robot: jsquire-copilot

adityasingh2400:
Good catch, that was a real gap. `ExpectSyncPipeline = !IsAsync` set the transport up for both modes, but the body only ever called `ModifyAssistantAsync`, so the synchronous path this PR fixes was never reached and the sync half of the fixture proved nothing.

Now branched on `IsAsync`, matching `StreamingRunSurfacesErrorEventAsException`:

```csharp
if (IsAsync)
{
    Assert.That(
        async () => await client.ModifyAssistantAsync("asst_abc", new AssistantModificationOptions(), cancellationSource.Token),
        Throws.InstanceOf<OperationCanceledException>());
}
else
{
    Assert.That(
        () => client.ModifyAssistant("asst_abc", new AssistantModificationOptions(), cancellationSource.Token),
        Throws.InstanceOf<OperationCanceledException>());
}
```

One thing I should be straight about: I could not run this locally. `global.json` pins SDK 10.0.400 and I only have 9.0.301, so I am relying on CI rather than a local green run. The sync overload takes `CancellationToken cancellationToken = default` at `AssistantClient.cs:201`, so the call shape matches, but please treat the pass as unverified by me until CI says otherwise.

PR #1273, opened 2026-08-05
Keep reading SSE events after an event that yields no updates

adityasingh2400:
The server-sent event enumerators advanced the outer event enumerator with an `if` rather than a `while`, so an event that deserializes to zero updates fell through to `return false` and ended the entire stream instead of being skipped. The visible effect is a stream that stops at the first unrecognized event and looks like a clean, early completion, so callers silently lose every later update.

Assistants streaming is the exposed caller, because `StreamingUpdate.FromSseItem` returns an empty sequence for any event kind it does not model, and its own comment describes those events as benign. The same two enumerators also back chat and audio streaming. Both the synchronous `SseUpdateEnumerator` and the asynchronous `AsyncSseUpdateEnumerator` now loop until an event actually produces an update, the terminal `[DONE]` frame arrives, or the events run out.

The existing unknown-event test placed the unmodeled event last, which is the one arrangement where the truncation is invisible. The new test puts unmodeled events in the middle and asserts that the later modeled updates still arrive. It fails on current main with 1 update instead of 3, on both the sync and async fixtures, and passes with the fix, alongside 672 passing Smoke tests on net10.0. The `net8.0` and `net9.0` test targets could not run here because only the .NET 10 runtime is installed, though the library builds clean for all three client target frameworks.

Heads up that #1226 touches these same two files to add constructors. There is no logical overlap with this change, but whichever merges second will need a trivial textual rebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bugs Fixed
 
-- Fixed streaming responses ending early when the service emits an event that the library does not model. The server-sent event enumerator stopped at the first event that produced no updates, so an unrecognized event in the middle of a stream silently terminated the whole stream and looked like a clean, early completion. Unrecognized events are now skipped and every later update still surfaces. This affects all streaming APIs, including `OpenAI.Assistants`, `OpenAI.Chat`, and `OpenAI.Audio`, on both the synchronous and asynchronous paths.
+- Fixed streaming responses ending early when the service emits an event that the library does not model. The server-sent event enumerator stopped at the first event that produced no updates, so an unrecognized event in the middle of a stream silently terminated the whole stream and looked like a clean, early completion. Unrecognized events are now skipped and every later update still surfaces. This affects all streaming APIs, including `OpenAI.Assistants`, `OpenAI.Chat`, `OpenAI.Responses`, and `OpenAI.Audio`, on both the synchronous and asynchronous paths.
 
 ## 2.13.0 (2026-08-10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Release History
 
+## (Unreleased)
+
+### Bugs Fixed
+
+- Fixed streaming responses ending early when the service emits an event that the library does not model. The server-sent event enumerator stopped at the first event that produced no updates, so an unrecognized event in the middle of a stream silently terminated the whole stream and looked like a clean, early completion. Unrecognized events are now skipped and every later update still surfaces. This affects all streaming APIs, including `OpenAI.Assistants`, `OpenAI.Chat`, and `OpenAI.Audio`, on both the synchronous and asynchronous paths.
+
 ## 2.13.0 (2026-08-10)
 
 ### Acknowledgments
--- a/OpenAI/src/Utility/AsyncSseUpdateCollection.cs
+++ b/OpenAI/src/Utility/AsyncSseUpdateCollection.cs
@@ -199,8 +199,19 @@ internal class AsyncSseUpdateCollection<T> : AsyncCollectionResult<T>
             // must be skipped so that later events still surface, otherwise a single
             // unrecognized event in the middle of a stream would end the whole stream and
             // look like a clean, early completion.
-            while (await _events.MoveNextAsync().ConfigureAwait(false))
+            while (true)
             {
+                // Cancellation is observed once per event rather than only on entry, because
+                // a run of events that yield no updates is consumed inside a single call.
+                // Passing the token to the event enumerator is not enough on its own, since
+                // events the parser has already buffered are returned without a read.
+                _cancellationToken.ThrowIfCancellationRequested();
+
+                if (!await _events.MoveNextAsync().ConfigureAwait(false))
+                {
+                    break;
+                }
+
                 if (_events.Current.Data.AsSpan().SequenceEqual(TerminalData))
                 {
                     _current = default;
--- a/OpenAI/src/Utility/AsyncSseUpdateCollection.cs
+++ b/OpenAI/src/Utility/AsyncSseUpdateCollection.cs
@@ -193,7 +193,13 @@ internal class AsyncSseUpdateCollection<T> : AsyncCollectionResult<T>
                 return true;
             }
 
-            if (await _events.MoveNextAsync().ConfigureAwait(false))
+            // Keep advancing the outer event enumerator instead of stopping at the first
+            // event. An event can legitimately deserialize to zero updates, for example an
+            // unmodeled event that the deserializer maps to an empty sequence. Such an event
+            // must be skipped so that later events still surface, otherwise a single
+            // unrecognized event in the middle of a stream would end the whole stream and
+            // look like a clean, early completion.
+            while (await _events.MoveNextAsync().ConfigureAwait(false))
             {
                 if (_events.Current.Data.AsSpan().SequenceEqual(TerminalData))
                 {
--- a/OpenAI/src/Utility/SseUpdateCollection.cs
+++ b/OpenAI/src/Utility/SseUpdateCollection.cs
@@ -172,6 +172,10 @@ internal class SseUpdateCollection<T> : CollectionResult<T>
             // look like a clean, early completion.
             while (_events.MoveNext())
             {
+                // Cancellation is observed per event rather than only on entry, because a
+                // run of events that yield no updates is consumed inside a single call.
+                _cancellationToken.ThrowIfCancellationRequested();
+
                 if (_events.Current.Data.AsSpan().SequenceEqual(TerminalData))
                 {
                     _current = default;
--- a/OpenAI/src/Utility/SseUpdateCollection.cs
+++ b/OpenAI/src/Utility/SseUpdateCollection.cs
@@ -170,11 +170,18 @@ internal class SseUpdateCollection<T> : CollectionResult<T>
             // must be skipped so that later events still surface, otherwise a single
             // unrecognized event in the middle of a stream would end the whole stream and
             // look like a clean, early completion.
-            while (_events.MoveNext())
+            while (true)
             {
-                // Cancellation is observed per event rather than only on entry, because a
-                // run of events that yield no updates is consumed inside a single call.
+                // Cancellation is observed once per event rather than only on entry, because
+                // a run of events that yield no updates is consumed inside a single call.
+                // The check sits ahead of the read so that a canceled token does not have to
+                // wait for the server to send another event before it is noticed.
                 _cancellationToken.ThrowIfCancellationRequested();
+
+                if (!_events.MoveNext())
+                {
+                    break;
+                }
 
                 if (_events.Current.Data.AsSpan().SequenceEqual(TerminalData))
                 {
--- a/OpenAI/src/Utility/SseUpdateCollection.cs
+++ b/OpenAI/src/Utility/SseUpdateCollection.cs
@@ -164,7 +164,13 @@ internal class SseUpdateCollection<T> : CollectionResult<T>
                 return true;
             }
 
-            if (_events.MoveNext())
+            // Keep advancing the outer event enumerator instead of stopping at the first
+            // event. An event can legitimately deserialize to zero updates, for example an
+            // unmodeled event that the deserializer maps to an empty sequence. Such an event
+            // must be skipped so that later events still surface, otherwise a single
+            // unrecognized event in the middle of a stream would end the whole stream and
+            // look like a clean, early completion.
+            while (_events.MoveNext())
             {
                 if (_events.Current.Data.AsSpan().SequenceEqual(TerminalData))
                 {
--- a/tests/Assistants/AssistantsMockTests.cs
+++ b/tests/Assistants/AssistantsMockTests.cs
@@ -127,4 +127,67 @@ public class AssistantsMockTests : ClientTestBase
         Assert.That(updates, Has.Count.EqualTo(1));
         Assert.That(updates[0].UpdateKind, Is.EqualTo(StreamingUpdateReason.RunCreated));
     }
+
+    [Test]
+    public async Task StreamingRunContinuesAfterBenignUnknownEvent()
+    {
+        // An unmodeled event yields zero updates. When such an event appears in the
+        // middle of a stream, the enumerator must skip it and keep reading, so that
+        // every later modeled event still surfaces. Placing the unknown event last
+        // cannot detect this, because there is nothing after it to lose.
+        MockPipelineResponse response = new MockPipelineResponse(200).WithContent(
+            """
+            event: thread.run.created
+            data: {"id":"run_abc","object":"thread.run","status":"queued"}
+
+            event: thread.run.some_future_event
+            data: {"id":"run_abc","object":"thread.run","status":"queued"}
+
+            event: thread.run.in_progress
+            data: {"id":"run_abc","object":"thread.run","status":"in_progress"}
+
+            event: thread.run.another_future_event
+            data: {"id":"run_abc","object":"thread.run","status":"in_progress"}
+
+            event: thread.run.completed
+            data: {"id":"run_abc","object":"thread.run","status":"completed"}
+
+            event: done
+            data: [DONE]
+            """);
+
+        OpenAIClientOptions options = new()
+        {
+            Transport = new MockPipelineTransport(_ => response)
+            {
+                ExpectSyncPipeline = !IsAsync
+            }
+        };
+
+        AssistantClient client = new(s_fakeCredential, options);
+
+        List<StreamingUpdate> updates = new();
+
+        if (IsAsync)
+        {
+            await foreach (StreamingUpdate update in client.CreateRunStreamingAsync("thread_abc", "asst_abc"))
+            {
+                updates.Add(update);
+            }
+        }
+        else
+        {
+            foreach (StreamingUpdate update in client.CreateRunStreaming("thread_abc", "asst_abc"))
+            {
+                updates.Add(update);
+            }
+        }
+
+        // All three modeled events must arrive. Before the fix, the stream ended at the
+        // first unmodeled event and only the run.created update was ever produced.
+        Assert.That(updates, Has.Count.EqualTo(3));
+        Assert.That(updates[0].UpdateKind, Is.EqualTo(StreamingUpdateReason.RunCreated));
+        Assert.That(updates[1].UpdateKind, Is.EqualTo(StreamingUpdateReason.RunInProgress));
+        Assert.That(updates[2].UpdateKind, Is.EqualTo(StreamingUpdateReason.RunCompleted));
+    }
 }
--- a/tests/OpenAI.Tests.csproj
+++ b/tests/OpenAI.Tests.csproj
@@ -41,6 +41,16 @@
     <Compile Include="..\examples\MutualTls\Example01_MutualTlsAsync.cs" LinkBase="Shared\Examples" />
   </ItemGroup>
 
+  <!-- The SSE collections are internal and the library is signed, so they are linked
+       in rather than exposed through InternalsVisibleTo. -->
+  <ItemGroup>
+    <Compile Include="..\OpenAI\src\Utility\SseUpdateCollection.cs" LinkBase="Utility\Shared" />
+    <Compile Include="..\OpenAI\src\Utility\AsyncSseUpdateCollection.cs" LinkBase="Utility\Shared" />
+    <Compile Include="..\OpenAI\src\Generated\Internal\ModelSerializationExtensions.cs" LinkBase="Utility\Shared" />
+    <Compile Include="..\OpenAI\src\Generated\Internal\TypeFormatters.cs" LinkBase="Utility\Shared" />
+    <Compile Include="..\OpenAI\src\Generated\Internal\SerializationFormat.cs" LinkBase="Utility\Shared" />
+  </ItemGroup>
+
   <ItemGroup>
     <AssemblyAttribute Include="System.Reflection.AssemblyMetadataAttribute">
       <_Parameter1>SourcePath</_Parameter1>
--- a/tests/Utility/SseUpdateCollectionTests.cs
+++ b/tests/Utility/SseUpdateCollectionTests.cs
@@ -67,6 +67,64 @@ public class SseUpdateCollectionTests
         Assert.That(updates, Is.EqualTo(new[] { "A", "B" }));
     }
 
+    // A run of events that yield no updates, so that the skip loop stays inside a single
+    // MoveNext call across several events.
+    private const string StreamContentWithConsecutiveUnknownEvents =
+        "event: modeled\ndata: {\"value\":\"A\"}\n\n" +
+        "event: " + UnknownEventType + "\ndata: {\"value\":\"skipped\"}\n\n" +
+        "event: " + UnknownEventType + "\ndata: {\"value\":\"skipped\"}\n\n" +
+        "event: modeled\ndata: {\"value\":\"B\"}\n\n" +
+        "data: [DONE]\n\n";
+
+    [Test]
+    public void SkippingEventsWithNoUpdatesObservesCancellation()
+    {
+        using CancellationTokenSource source = new();
+
+        SseUpdateCollection<string> collection = new(
+            () => CreatePage(StreamContentWithConsecutiveUnknownEvents),
+            item => DeserializeEventAndCancelOnUnknown(item, source),
+            source.Token);
+
+        using IEnumerator<string> enumerator = collection.GetEnumerator();
+
+        Assert.That(enumerator.MoveNext(), Is.True);
+        Assert.That(enumerator.Current, Is.EqualTo("A"));
+
+        // The token is canceled while the second call is already inside the skip loop.
+        Assert.Throws<OperationCanceledException>(() => enumerator.MoveNext());
+    }
+
+    [Test]
+    public void SkippingEventsWithNoUpdatesObservesCancellationAsync()
+    {
+        using CancellationTokenSource source = new();
+
+        AsyncSseUpdateCollection<string> collection = new(
+            () => Task.FromResult(CreatePage(StreamContentWithConsecutiveUnknownEvents)),
+            item => DeserializeEventAndCancelOnUnknown(item, source),
+            source.Token);
+
+        Assert.ThrowsAsync<OperationCanceledException>(async () =>
+        {
+            await foreach (string _ in collection)
+            {
+            }
+        });
+    }
+
+    private static IEnumerable<string> DeserializeEventAndCancelOnUnknown(
+        SseItem<byte[]> item,
+        CancellationTokenSource source)
+    {
+        if (item.EventType == UnknownEventType)
+        {
+            source.Cancel();
+        }
+
+        return DeserializeEvent(item);
+    }
+
     private static IEnumerable<string> DeserializeEvent(SseItem<byte[]> item)
     {
         if (item.EventType == UnknownEventType)
@@ -78,11 +136,13 @@ public class SseUpdateCollectionTests
         return [document.RootElement.GetProperty("value").GetString()!];
     }
 
-    private static ClientResult CreatePage()
+    private static ClientResult CreatePage() => CreatePage(StreamContent);
+
+    private static ClientResult CreatePage(string content)
     {
         MockPipelineResponse response = new(200, "OK")
         {
-            ContentStream = new MemoryStream(Encoding.UTF8.GetBytes(StreamContent)),
+            ContentStream = new MemoryStream(Encoding.UTF8.GetBytes(content)),
         };
 
         return ClientResult.FromResponse(response);
--- a/tests/Utility/SseUpdateCollectionTests.cs
+++ b/tests/Utility/SseUpdateCollectionTests.cs
@@ -113,6 +113,127 @@ public class SseUpdateCollectionTests
         });
     }
 
+    // The terminal event is the next frame after the one that cancels, so the parser has it
+    // buffered and can return it without another read.
+    private const string StreamContentUnknownThenTerminal =
+        "event: modeled\ndata: {\"value\":\"A\"}\n\n" +
+        "event: " + UnknownEventType + "\ndata: {\"value\":\"skipped\"}\n\n" +
+        "data: [DONE]\n\n";
+
+    [Test]
+    public void CancellationIsNotOvertakenByABufferedTerminalEvent()
+    {
+        using CancellationTokenSource source = new();
+
+        SseUpdateCollection<string> collection = new(
+            () => CreatePage(StreamContentUnknownThenTerminal),
+            item => DeserializeEventAndCancelOnUnknown(item, source),
+            source.Token);
+
+        using IEnumerator<string> enumerator = collection.GetEnumerator();
+
+        Assert.That(enumerator.MoveNext(), Is.True);
+        Assert.Throws<OperationCanceledException>(() => enumerator.MoveNext());
+    }
+
+    [Test]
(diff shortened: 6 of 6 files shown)


OpenAI/src/Utility/SseUpdateCollection.cs
```diff
@@ -164,7 +164,13 @@ public bool MoveNext()
                 return true;
```
jsquire, line 164: Please track whether the terminal `[DONE]` event was observed and throw an appropriate incomplete-stream exception when the SSE source reaches EOF first. This prevents a truncated response from being reported as a successful, clean completion.

OpenAI/src/Utility/AsyncSseUpdateCollection.cs
```diff
@@ -193,7 +193,13 @@ async ValueTask<bool> IAsyncEnumerator<U>.MoveNextAsync()
                 return true;
```
jsquire, line 193: Please implement an actual per-read idle timeout, rather than relying only on caller cancellation. Use the existing `ClientPipelineOptions.NetworkTimeout` as the source, thread it into the shared SSE collection, and enforce it for both enumerators. Reset the timeout after each event and keep it linked to the caller token. A small illustrative shape would be:

```csharp
using CancellationTokenSource readTimeout =
    CancellationTokenSource.CreateLinkedTokenSource(callerToken);
readTimeout.CancelAfter(networkTimeout);
SseItem<byte[]> item = await ReadNextEventAsync(readTimeout.Token);
```

The synchronous path should use the equivalent timeout-aware read helper. With `NetworkTimeout = 30s`, if event A is received and the connection then remains open without sending another event, the next read should terminate after 30 seconds instead of waiting indefinitely. Please implement this through one shared timeout-aware read abstraction so sync and async behavior stay consistent.

jsquire (commented):
3. **Regression coverage**

Please add at least one regression test for a non-Assistants streaming API that uses a generated deserializer, such as Chat or Responses. The test should place an unknown event between two modeled events and assert that both modeled updates are returned, for example:

```text
modeled event A
unknown event
modeled event B
[DONE]
```

This verifies that the shared fix works beyond `StreamingUpdate.FromSseItem`.

4. **Changelog scope**

Please include `OpenAI.Responses` in the changelog entry. The shared enumerators are also used by Responses streaming APIs, so the documented impact should match the full affected API surface.

5. **Synchronous timeout**

Please apply the same per-read timeout behavior to the synchronous SSE enumerator, using the existing `ClientPipelineOptions.NetworkTimeout` value rather than adding an unrelated timeout setting. Thread the configured timeout into the shared collection and use the same timeout-aware read helper as the async path; link it with the caller token and reset it after each received event. A small illustrative shape is:

```csharp
using CancellationTokenSource readTimeout =
    CancellationTokenSource.CreateLinkedTokenSource(callerToken);
readTimeout.CancelAfter(networkTimeout);
SseItem<byte[]> item = ReadNextEvent(readTimeout.Token);
```

With `NetworkTimeout = 30s`, a stream that sends event A and then remains open without sending another event must cause the next `MoveNext()` to terminate after 30 seconds, rather than block indefinitely. This should be an actual bounded read, not only a cancellation check before calling `_events.MoveNext()`.

adityasingh2400:
Thanks, this is a careful review. Item 4 is pushed in `1801ff95`, the changelog entry now names `OpenAI.Responses`.

On the other four I want to put evidence in front of you before writing code, because three of them do not hold the way they read and one is a design call I do not think is mine.

**3. Regression coverage on Chat or Responses**

A test there cannot fail, before or after this change. Both use the single-value constructor overload, and `DeserializeSseToSingleViaJson` wraps the result in a one-element collection:

```csharp
=> DeserializeSseToMultipleViaJson<U>((e, d, o) => [jsonSingleDeserializationFunc.Invoke(e, d, o)]);
```

So those paths always produce exactly one update per event, never zero. Both deserializers also fall through to a model rather than nothing, `InternalUnknownResponseStreamEvent` for Responses. The zero-update case that this PR fixes comes only from the multi-value deserializer, and `StreamingUpdate.FromSseItem` is the only one in the tree, ending in `_ => []`. Every other construction site passes a single-value func.

I can add the test, but it would pass with the fix reverted, so it would document behaviour rather than guard this fix. Happy to write it framed that way, or to test the shared enumerator directly against a stub multi-deserializer, which does exercise the fix independently of Assistants. Tell me which you prefer.

**1. Throw an incomplete-stream exception when EOF arrives before `[DONE]`**

Responses does not send a terminator. From your own recording, `tests/SessionRecords/ResponsesTests/StreamingResponses.json`, the body carries `event:` frames and ends after `response.completed` with no `[DONE]`. All 15 `ResponsesTests` streaming records look like this, while the Assistants records all end with `event: done` / `data: [DONE]`.

So this check would throw on every Responses stream, including the recorded tests. It seems right for Assistants and wrong as a shared rule, which makes it a per-API decision rather than something to put in the shared enumerator.

**2 and 5. Bound the SSE read with `ClientPipelineOptions.NetworkTimeout`**

`NetworkTimeout` appears nowhere in library source, only in three test files as a caller-set option. The enumerator receives `(deserializerFunc, ClientResult, CancellationToken, disposalActions)` and has no handle on the pipeline or its options, so there is nothing existing to reuse at this layer. Honoring it here means threading options from each client through both collection types, which is 12 construction sites across `AssistantClient`, `ChatClient` and `ResponsesClient`.

I am happy to do that, but it is an internal API change and a second timeout mechanism layered on the transport one, so I would rather you decide the shape than have me pick.

Worth saying plainly: 1, 2 and 5 are all real gaps, they are just wider than the one-line `if` to `while` here. If you want them in this PR I will do them once you have picked the direction. If you would rather keep this one narrow, they would make a clean follow-up and I am glad to take it.

OpenAI/src/Utility/SseUpdateCollection.cs
adityasingh2400: Replied in detail in the PR comment. Short version: Responses does not send a terminator, so this check would throw on every Responses stream. `tests/SessionRecords/ResponsesTests/StreamingResponses.json` ends after `response.completed` with no `[DONE]`, and all 15 Responses streaming records look the same, while the Assistants records all end with `event: done` / `data: [DONE]`.

That makes it right for Assistants and wrong as a shared rule, so I think it belongs per-API rather than in this enumerator. Happy to implement it that way if you agree.

OpenAI/src/Utility/AsyncSseUpdateCollection.cs
adityasingh2400: Replied in detail in the PR comment. Short version: `NetworkTimeout` appears nowhere in library source, only in three test files as a caller-set option, and this enumerator receives just `(deserializerFunc, ClientResult, CancellationToken, disposalActions)`. There is no handle on the pipeline or its options here, so there is nothing existing to reuse at this layer.

Honoring it means threading options from each client through both collection types, 12 construction sites across `AssistantClient`, `ChatClient` and `ResponsesClient`, and it adds a second timeout mechanism on top of the transport one. Glad to do it, I would just rather you pick the shape first.

OpenAI/src/Utility/SseUpdateCollection.cs
jsquire: Thanks—the Responses evidence is correct. Please do not add a shared unconditional `[DONE]` requirement. Add a terminal-event predicate to both shared collection constructors and store it alongside the deserializer:

```csharp
private readonly Func<SseItem<byte[]>, bool> _isTerminalEvent;
```

In both enumerators, continue looping when the deserializer returns zero updates, return normally only after the predicate matches, and throw if the event source reaches EOF first:

```csharp
bool terminalSeen = false;

while (_events.MoveNext())
{
    if (_isTerminalEvent(_events.Current))
    {
        terminalSeen = true;
        break;
    }

    _updates = _eventDeserializerFunc.Invoke(_events.Current).GetEnumerator();
    // Yield all updates, including zero updates, then continue to the next event.
}

if (!terminalSeen)
{
    throw new InvalidOperationException("The SSE stream ended before its terminal event.");
}
```

Pass the policy from each API instead of hard-coding it in the shared utility. Assistants, Chat, and Audio should recognize `data: [DONE]`; Responses should recognize the `response.completed` event label. Apply the same policy to sync and async collections and add coverage for both terminal forms.

OpenAI/src/Utility/AsyncSseUpdateCollection.cs
jsquire: Agreed—please do not add a second timeout mechanism in this PR. The existing value is configured at client construction:

```csharp
OpenAIClientOptions options = new()
{
    NetworkTimeout = TimeSpan.FromSeconds(30)
};

ChatClient client = new("model", credential, options);
```

`ChatClient` passes `options` into `OpenAIClient.CreatePipeline(...)`, then the streaming factory passes only the request delegate, deserializer, and cancellation token to `AsyncSseUpdateCollection`; the collection has no access to `options.NetworkTimeout`. Please leave transport behavior unchanged here and track a follow-up that first verifies whether `NetworkTimeout` bounds response-body reads. If it does not, that follow-up must capture the value in each client constructor and pass it through all shared collection call sites before adding a timeout-aware stream wrapper.

The follow-up wiring would be shaped like this:

```csharp
if (options is null)
{
    options = new OpenAIClientOptions();
}

_networkTimeout = options.NetworkTimeout;
Pipeline = OpenAIClient.CreatePipeline(authenticationPolicy, options);

return new AsyncSseUpdateCollection<StreamingChatCompletionUpdate>(
    sendRequestAsync,
    deserialize,
    requestOptions.CancellationToken,
    _networkTimeout);
```

jsquire:
Agreed on the regression-test scope. Please add the test at the shared utility level using the multi-value deserializer shape that actually exposes this bug; do not add a Chat or Responses test that passes with the fix reverted.

The test should exercise both collection types with this exact behavior:

```text
modeled event A -> one update
unknown event -> empty sequence
modeled event B -> one update
terminal event
```

Assert that both updates are returned in order for sync and async enumeration. Against the pre-PR implementation, enumeration stops after event A; with the `while` change, it returns A and B. Keep the stub focused on returning `[]` only for the unknown event.

Thanks as well for pushing the changelog correction in `1801ff95`; the Responses scope is now covered.

adityasingh2400:
Terminal-event policy is implemented and pushed in `9627b5c0`. The whole solution builds clean on the pinned 10.0.100 SDK.

What landed:

- `Func<SseItem<byte[]>, bool> isTerminalEvent` added to both collections and stored alongside the deserializer, threaded through all five constructor overloads on each.
- Both enumerators keep looping past zero-update events, set `_terminalSeen` and return normally only when the predicate matches, and throw `InvalidOperationException` if the event source reaches EOF first. `_terminalSeen` is an instance field rather than a local, otherwise a second `MoveNext` after a clean terminal return would fall through and throw.
- The hard-coded `[DONE]` constants are gone from both files.
- 18 call sites now pass a policy, not 12. `AudioClient` has 6 that I missed in my earlier count, four of them target-typed `new(...)`. Assistants 6, Audio 6, Chat 2 use `[DONE]`, Responses 4 use the `response.completed` label.

I checked the policy against every recorded stream before wiring it: all 15 `ResponsesTests` recordings end with `response.completed` and none contains `[DONE]`, and all 6 Assistants recordings end with `[DONE]`. So nothing in the suite should start throwing.

One judgement call to flag. Rather than repeat a lambda 18 times I added `SseTerminalEvents` with `IsDoneData` and `ForEventLabel(string)`, and each client picks one. The shared collections still have no default and no knowledge of any terminator. If you would rather see the predicates inline at each call site, say so and I will change it.

**The utility-level test is blocked and I want your call before going further.**

`SseUpdateCollection<T>` is `internal` and the assembly is strong-name signed, so the test project cannot construct it. Adding `<InternalsVisibleTo Include="OpenAI.Tests" />` fails with CS1726, since signed assemblies need a public key, and the existing `Azure.AI.OpenAI` entry is split into a keyed variant plus a bare one gated on the `Unsigned` configuration. I wrote the test, hit this, and backed it out rather than guess at your signing setup.

Three ways forward, your preference:

1. Mirror the `Azure.AI.OpenAI` pattern with the test assembly public key, if there is one I should use.
2. Gate the `InternalsVisibleTo` on `Configuration == Unsigned` only, if that is the configuration the test suite runs under.
3. Make `SseTerminalEvents` public and keep the collections internal, testing the policy in isolation while covering the enumerator behaviour through `AssistantsMockTests`, which already reaches it.

The test itself is written and ready: modeled A, unknown event returning `[]`, modeled B, terminal, asserting both updates in order for sync and async, plus a truncated-stream case for the new throw and an event-label case for the Responses terminator. It just needs one of the three above to compile.

adityasingh2400:
I have reverted `9627b5c0`. The branch is back to `1801ff95`, which is the original one-line fix plus the changelog correction, and the suite is green there.

Codex was right on both counts, and I found a third problem while fixing them. Reverting rather than leaving a broken head on the PR while I work through it.

**1. Swallowing `response.completed`.** Confirmed. Checking the predicate before deserializing meant the terminal event was consumed as a control frame and its `StreamingResponseCompletedUpdate` was never yielded. `ResponsesTests.cs` reads that update at lines 97, 199 and 999, so this would have broken real callers.

**2. `response.incomplete` and `response.failed`.** Confirmed. Both are modeled at lines 151 and 153 of the generated deserializer. Neither matched my predicate, so a stream ending on a valid non-success outcome would run to EOF and throw.

I rewrote the policy to fix both, replacing the `bool` predicate with a disposition so a terminal event can be classified as either a control frame or a modeled event that gets yielded before enumeration stops. That builds clean, but it broke `StreamingRunIgnoresBenignUnknownEvent` and `StreamingRunContinuesAfterBenignUnknownEvent` in both sync and async, with "The SSE stream ended before its terminal event."

**3. The `[DONE]` terminator is not being matched at all in the mock tests, and never was.** The pre-change code passes those tests only because it returns false silently at EOF. Once EOF throws, it becomes visible that the loop is exiting through EOF rather than through the terminal branch, on a stream that plainly ends `event: done` / `data: [DONE]`. The comparison itself is byte-identical to the one already in the tree, so something upstream of it is different from what I assumed, either in how `MockPipelineResponse` surfaces the body or in how the parser dispatches that final event. I tried adding the SSE blank-line terminator to the fixtures and it did not change the result.

I did not want to keep guessing at that on your PR, so the revert puts it back to a known-good state.

Two things worth your input before I take another run at it:

- The EOF throw is what exposed this. It may be surfacing a genuine gap between the mock fixtures and real traffic rather than a library bug, since the recorded Assistants traffic does end with `\n\n` after `[DONE]` and the fixtures do not. If you know whether `MockPipelineResponse.WithContent` is expected to round-trip a trailing blank line, that would save me a lot of time.
- Given the terminal policy now needs three states rather than a bool, and touches 18 call sites plus a behaviour change on every stream, it may be cleaner as its own PR on top of this one. Happy either way, but this PR is currently a one-line fix and that is a much easier thing to land.

adityasingh2400:
Answering my own question from the last comment, so you do not have to spend time on it.

`MockPipelineResponse` is not the problem. `SseParser` is behaving correctly and the mock fixtures are malformed. I ran the fixture content through `SseParser.Create(...).Enumerate()` directly, varying only what follows the last `data:` line:

```
--- content ends with no trailing newline ---
    event=thread.run.created            matches[DONE]=False
    event=thread.run.some_future_event  matches[DONE]=False
                                          <- the [DONE] event is never dispatched

--- content ends with one \n ---
    event=thread.run.created            matches[DONE]=False
    event=thread.run.some_future_event  matches[DONE]=False
                                          <- still never dispatched

--- content ends with \n\n ---
    event=thread.run.created            matches[DONE]=False
    event=thread.run.some_future_event  matches[DONE]=False
    event=done  data=[DONE]             matches[DONE]=True
```

An SSE event is only dispatched once a blank line terminates it, so a stream whose final event has no blank line after it loses that event entirely. The mock fixtures end at `data: [DONE]` with the closing `"""` on the next line, so the terminator never reaches the enumerator. The existing tests pass only because the current code returns false silently at EOF, which is indistinguishable from a clean terminal stop. The recorded traffic in `tests/SessionRecords/AssistantsTests/` does end with `\n\n`, so this is a fixture defect, not a library one.

That also explains why my earlier attempt to fix it failed. I added one blank line before the closing `"""`, which yields a single `\n`, and the probe shows one `\n` is not enough. It needs two blank lines, or an explicit `+ "\n"`.

So the EOF throw you asked for is implementable, and it is doing exactly what it should: it turned a silent, invisible fixture bug into a loud one. The fixtures need fixing either way, independently of this PR.

The branch is still at `1801ff95`, the one-line fix plus changelog, green. I did not re-push the terminal-policy work because it is a behaviour change across 18 call sites plus a fixture correction, and I would rather you confirm the shape first. Say the word and I will land it, here or as a follow-up PR, whichever you prefer.

OpenAI.Responses/src/Custom/ResponsesClient.cs
```diff
@@ -237,7 +237,8 @@ public virtual CollectionResult<StreamingResponseUpdate> CreateResponseStreaming
         return new SseUpdateCollection<StreamingResponseUpdate>(
             () => CreateResponse((BinaryContent)options, cancellationToken.ToRequestOptions(streaming: true)),
             StreamingResponseUpdate.DeserializeStreamingResponseUpdate,
-            cancellationToken);
+            cancellationToken,
+            SseTerminalEvents.ForEventLabel("response.completed"));
```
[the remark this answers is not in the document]
adityasingh2400: Correct, and thanks. I confirmed it before acting: `StreamingResponseCompletedUpdate` is consumed by `ResponsesTests.cs` at lines 97, 199 and 999, so checking the terminal predicate before deserializing would have stopped that update ever reaching callers.

`9627b5c0` is reverted. The branch is back to `1801ff95`.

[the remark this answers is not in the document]
adityasingh2400: Also correct. `response.failed` and `response.incomplete` are modeled at lines 151 and 153 of the generated deserializer, and neither matched the predicate, so a stream ending on a valid non-success outcome would have run to EOF and thrown.

Reverted along with the other finding. Details and a follow-up diagnosis are in the PR comments.

tests/Assistants/AssistantsMockTests.cs
```diff
@@ -127,4 +127,67 @@ public async Task StreamingRunIgnoresBenignUnknownEvent()
...
+
+            event: thread.run.another_future_event
+            data: {"id":"run_abc","object":"thread.run","status":"in_progress"}
+
+            event: thread.run.completed
+            data: {"id":"run_abc","object":"thread.run","status":"completed"}
+
+            event: done
+            data: [DONE]
+            """);
```
jsquire, line 157: The final SSE event needs an empty-line delimiter. Without it, `SseParser` discards `[DONE]` at EOF and this test does not exercise the terminal branch.

```suggestion
event: done
data: [DONE]
""" + "\n\n");
```

---

:robot: jsquire-copilot

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
...
 - Added an opt-out for SDK telemetry, via the `OpenAI.DisableTelemetry` `AppContext` switch or the `OPENAI_DISABLE_TELEMETRY` environment variable. When enabled, the library sends neither the `X-Stainless-*` headers nor the `User-Agent` header that it would otherwise add. See [Telemetry and privacy](https://github.com/openai/openai-dotnet/blob/main/docs/Observability.md#telemetry-and-privacy) for details.
 
+### Bugs Fixed
+
+- Fixed streaming responses ending early when the service emits an event that the library does not model. The server-sent event enumerator stopped at the first event that produced no updates, so an unrecognized event in the middle of a stream silently terminated the whole stream and looked like a clean, early completion. Unrecognized events are now skipped and every later update still surfaces. This affects all streaming APIs, including `OpenAI.Assistants`, `OpenAI.Chat`, `OpenAI.Responses`, and `OpenAI.Audio`, on both the synchronous and asynchronous paths.
```
jsquire: Only Assistants currently has a multi-value deserializer that can return an empty update sequence. Chat, Audio, and Responses use single-value deserializers.

```suggestion
- Fixed Assistants streaming responses ending early when the service emits an event that the library does not model. Events that deserialize to zero updates are now skipped so later updates continue to surface on synchronous and asynchronous paths.
```

---

:robot: jsquire-copilot

adityasingh2400:
Regression coverage is pushed in `0a6d81ff`, built at the shared utility level exactly as you asked.

`tests/Utility/SseUpdateCollectionTests.cs` drives both collection types over the sequence you specified, modeled A, an unknown event the stub maps to `[]`, modeled B, then the terminal event, and asserts `["A", "B"]` in order for sync and async. The stub returns `[]` only for the unknown event.

I verified the fail-before rather than assuming it. With `OpenAI/src/Utility/SseUpdateCollection.cs` and `AsyncSseUpdateCollection.cs` checked out at `bf850f54`, both tests fail with 1 element instead of 2. With the fix restored, both pass, and the Smoke suite is 674 passed, 0 failed, 5 skipped on net10.0. The full solution builds clean.

One thing worth your review. The collections are `internal` and the library is signed while the test project is not, so `InternalsVisibleTo` is not available here. I followed the pattern the test project already uses for `Telemetry` and `Argument.cs` and linked the sources in:

```xml
<Compile Include="..\OpenAI\src\Utility\SseUpdateCollection.cs" LinkBase="Utility\Shared" />
<Compile Include="..\OpenAI\src\Utility\AsyncSseUpdateCollection.cs" LinkBase="Utility\Shared" />
<Compile Include="..\OpenAI\src\Generated\Internal\ModelSerializationExtensions.cs" LinkBase="Utility\Shared" />
<Compile Include="..\OpenAI\src\Generated\Internal\SerializationFormat.cs" LinkBase="Utility\Shared" />
<Compile Include="..\OpenAI\src\Generated\Internal\TypeFormatters.cs" LinkBase="Utility\Shared" />
```

The last three are only there because the JSON deserializer overloads reference `ModelSerializationExtensions.WireOptions`, which pulls in `TypeFormatters` and `SerializationFormat`. If you would rather not link generated internals into the test assembly, say the word and I will switch to signing the test project instead.

The terminal-event policy and the `NetworkTimeout` work are both still out, per your last two notes. This branch is the one-line fix, the changelog correction, and now this test.

adityasingh2400:
Follow-up on the push above. Adding the `ItemGroup` put the branch in conflict, because `#1263` landed its own new `ItemGroup` in the same spot in `tests/OpenAI.Tests.csproj`.

Rebased onto `acf0212e` and kept both groups. Branch is now `55d35580`, zero commits behind. Rebuilt clean and the Smoke suite is 690 passed, 0 failed, 5 skipped.

OpenAI/src/Utility/SseUpdateCollection.cs
```diff
@@ -164,7 +164,13 @@ public bool MoveNext()
...
             }
 
-            if (_events.MoveNext())
+            // Keep advancing the outer event enumerator instead of stopping at the first
+            // event. An event can legitimately deserialize to zero updates, for example an
+            // unmodeled event that the deserializer maps to an empty sequence. Such an event
+            // must be skipped so that later events still surface, otherwise a single
+            // unrecognized event in the middle of a stream would end the whole stream and
+            // look like a clean, early completion.
+            while (_events.MoveNext())
```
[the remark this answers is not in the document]
adityasingh2400: Confirmed, and fixed in `c9ca3c9a`. This is a regression from my own change. The old `if (_events.MoveNext())` consumed at most one event per call, so cancellation was checked once per event. The `while` loop can consume many events per call, which drops that to once per yielded update.

I added the recheck inside the sync loop and a test that fails without it.

One correction to the scope. I also tried this on the async path and the test passes there without any change, because `CreateEventEnumeratorAsync` passes the token into `GetAsyncEnumerator(_cancellationToken)`, so `MoveNextAsync` throws on its own. I left the async loop alone rather than adding a redundant check, and kept the async test as coverage for that.

adityasingh2400:
`c9ca3c9a` addresses the Codex P2 on cancellation, which was a real regression from my own change and worth calling out since you will be reviewing the loop.

The pre-PR `if (_events.MoveNext())` consumed one event per call, so the check at the top of `MoveNext` ran once per event. My `while` loop can consume an arbitrary run of zero-update events inside a single call, which drops cancellation to once per yielded update. The fix rechecks the token per event, and `SkippingEventsWithNoUpdatesObservesCancellation` fails without it.

The async path did not need the same change. `CreateEventEnumeratorAsync` passes the token into `GetAsyncEnumerator(_cancellationToken)`, so `MoveNextAsync` already throws. I verified that by running the async version of the test against the unmodified loop and it passed, so I left that loop alone rather than adding a redundant check, and kept the test.

Branch is `c9ca3c9a`, zero behind. Solution builds clean, Smoke is 692 passed, 0 failed, 5 skipped.

Separately, both workflow runs are sitting at `action_required` and have been since 07 August, so none of this has been exercised in CI yet. If you approve the run you will see the four new tests.

OpenAI/src/Utility/SseUpdateCollection.cs
```diff
@@ -164,8 +164,18 @@ public bool MoveNext()
...
```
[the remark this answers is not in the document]
adityasingh2400: Correct on both points, fixed in `8eb9ba5a`. The check now sits ahead of the read rather than inside the loop body, and the loop is `while (true)` with an explicit break.

Your note about the test prebuffering was fair, so I wrote one that cannot be satisfied that way. `CancellationDoesNotWaitForTheNextEvent` uses a stream that serves two events and then blocks forever on read. Against the previous commit that test hangs and fails on its 10 second guard. With the check ahead of the read it returns immediately.

OpenAI/src/Utility/AsyncSseUpdateCollection.cs
```diff
@@ -193,7 +193,13 @@ async ValueTask<bool> IAsyncEnumerator<U>.MoveNextAsync()
...
             }
 
-            if (await _events.MoveNextAsync().ConfigureAwait(false))
+            // Keep advancing the outer event enumerator instead of stopping at the first
+            // event. An event can legitimately deserialize to zero updates, for example an
+            // unmodeled event that the deserializer maps to an empty sequence. Such an event
+            // must be skipped so that later events still surface, otherwise a single
+            // unrecognized event in the middle of a stream would end the whole stream and
+            // look like a clean, early completion.
+            while (await _events.MoveNextAsync().ConfigureAwait(false))
```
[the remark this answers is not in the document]
adityasingh2400: You are right and I was wrong. I told the reviewer an hour ago that the async path did not need this, based on a test that only proved the weaker case.

I ran your exact scenario. With a stream of modeled A, then an unknown event that cancels, then a buffered `[DONE]`, enumeration finished normally and produced only `A`, never throwing. My earlier async test passed only because a modeled `B` sat between the cancel and the terminal, which forced a fresh read.

Fixed in `8eb9ba5a` and covered by `CancellationIsNotOvertakenByABufferedTerminalEventAsync`, which fails against the previous commit. I have posted a correction on the PR.

adityasingh2400:
Correction to my comment above, before you spend time on it.

I said the async path did not need the cancellation change. That was wrong, and the test I based it on was too weak. Codex pushed back and it was right.

The case I missed is when the frame after the canceling event is already buffered. With modeled A, then an unknown event that cancels, then `[DONE]`, the async enumeration finishes **normally** and yields only `A`. It never throws, so a caller that cancels mid stream gets a clean completion instead of an `OperationCanceledException`. My earlier test passed only because a modeled `B` sat between the cancel and the terminal event, which forced a fresh read that did observe the token.

The second thing it caught applies to both paths. My check was inside the loop body, so it ran after the blocking read in the loop condition. If the server goes quiet, cancellation is not seen until another event arrives.

`8eb9ba5a` puts the check ahead of the read in both collections and turns each loop into `while (true)` with an explicit break. Three tests now cover this, and I verified each one fails against `c9ca3c9a`:

- `CancellationIsNotOvertakenByABufferedTerminalEventAsync`, which previously completed normally
- `CancellationDoesNotWaitForTheNextEvent`, which uses a stream that serves two events and then blocks forever on read, and previously hung until its 10 second guard
- `CancellationIsNotOvertakenByABufferedTerminalEvent`, the sync counterpart

Seven tests in the file, all passing. Solution builds clean, Smoke is 695 passed, 0 failed, 5 skipped. Branch is `8eb9ba5a`, zero behind.

adityasingh2400:
Status on the terminal-event policy, since I left that thread open after the revert.

I mapped the call sites first. The shared collections are constructed in 16 places, 12 across Chat, Assistants and Audio in `OpenAI/src`, and 4 in `OpenAI.Responses/src/Custom/ResponsesClient.cs`.

The blocker is that the two terminal forms need opposite handling, so a single `Func<SseItem<byte[]>, bool>` does not express both on its own:

- `[DONE]` is a control frame. The loop checks it ahead of the deserializer today and has to keep doing that, because the Chat, Assistants and Audio deserializers do not model it.
- `response.completed` is a modeled event. It has to be deserialized and yielded before the stream ends, since `tests/Responses/ResponsesTests.cs` reads `StreamingResponseCompletedUpdate` at lines 97, 199 and 999. My reverted attempt evaluated the predicate first and swallowed exactly that.

Two shapes fit your sketch, and I would rather you pick than guess a second time:

1. Predicate plus a per-API flag saying whether the terminal event also carries updates.
2. Predicate evaluated after deserialization, with the `[DONE]` skip staying in the shared loop as it is now.

Either one is a small change on top of the current head. Say which and I will push it with coverage for both terminal forms.

jsquire (commented):
3. **Network timeout**

Apologies for the incorrect request. We should have traced this through the pinned `System.ClientModel` implementation before raising the original comment.

`System.ClientModel` 1.14.0 already applies `NetworkTimeout` to streaming response reads through this path.

```text
OpenAIClientOptions.NetworkTimeout
    -> ClientPipeline
    -> PipelineMessage.NetworkTimeout
    -> PipelineTransport
    -> ReadTimeoutStream
    -> SseParser
```

Streaming requests set `BufferResponse = false`. `PipelineTransport` wraps the response body in `ReadTimeoutStream`, which applies and resets the timeout for every synchronous and asynchronous read. The default is 100 seconds unless configured otherwise.

Please consider this request withdrawn. No timeout changes are needed in this PR.

4. **Terminal policy**

Thanks for investigating this. Please keep the terminal-policy work reverted and keep this PR focused on skipping zero-update events.

Responses must surface modeled `response.completed`, `response.failed`, and `response.incomplete` updates. Assistants uses a separate `[DONE]` control frame. A shared completion policy therefore requires broader API-specific design and should not be introduced in this fix.

The existing Assistants test covers the only current path that can return an empty update sequence. A separate utility-level test is not required for this PR.

---

:robot: jsquire-copilot

adityasingh2400:
Thanks for tracing the `System.ClientModel` path, and for withdrawing both requests.

Nothing to change for either. The branch has carried no timeout work at any point, and the terminal-policy attempt has been reverted since `1801ff95`, so `8eb9ba5a` is already the focused form you are asking for. It is the one-line `while` change plus the changelog entry and the tests.

On the utility-level tests, I would rather you decide than assume. You asked for them earlier and I added `tests/Utility/SseUpdateCollectionTests.cs`, seven tests. Three of them cover the cancellation regression my own `while` change introduced, and each fails against `c9ca3c9a`:

- `CancellationIsNotOvertakenByABufferedTerminalEventAsync`, which previously completed normally instead of throwing
- `CancellationDoesNotWaitForTheNextEvent`, which previously hung until its 10 second guard
- `CancellationIsNotOvertakenByABufferedTerminalEvent`, the sync counterpart

The Assistants test covers the zero-update path, but not those three. Happy to drop the file if you want the PR smaller, just say the word and I will push the removal.

jsquire (approved):
Thank you for your contribution, @adityasingh2400, and your willingness to work through feedback.   Looks like we're in good shape now.   Merging.

PR #1277, opened 2026-08-06
Reduce memory pressure for large data URIs

### LINKED ISSUE #1276: Mitigate copies while building chat message parts.
I want to create `ChatMessageContentPart` / `ResponseContentPart` without additional buffer copies (one image can be up to 100mb).

We have 2 methods for both classes:
* `CreateImagePart(Uri imageUri)`
* `CreateImagePart(BinaryData imageBytes)`

Pseudo code:

```cs
var data = new byte[1_000_000];
var stream = new MemoryStream(data, index: 0, count: data.Length, writable: true, publiclyVisible: true);

BinaryData binaryData = default!;

//...

if (stream is MemoryStream memory)
{
    if (memory.TryGetBuffer(out ArraySegment<byte> tempSegment))
    {
        binaryData = new BinaryData(tempSegment, "image/png");
    }
    else
    {
        //...
    }
}

//Copies here
var responsePart = ResponseContentPart.CreateInputImagePart(binaryData, imageDetailLevel: null);
var chatPart = ChatMessageContentPart.CreateImagePart(binaryData, "image/png", imageDetailLevel: null);
```

Copies here due to 
https://github.com/openai/openai-dotnet/blob/765064623431204cd75aaeea04fc823f2444b309/OpenAI/src/Utility/DataEncodingHelpers.cs#L47-L51

Base64 can be created using `public static string ToBase64String(byte[] inArray, int offset, int length);` or `public static string ToBase64String(ReadOnlySpan<byte> bytes`

Example:

```cs
if (stream_.TryGetBuffer(out ArraySegment<byte> tempSegment))
{
    return Convert.ToBase64String(
        tempSegment.Array!,
        tempSegment.Offset + offset,
        count);
}
```

I tried next:
1) `CreateImagePart(Uri imageUri)` - impossible due to size limitations ('Invalid URI: The Uri string is too long.'), parse overhead etc.
2) override `JsonModelWriteCore` in `ChatMessageContentPart` - impossible due to 0 public constructors in `ChatMessageContentPart`

Solutions:

1) Still take `BinaryData`, but try use `ToMemory()` or `ToStream()`.
2) Allow to create part from base64 `String` or  `Stream` or `Span`.

jsquire:
# Summary
 
 Reduce memory pressure when creating Base64 data URIs from large image and file payloads.
 
 This addresses the unnecessary buffer copy reported in #1276. Previously, `BinaryData.ToArray()` copied the entire payload before Base64 encoding, which could add significant memory pressure for images approaching the supported size limit.
 
 ## Changes
 
 - Use span-based Base64 encoding on .NET 8 and later, avoiding the intermediate byte-array allocation.
 - Reuse array-backed `BinaryData` directly on .NET Standard 2.0 while preserving the original array offset and count.
 - Retain the existing copy-based fallback for non-array-backed memory on .NET Standard 2.0.
 - Keep the optimization isolated to `DataEncodingHelpers` without changing public content-part APIs or serialization behavior.
 - Update serialization coverage to verify that sliced memory encodes only the intended bytes.
 
 ## Validation
 
 - Built the library for `netstandard2.0`, `net8.0`, and `net10.0`.
 - Ran the focused Responses regression test on `net8.0`, `net9.0`, and `net10.0`.
 - Ran the existing Chat image-content serialization tests on `net8.0`.

### DIFF
--- a/OpenAI/src/Utility/DataEncodingHelpers.cs
+++ b/OpenAI/src/Utility/DataEncodingHelpers.cs
@@ -1,6 +1,9 @@
 using System;
-using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
+
+#if !NET8_0_OR_GREATER
+using System.Runtime.InteropServices;
+#endif
 
 #nullable enable
 
--- a/OpenAI/src/Utility/DataEncodingHelpers.cs
+++ b/OpenAI/src/Utility/DataEncodingHelpers.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 
 #nullable enable
@@ -46,7 +47,24 @@ internal static partial class DataEncodingHelpers
 
     public static string CreateDataUri(BinaryData bytes, string bytesMediaType)
     {
-        string base64Bytes = Convert.ToBase64String(bytes.ToArray());
+        ReadOnlyMemory<byte> memory = bytes.ToMemory();
+#if NET8_0_OR_GREATER
+        string base64Bytes = Convert.ToBase64String(memory.Span);
+#else
+        string base64Bytes;
+        if (MemoryMarshal.TryGetArray(memory, out ArraySegment<byte> segment)
+            && segment.Array is not null)
+        {
+            base64Bytes = Convert.ToBase64String(
+                segment.Array,
+                segment.Offset,
+                segment.Count);
+        }
+        else
+        {
+            base64Bytes = Convert.ToBase64String(memory.ToArray());
+        }
+#endif
         return $"data:{bytesMediaType};base64,{base64Bytes}";
     }
 }
--- a/tests/Responses/ResponsesSmokeTests.cs
+++ b/tests/Responses/ResponsesSmokeTests.cs
@@ -297,7 +297,10 @@ public partial class ResponsesSmokeTests
         }
 
         string imageMediaType = "image/png";
-        BinaryData imageBytes = BinaryData.FromBytes(Encoding.UTF8.GetBytes("image data"), imageMediaType);
+        byte[] imageBuffer = Encoding.UTF8.GetBytes("prefiximage datasuffix");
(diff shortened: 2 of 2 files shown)


tests/Responses/ResponsesSmokeTests.cs
```diff
@@ -297,7 +297,10 @@ static void AssertExpectedImagePart(ResponseContentPart filePart)
         }
 
         string imageMediaType = "image/png";
-        BinaryData imageBytes = BinaryData.FromBytes(Encoding.UTF8.GetBytes("image data"), imageMediaType);
+        byte[] imageBuffer = Encoding.UTF8.GetBytes("prefiximage datasuffix");
+        BinaryData imageBytes = BinaryData.FromBytes(
+            imageBuffer.AsMemory("prefix".Length, "image data".Length),
+            imageMediaType);
```
joseharriaga, line 303: Q: What's the intention for this test change?

jsquire: The intent is to exercise `BinaryData` backed by a sliced `ReadOnlyMemory<byte>` with a nonzero offset and bounded count. The .NET Standard path passes the underlying array, offset, and count to `Convert.ToBase64String`, so this verifies that the prefix and suffix are not accidentally encoded and that the public Responses serialization behavior remains unchanged.

OpenAI/src/Utility/DataEncodingHelpers.cs
```diff
@@ -46,7 +50,24 @@ public static bool TryParseDataUri(string dataUri, out BinaryData? bytes, out st
...
+            && segment.Array is not null)
+        {
+            base64Bytes = Convert.ToBase64String(
+                segment.Array,
+                segment.Offset,
+                segment.Count);
+        }
+        else
+        {
+            base64Bytes = Convert.ToBase64String(memory.ToArray());
```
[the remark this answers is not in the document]
jsquire: This is definitively and provably incorrect.

joseharriaga (approved):

PR #1278, opened 2026-08-07
[DO NOT MERGE] Prototype design for automated REST specification sync

jsquire:
> **Draft / prototype for handoff.** This is not intended to merge as-is. It is a working
> end-to-end implementation meant to be driven to completion by someone else, and the design
> documents under `specification/tools/design/` are included for that purpose rather than as
> repository documentation.

## What this does

Adds a scheduled workflow that, on the second of each month, snapshots the OpenAI REST API
specification, splits it into the 24 per-feature specifications the SDK is organized around,
compares it against the previous snapshot, and opens a pull request with an AI-authored explanation
of what changed and what it likely means for our TypeSpec.

The intent is that a maintainer opens a PR that has already done the reading, so the question shifts
from "did anything change upstream, and where?" to "do I agree with this read, and which of these do
we act on?".

## How it is put together

Everything that must be exact and repeatable is deterministic code. Everything that benefits from
judgment is an agent with no write access.

| Piece | What it is |
|---|---|
| `specification/tools/OpenAI.SpecProcessor` | .NET console tool: sanitize, clean, split, validate, diff, report. Excluded from `OpenAI.slnx`, so it cannot enter the SDK build or ship in a package. |
| `scripts/Sync-OpenAISpec.ps1` | The orchestrator. CI and a laptop run the same script, so there is no drift between how the workflow runs it and how someone tests it. |
| `.github/workflows/openai-spec-sync.md` | The gh-aw workflow. Deterministic steps run first; the agent gets a finished report and produces only prose. |
| `specification/openai/{current,previous}` | The seeded snapshot pair, which is the entire comparison mechanism. |
| `specification/tools/design/` | The design documents. Start with `openai-spec-sync-design.md`. |

## Two decisions worth your attention

**A run happens when either the source content or the processor's own behavior has changed.**
Comparing source bytes alone answers "did upstream change" while quietly assuming "we would produce
the same thing anyway", which stops being true the month someone edits the feature map. A processing
identity fingerprint, derived from the feature map rather than hand-maintained, closes that.

**Publishing is a separate decision from running.** The input check decides whether to run; a
comparison of the generated output decides whether to publish. So a forced diagnostic run cannot
open a PR that moves nothing but a timestamp, and an upstream edit landing entirely in an excluded
path correctly produces nothing.

## Why the diff is large

This first PR carries the seeded `current` and `previous` snapshots, which is roughly 125k lines of
generated YAML. Subsequent PRs from the workflow will be small. Committing both sides is deliberate:
it makes the baseline a reviewed artifact rather than a moving target, and it is why byte-stability
is a hard requirement rather than a nicety.

## State

- 111 tests passing (`specification/tools/OpenAI.SpecProcessor.Tests`), run by `spec-tooling-checks.yml`
- `gh aw compile --strict`: 0 errors, 0 warnings; lock file drift is caught by `agentic-lockfile-check.yml`
- Verified locally: the no-op path, the identity-mismatch path, forced-run determinism, dry-run
  cleanup on success and failure, and that a validation failure leaves both snapshots untouched

## What is left

1. Confirm the org has centralized Copilot billing so `permissions: copilot-requests: write` works.
   If not, the fallback is a `COPILOT_GITHUB_TOKEN` secret.

2. Confirm "Allow GitHub Actions to create and approve pull requests" is enabled. If it is not,
   gh-aw silently degrades to opening an issue instead, which looks like success until someone
   notices there is no PR.

3. Run the workflow with `dry_run=true` for an artifact, then `dry_run=false` for a real PR, before
   letting the schedule run unattended.

4. Decide whether `/content_provenance_checks` gets a feature area. It is deliberately unassigned
   today and is reported every month as a known gap.

5. Decide what to do with `specification/tools/design/`, which should probably not merge in its
   current form.


jsquire:
Opened in error by an automated assistant without authorization. Closing immediately. Please disregard.

PR #1285, opened 2026-08-12
Add versioned API listings for easy diffs

joseharriaga:
- Created an archival copy of the API listings corresponding to version 2.13.0 and added them under `./api/released/`
- Moved the generated API listings under `./api/in-progress/` and modified the `Export-Api.ps1` script accordingly
- Updated documentation


jsquire (commented):
I really don't like the approach of renaming a directory for each version.   We don't want to accumulate redundant copies for history there, lest we bloat the repository.   I also really hate the name `wip`.   

What I'd propose:

- Call them `published` and ( `local` or `working` or `in-progress`)
- Add a metadata file in the `api` directory that tells us what published version is current

scripts/Export-Api.ps1
```diff
@@ -6,16 +6,16 @@
     This script invokes the MSBuild GenerateApi target to produce C# source files
     representing the public API contract of the OpenAI library. GenAPI writes one
     intermediate file per target framework, and this script splits those files by
-    namespace into target framework subdirectories under the 'api' folder at the
-    repository root.
+    namespace into target framework subdirectories under the 'api/wip' folder at
+    the repository root.
```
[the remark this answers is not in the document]
joseharriaga: Fixed.

api/in-progress/net10.0/OpenAI.Audio.net10.0.cs
jsquire, line 1: This feels like a good opportunity to stop keeping all TFMs and just export a baseline.   The pipelines that run API Compat will extract the current target and check against that, so we'll have coverage so long as long as we parameterize the script.   There's really no value in redundant copies because we force all TFMs to have the same public API.     Its just bloat.

api/README.md
```diff
@@ -1,13 +1,15 @@
...
 
 ```powershell
 ./scripts/Export-Api.ps1
 ```
 
-The most common reason to regenerate these files is a change to the public API surface (adding, removing, or renaming public types or members). In that case, modify the relevant source files first and then re-run the script above.
+The most common reason to regenerate the API listings is to reflect a change in the public API surface (such as adding, removing, or renaming public types or members). In that case, modify the relevant source files first and then re-run the script above.
 
-API listings are split by target framework and namespace. For example, the `net10.0` listings are written to `api/net10.0/` as files such as `OpenAI.net10.0.cs`, `OpenAI.Assistants.net10.0.cs`, and `OpenAI.Chat.net10.0.cs`.
+Note that the API listings under `api/released/` reflect the latest released public API surface, which corresponds to version `2.13.0`.
```
jsquire: This seems like it's going to rot over time.   I think we'd be better off with a localized `metadata` file (json, md, txt, whatever) that tracks the current release version association.   The README is easy to overlook.

joseharriaga: @copilot Put the released API version in its `api-version.txt` file and modify this sentence in the api/README.md to point to it.

api/in-progress/net10.0/OpenAI.Audio.net10.0.cs
joseharriaga: I will do this in a separate PR.

jsquire (approved):

api/released/net10.0/OpenAI.Responses.net10.0.cs
```diff
@@ -0,0 +1,1979 @@
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override readonly int GetHashCode();
+        public static bool operator ==(ResponsePromptCacheRetentionPolicy left, ResponsePromptCacheRetentionPolicy right);
+        public static implicit operator ResponsePromptCacheRetentionPolicy(string value);
+        public static implicit operator ResponsePromptCacheRetentionPolicy?(string value);
+        public static bool operator !=(ResponsePromptCacheRetentionPolicy left, ResponsePromptCacheRetentionPolicy right);
+        public override readonly string ToString();
+    }
+    [Experimental("OPENAI001")]
+    public readonly partial struct ResponseReasoningContext : IEquatable<ResponseReasoningContext> {
```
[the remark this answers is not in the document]
joseharriaga: This was due to a bad auto merge. Fixed!

scripts/Export-Api.ps1
```diff
@@ -159,7 +159,7 @@ $configuration = "Release"
 
 # Resolve paths
 $repoRootPath = Join-Path $PSScriptRoot ".." -Resolve
-$outputDirectory = Join-Path $repoRootPath "api"
+$outputDirectory = Join-Path $repoRootPath "api" "in-progress"
```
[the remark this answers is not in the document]
joseharriaga: Automating release-time copy will be done in a separate PR.

Issue #1269, opened 2026-08-04
Adopt new streaming API pattern


joseharriaga:
- Start with the Responses API.
- The objective is to ensure that the public API reflects the new pattern. This initial implementation will need to be done in custom code while the generated approach is finalized.
- Must eventually be applied to all streaming APIs that are currently marked as Experimental.

PR #1287, opened 2026-08-12
Adopt new streaming API pattern 

MaiLinhP:
Addresses https://github.com/openai/openai-dotnet/issues/1269

This branch upgrades the TypeSpec code generator to `1.0.0-alpha.20260821.3` (with `@typespec/http` 1.15, `sdk-core` 0.71.2) and migrates all SSE streaming for Audio, Images, and Responses from `AsyncCollectionResult<T>`/`CollectionResult<T>` to `AsyncStreamingClientResult<T>` backed by `SseStreamingClientResult`. 

It splits the `createSpeech`, `createImage`, and `createImageEdit` TSP operations into separate non-streaming and streaming variants using `@sharedRoute`, so the generator produces distinct protocol methods — preserving `ClientResult` methods for non-streaming and creating `AsyncStreamingClientResult<SseItem<BinaryData>>` methods for streaming. 

An ExperimentalAttributeVisitor addition applies `SCME0005` to methods returning the new streaming type, while preserving existing `OPENAI001` diagnostics. For methods where both experimental flags are needed, new experimental code will be applied in upcoming PR. 

The generator upgrade also introduces `*At` → `*On` and `Url` → `Uri` naming conventions; stable public types (OpenAIFile, ChatCompletion, etc.) preserve their original names via `[CodeGenMember]` wrappers, while experimental types adopt the new names directly.

Tests are migrated to new types, and redundant sync tests removed.


OpenAI/src/Custom/Chat/ChatClient.cs
```diff
@@ -263,24 +268,27 @@ internal AsyncCollectionResult<StreamingChatCompletionUpdate> CompleteChatStream
...
         options ??= new();
         var clonedOptions = CreateChatCompletionOptions(messages, options, stream: true);
 
         using BinaryContent content = clonedOptions.ToBinaryContent();
-        return new SseUpdateCollection<StreamingChatCompletionUpdate>(
-            () => CompleteChat(content, cancellationToken.ToRequestOptions(streaming: true)),
+        using PipelineMessage message = CreateCompleteChatRequest(content, cancellationToken.ToRequestOptions(streaming: true));
+        message.BufferResponse = false;
+        PipelineResponse response = Pipeline.ProcessMessage(message, cancellationToken.ToRequestOptions(streaming: true));
+        return SseStreamingClientResult.Create(
```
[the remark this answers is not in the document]
MaiLinhP: Fixed

OpenAI/src/Custom/Chat/ChatClient.cs
```diff
@@ -252,8 +254,11 @@ internal AsyncCollectionResult<StreamingChatCompletionUpdate> CompleteChatStream
         var clonedOptions = CreateChatCompletionOptions(messages, options, stream: true);
 
         using BinaryContent content = clonedOptions.ToBinaryContent();
-        return new AsyncSseUpdateCollection<StreamingChatCompletionUpdate>(
-            async () => await CompleteChatAsync(content, requestOptions).ConfigureAwait(false),
+        using PipelineMessage message = CreateCompleteChatRequest(content, requestOptions);
+        message.BufferResponse = false;
+        PipelineResponse response = await Pipeline.ProcessMessageAsync(message, requestOptions).ConfigureAwait(false);
```
jsquire: Could we preserve the standard response validation before creating the streaming result? `Pipeline.ProcessMessageAsync` returns a raw `PipelineResponse`, and `SseStreamingClientResult.Create` currently wraps it without calling `ClientResult.FromResponse`. For a 401, 4xx, or 5xx response, enumeration can attempt to parse the error body as SSE instead of throwing `ClientResultException`, unlike the non-streaming methods. Please validate the response first, then pass it to the streaming adapter.

---
:robot: jsquire-copilot

OpenAI/src/Custom/Audio/AudioClient.cs
```diff
@@ -155,8 +156,9 @@ public virtual async Task<ClientResult<BinaryData>> GenerateSpeechAsync(string t
         CreateSpeechGenerationOptions(text, voice, ref options);
 
         using BinaryContent content = options.ToBinaryContent();
-        ClientResult result = await GenerateSpeechAsync(content, cancellationToken.ToRequestOptions()).ConfigureAwait(false);
-        return ClientResult.FromValue(result.GetRawResponse().Content, result.GetRawResponse());
+        using PipelineMessage message = CreateGenerateSpeechRequest(content, cancellationToken.ToRequestOptions());
+        PipelineResponse response = await Pipeline.ProcessMessageAsync(message, cancellationToken.ToRequestOptions()).ConfigureAwait(false);
+        return ClientResult.FromValue(response.Content, response);
```
jsquire, line 164: Could we keep the `ClientResult.FromResponse` validation in this path before creating the value result? The previous implementation obtained a `ClientResult` from the generated method, which performed the standard status validation. This direct `FromValue` call can treat a non-success response as a successful result or surface a deserialization error instead of `ClientResultException`. The same pattern appears in the synchronous speech methods and the transcription methods.

---
:robot: jsquire-copilot

OpenAI/src/Custom/Chat/ChatClient.cs
```diff
@@ -226,20 +228,20 @@ public virtual ClientResult<ChatCompletion> CompleteChat(params ChatMessage[] me
...
+    ///     <see cref="AsyncStreamingClientResult{T}"/> implements the <see cref="IAsyncEnumerable{T}"/> interface and can be
     ///     enumerated over using the <c>await foreach</c> pattern.
     /// </remarks>
     /// <param name="messages"> The messages comprising the chat so far. </param>
     /// <param name="options"> The options to configure the chat completion. </param>
     /// <param name="cancellationToken"> A token that can be used to cancel this method call. </param>
     /// <exception cref="ArgumentNullException"> <paramref name="messages"/> is null. </exception>
     /// <exception cref="ArgumentException"> <paramref name="messages"/> is an empty collection, and was expected to be non-empty. </exception>
-    public virtual AsyncCollectionResult<StreamingChatCompletionUpdate> CompleteChatStreamingAsync(IEnumerable<ChatMessage> messages, ChatCompletionOptions options = null, CancellationToken cancellationToken = default)
+    public virtual Task<AsyncStreamingClientResult<StreamingChatCompletionUpdate>> CompleteChatStreamingAsync(IEnumerable<ChatMessage> messages, ChatCompletionOptions options = null, CancellationToken cancellationToken = default)
```
jsquire: Could we add a streaming-specific `Experimental` code to the public streaming members? `AsyncStreamingClientResult<T>` is marked experimental by `System.ClientModel`, and the file-level pragma only suppresses the warning while compiling this assembly. It does not provide a targeted opt-in for consumers. The Audio streaming members already carry an `Experimental` attribute, so the Chat, Assistants, and Responses streaming members should follow the same pattern with a code dedicated to streaming APIs rather than `OPENAI001`.

---
:robot: jsquire-copilot

jsquire (commented):
Could we also update the streaming documentation to use the new result type names? The chat section still describes `CollectionResult<T>` and `AsyncCollectionResult<T>`, and the Assistants section has the same stale terminology. Both APIs now return `AsyncStreamingClientResult<T>` and use `await foreach`.

---
:robot: jsquire-copilot

PR #1290, opened 2026-08-13
Use the public dotnet package feed

jsquire:
# Summary

Use the public `dotnet-public` Azure Artifacts feed for packages currently mapped to the `nuget` source, while retaining the existing dedicated feeds and deterministic package source mappings.

This replaces the earlier local `microsoft.nuget.config` override approach. That approach did not cover all supported restore experiences because Visual Studio intentionally ignores project-level `RestoreConfigFile`, and `dotnet tool restore` does not evaluate `Directory.Build.props`.

## Rationale

Microsoft guidance states that public `nuget.config` files should use the well-known public `dotnet-public` feed rather than internal quarantine feeds. The feed is anonymously accessible outside Microsoft, so the same tracked configuration works for Microsoft internal contributors, community contributors, and CI.

The repository continues to use `<clear />`. This prevents machine-level and user-level package sources from changing restore behavior and ensures that package source mapping remains deterministic. Removing `<clear />` would not provide an ordered fallback because NuGet queries eligible sources in parallel rather than trying one source only after another fails.

NuGet.org is therefore not added as a package fallback. An internal contributor whose network blocks `api.nuget.org` could still encounter restore failures if it were configured, while mapping packages to both feeds would make their origin less deterministic.

The existing source key remains `nuget`, so the current wildcard package source mapping continues to apply without changes.

### Microsoft internal contributors

Visual Studio, `dotnet restore`, MSBuild restore, and `dotnet tool restore` use the tracked repository configuration without local files or command-line arguments. Packages mapped to `nuget` are restored from the Microsoft-accessible public `dotnet-public` feed. No internal feed URL, credentials, or machine-specific setup is committed to the repository.

### Non-Microsoft contributors

The `dotnet-public` feed is publicly and anonymously accessible. Community contributors and external CI use the same repository configuration without Microsoft credentials or local setup.

## Package availability

Cold restores were performed with empty isolated global package and HTTP caches. All current dependencies restored successfully.

The test downloaded 92 package and version entries across 82 package IDs. Of those entries, 88 came from `dotnet-public` and four TypeSpec and code-generation packages came from `azure-sdk-for-net`. No package was obtained from a machine-level source or NuGet.org.

This verifies coverage for the repository's current main solution, codegen solution, and local tool manifest. It does not establish that `dotnet-public` contains every package and version available on NuGet.org. Future dependency additions must be available from the source selected by package source mapping before they can be adopted.

## NuGet auditing

`dotnet-public` does not advertise a NuGet vulnerability data resource. The configuration adds `https://data.nuget.org/v3/index.json` under `auditSources`. This endpoint provides vulnerability metadata without using `api.nuget.org` as a package source.

## Changes

- Change the `nuget` package source URL from `api.nuget.org` to the public `dotnet-public` feed.
- Add the NuGet vulnerability-only endpoint under `auditSources`.
- Remove the conditional `RestoreConfigFile` properties from both `Directory.Build.props` files.
- Remove the obsolete `/microsoft.nuget.config` ignore entry.
- Retain `<clear />`, the existing source keys, and all package source mappings.

## Validation

- Restored `OpenAI.slnx` with empty isolated package and HTTP caches.
- Restored `codegen/generator/OpenAI.Library.Plugin.slnx` with empty isolated package and HTTP caches.
- Restored the repository tool manifest with empty isolated package and HTTP caches.
- Confirmed all current package versions were supplied by the three tracked public feeds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,7 +153,6 @@ UpgradeLog*.XML
 packages
 packages/repositories.config
 testPackages
-/microsoft.nuget.config
 
 # Mac development
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -153,6 +153,7 @@ UpgradeLog*.XML
 packages
 packages/repositories.config
 testPackages
+/microsoft.nuget.config
 
 # Mac development
 .DS_Store
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,7 +3,6 @@
   <!-- Custom Properties -->
   <PropertyGroup>
     <RepositoryRoot>$(MSBuildThisFileDirectory)</RepositoryRoot>
-    <RestoreConfigFile Condition="Exists('$(MSBuildThisFileDirectory)microsoft.nuget.config')">$(MSBuildThisFileDirectory)microsoft.nuget.config</RestoreConfigFile>
 
     <ClientTargetFrameworks>net10.0;net8.0;netstandard2.0</ClientTargetFrameworks>
     <TestTargetFrameworks>net10.0;net9.0;net8.0</TestTargetFrameworks>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,6 +3,7 @@
   <!-- Custom Properties -->
   <PropertyGroup>
     <RepositoryRoot>$(MSBuildThisFileDirectory)</RepositoryRoot>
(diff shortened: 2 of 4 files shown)


jsquire:
Exemptions were approved to mitigate the issues here.  Closing.

PR #1292, opened 2026-08-15
feat: add X.509 workload identity federation for HTTP clients

jbeckwith-oai:
## What and why

Implements the HTTP-only first phase of [X.509 Workload Identity Federation in OpenAI SDKs: Design Proposal](https://app.notion.com/p/openai/X-509-Workload-Identity-Federation-in-OpenAI-SDKs-Design-Proposal-3bc8e50b62b0814a9c36c3cf6d897f04). Applications can exchange the X.509 identity already configured on their native .NET HTTP handler for short-lived bearer tokens without introducing SDK-managed certificate, key, trust, proxy, or rotation configuration.

The exchange is lazy, cached against a monotonic clock, refreshed through a cancellation-aware `SemaphoreSlim` single-flight, bounded by a half-TTL refresh-buffer clamp, retried for bounded transient failures with `Retry-After`, and invalidated/replayed at most once on HTTP 401 for known replayable request content only.

## Public API and .NET idioms

```csharp
using SocketsHttpHandler handler = new() { AllowAutoRedirect = false };
handler.SslOptions.ClientCertificateContext =
    SslStreamCertificateContext.Create(clientCertificate, intermediates, offline: true);

using X509WorkloadIdentityCredential credential = new(
    identityProviderId,
    serviceAccountId,
    new X509WorkloadIdentityCredentialOptions
    {
        Handler = handler,
        RefreshBuffer = TimeSpan.FromMinutes(5),
    });

OpenAIClient client = new(credential);
```

- Add typed `OpenAIClient(X509WorkloadIdentityCredential)` and `OpenAIClient(X509WorkloadIdentityCredential, OpenAIClientOptions)` overloads; child HTTP clients reuse the existing client pipeline and credential cache.
- Follow neighboring SDK constructor overloads, mutable options, experimental credential annotations, `AuthenticationPolicy`, sync/async pipeline parity, `CancellationToken`, `SemaphoreSlim`, and conventional `IDisposable` ownership.
- Use `TimeProvider.System` on modern TFMs and the idiomatic `Stopwatch` monotonic fallback on `netstandard2.0`; new production files explicitly enable nullable analysis.
- Document long-lived DI registration, group certificate rotation, and an application-owned `OPENAI_AUTH_MODE=api_key|x509` rollout example.

## Security and transport ownership

- The **only** transport input is a supported, directly configured `HttpClientHandler` or `SocketsHttpHandler`; custom/delegating/subclass handlers and separately configured SDK transports are rejected.
- Redirect refusal is verified on the same handler used to construct exactly one SDK-owned `HttpClient`, rechecked on every sync/async token acquisition including cache hits and immediately before exchange, and therefore applies to both token exchange and HTTP API requests. No unrelated `HttpClient` + handler pair is accepted.
- The exact shared `HttpClient` backs both the immutable `POST https://mtls.auth.openai.com/oauth/token` exchange and `HttpClientPipelineTransport`; no public exchange-URL override exists.
- Exchange JSON contains provider/service-account identifiers, the OAuth token-exchange grant, and the X.509 token type; it structurally omits `subject_token`.
- The SDK owns/disposes only the `HttpClient` wrapper (`disposeHandler: false`). The caller owns the handler, certificate/private key, server trust, proxy, HSM integration, connection pool, and rotation lifecycle.
- Error reads are capped and token/certificate/provider error material is not surfaced; every rejected bearer token is invalidated, including failed one-time replays, while non-replayable custom content is never resent. Exhausted token-endpoint network failures become non-retryable exchange errors so outer pipeline retries cannot multiply the three-attempt bound.
- Only X.509-mode clients default to `https://mtls.api.openai.com/v1`; existing API-key clients and caller-configured API-key transports retain existing behavior.

## Validation matrix

- Focused TLS/authentication suite: **44 passed on .NET 8, 44 passed on .NET 9, and 44 passed on .NET 10** under the exact CI playback-mode environment; coverage includes exact exchange host/path/body and omitted subject token, redirect refusal/mutation and cached-handler immutability, the same real client certificate on both TLS legs, private-key/handler ownership after disposal, sync/async, single-flight, waiter cancellation, expiry/refresh clamping, bounded transient and network retries, `Retry-After`, reusable frozen client options, replayable/non-replayable 401 handling and replacement-token invalidation, error redaction, child pipeline sharing, Realtime exclusion, and API-key/custom-transport regressions. The fixture follows the adjacent mutual-TLS fixture's `[NonParallelizable]` convention to avoid starving unrelated globally parallel playback tests.
- Full SDK suite: **.NET 8: 1,608 passed / 87 existing skips; .NET 9: 1,620 passed / 87 existing skips; .NET 10: 1,622 passed / 87 existing skips.**
- Independent packaged-consumer validation: built the `.nupkg` from the exact head and restored it into a clean standalone project with no SDK project references; **seven real ephemeral-certificate mTLS/transport scenarios passed independently on each of .NET 8, 9, and 10**. Exercised the public model client, exact token endpoint/body, certificate presentation on both TLS legs, bearer injection, cache/single-flight, reusable frozen options, TTL/401 refresh, caller cancellation, `Retry-After`, error redaction, redirect rejection, bounded network failures under pipeline retries, and caller-owned handler/private-key survival.
- Full clean TypeSpec/code-generation integrity job: `scripts/Invoke-CodeGen.ps1 -Clean`; generated files unchanged and experimental public-API validation passed.
- Generator visitor suite: **10 passed**.
- Public API generation: `scripts/Export-Api.ps1`; `netstandard2.0`, `net8.0`, and `net10.0` exports verified byte-for-byte reproducible.
- Documentation integrity: `scripts/Update-Snippets.ps1`; generated snippets unchanged.
- Full solution Release build with current .NET analyzers: **0 warnings, 0 errors** across supported target frameworks.
- `dotnet format OpenAI.slnx --no-restore --verify-no-changes --include <changed C# files>` and `git diff --check`: passed.
- Release package and symbols package produced successfully; standalone examples project Release build passed.
- Strict thermo-nuclear maintainability review of the exact final diff: passed; mode selection remains centralized, no unverifiable transport association or thin wrapper abstractions, and every changed source/test file remains below 1,000 lines.

## Compatibility, scope, and residual risks

This is additive to existing stable client constructors. Existing API-key authentication, API-key mTLS, custom `AuthenticationPolicy`, custom transports outside X.509 mode, and HTTP child clients retain their behavior. This repository does not currently contain a JWT/ID-token workload-identity foundation, so there is no pre-existing JWT/ID-token WIF implementation to modify or regress.

Realtime/WebSockets are explicitly **out of scope for Phase 1** and fail closed for X.509-authenticated clients; they require a separate Phase 2 design. Applications must retain the caller-owned handler for the credential/client lifetime and rotate the handler, credential, and clients together. Clean code generation reports the repository's existing TypeSpec warnings and one pre-existing npm audit advisory in its locked generator dependencies; neither is introduced or modified here.

Validated against upstream `main` at `1b0bb63beaa2543c7be8a035b038a7e459df4415`; feature head `8e73d66f476b86c4908a66d9497a6323473bceb9`.


OpenAI/src/Utility/WorkloadIdentityAuthenticationPolicy.cs
```diff
@@ -0,0 +1,78 @@
...
+        int currentIndex)
+    {
+        string token = await _credential.GetTokenAsync(message.CancellationToken).ConfigureAwait(false);
+        message.Request.Headers.Set("Authorization", "Bearer " + token);
+        await ProcessNextAsync(message, pipeline, currentIndex).ConfigureAwait(false);
+
+        if (ShouldReplay(message))
+        {
+            message.SetProperty(typeof(WorkloadIdentityAuthenticationPolicy), s_replayMarker);
+            _credential.Invalidate(token);
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d93bb5bb31eef1d2d0c44b66aecf4138bd224ea8: both synchronous and asynchronous authentication policies now invalidate the rejected cached token for every HTTP 401 before independently deciding whether the original content is safe to replay. `UnauthorizedNonReplayableRequestInvalidatesTokenForNextRequest` proves the first non-replayable body is sent exactly once, the next independent non-replayable request exchanges a fresh token, and that request succeeds. The focused 41-test suite passes on .NET 8, 9, and 10, and the clean packaged-client real-mTLS consumer independently verifies 401 refresh.

tests/MutualTls/X509WorkloadIdentityTests.cs
```diff
@@ -0,0 +1,825 @@
...
+
+    [Test]
+    public void ClientRejectsIndependentTransport()
+    {
+        using HttpClientHandler handler = new() { AllowAutoRedirect = false };
+        using X509WorkloadIdentityCredential credential = CreateCredential(handler);
+
+        Assert.Throws<ArgumentException>(() => new OpenAIClient(credential, new()
+        {
+            Transport = new HttpClientPipelineTransport(),
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d93bb5bb31eef1d2d0c44b66aecf4138bd224ea8: `ClientRejectsIndependentTransport` now creates the rejected transport with `using HttpClientPipelineTransport transport = new();` and passes that same variable to `OpenAIClientOptions`, so the transport is disposed even when construction throws. Verified by the focused and full .NET 8/9/10 suites and the analyzer-enabled zero-warning solution build.

tests/MutualTls/X509WorkloadIdentityTests.cs
```diff
@@ -0,0 +1,825 @@
...
+                UseProxy = false,
+                SslOptions = new SslClientAuthenticationOptions
+                {
+                    RemoteCertificateValidationCallback = (_, certificate, _, _) =>
+                        certificate is not null
+                        && certificate.GetCertHashString() == _certificate.GetCertHashString(),
+                },
+                ConnectCallback = async (_, cancellationToken) =>
+                {
+                    TcpClient client = new();
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d93bb5bb31eef1d2d0c44b66aecf4138bd224ea8: the test handler's `ConnectCallback` now wraps `ConnectAsync` and `GetStream()` in `try/catch`, disposes its `TcpClient` on connection/setup failure, and preserves the existing transferred stream ownership on success. Verified by all 41 real-TLS tests on .NET 8, 9, and 10 plus the full runtime matrix.

examples/MutualTls/Example02_X509WorkloadIdentityAsync.cs
```diff
@@ -0,0 +1,63 @@
...
+                ? new OpenAIClient(credential)
+                : new OpenAIClient(credential, new() { Endpoint = GetEndpoint(endpoint) });
+
+            ChatCompletion completion = await client.GetChatClient("gpt-4o-mini")
+                .CompleteChatAsync("Reply with exactly: X.509 workload identity succeeded.");
+            Console.WriteLine(completion.Content[0].Text);
+        }
+        finally
+        {
+            foreach (X509Certificate2 certificate in certificates)
```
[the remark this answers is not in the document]
jbeckwith-oai: Investigated against the actual .NET types and the adjacent `Example01_MutualTlsAsync`: `X509Certificate2Collection` does not implement `IDisposable`, and `X509Certificate2[]` also cannot be used in a `using` declaration. Every certificate returned from the PFX must remain alive together for the complete TLS/API operation and then be disposed individually, so the existing `try/finally` plus certificate loop is the correct deterministic ownership pattern; the neighboring mTLS example uses the identical pattern. Replacing it with `using` on an array would not compile, while disposing individual certificates before the request would break the TLS chain. Kept the existing idiomatic implementation and independently verified real certificate presentation, private-key ownership, and handler lifetime from a clean packaged consumer on .NET 8/9/10.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,413 @@
...
+    internal HttpClientPipelineTransport Transport { get; }
+
+    internal string GetToken(CancellationToken cancellationToken)
+    {
+        ThrowIfDisposed();
+        CachedToken? token = Volatile.Read(ref _cachedToken);
+        if (token is not null && IsFresh(token))
+        {
+            return token.Value;
+        }
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d93bb5bb31eef1d2d0c44b66aecf4138bd224ea8: synchronous `GetToken` now calls `ValidateHandler(_handler)` before inspecting or returning any cached token, while retaining the immediately-before-exchange validation. `UsedHandlerCannotEnableRedirectsWhileTokenIsCached` additionally proves native .NET freezes `AllowAutoRedirect` after first use and exercises both cached synchronous and asynchronous API requests. All focused/full runtime suites and packaged real-mTLS scenarios pass.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    internal async ValueTask<string> GetTokenAsync(CancellationToken cancellationToken)
+    {
+        ThrowIfDisposed();
+        CachedToken? token = Volatile.Read(ref _cachedToken);
+        if (token is not null && IsFresh(token))
+        {
+            return token.Value;
+        }
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d93bb5bb31eef1d2d0c44b66aecf4138bd224ea8: asynchronous `GetTokenAsync` now calls `ValidateHandler(_handler)` before inspecting or returning any cached token, while retaining the immediately-before-exchange validation. The new cached-handler regression exercises both async and sync cache hits and confirms redirect settings remain disabled after the handler has entered use. Verified on .NET 8, 9, and 10, including the independent packaged-client mTLS matrix.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,413 @@
...
+        Argument.AssertNotNullOrWhiteSpace(identityProviderId, nameof(identityProviderId));
+        Argument.AssertNotNullOrWhiteSpace(serviceAccountId, nameof(serviceAccountId));
+        Argument.AssertNotNull(options, nameof(options));
+        HttpMessageHandler handler = options.Handler
+            ?? throw new ArgumentNullException(nameof(options.Handler));
+
+        if (options.RefreshBuffer < TimeSpan.Zero)
+        {
+            throw new ArgumentOutOfRangeException(nameof(options), "The refresh buffer cannot be negative.");
+        }
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in d93bb5bb31eef1d2d0c44b66aecf4138bd224ea8 by changing the exception parameter to `nameof(options.RefreshBuffer)`. `ConstructorRejectsInvalidConfiguration` now asserts that `ArgumentOutOfRangeException.ParamName` is exactly `RefreshBuffer`; the regression passes across .NET 8, 9, and 10.

OpenAI/src/Utility/WorkloadIdentityAuthenticationPolicy.cs
```diff
@@ -0,0 +1,91 @@
...
+        if (!ShouldReplay(message))
+        {
+            return;
+        }
+
+        message.SetProperty(typeof(WorkloadIdentityAuthenticationPolicy), s_replayMarker);
+        message.ExtractResponse()?.Dispose();
+        token = await _credential.GetTokenAsync(message.CancellationToken).ConfigureAwait(false);
+        message.Request.Headers.Set("Authorization", "Bearer " + token);
+        await ProcessNextAsync(message, pipeline, currentIndex).ConfigureAwait(false);
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8e73d66f476b86c4908a66d9497a6323473bceb9: both synchronous and asynchronous authentication paths inspect the single permitted replay response and invalidate its replacement token if that replay also returns HTTP 401, without initiating a third attempt. `UnauthorizedReplayInvalidatesReplacementTokenForNextRequest` proves the original and replay each execute once, both rejected tokens are discarded, and the next independent non-replayable request obtains token 3 and succeeds without resending its body. All 44 focused tests pass on .NET 8/9/10, and the packaged public-client mTLS matrix passes independently.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -394,6 +419,25 @@ protected internal OpenAIClient(ClientPipeline pipeline, OpenAIClientOptions opt
...
+
+        if (options.Transport is not null)
+        {
+            throw new ArgumentException(
+                "X.509 workload identity must use the transport created from its credential's handler.",
+                nameof(options));
+        }
+
+        options.Transport = credential.Transport;
+        options.Endpoint ??= new Uri("https://mtls.api.openai.com/v1");
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8e73d66f476b86c4908a66d9497a6323473bceb9: `ConfigureWorkloadIdentityOptions` now recognizes the exact credential-created transport by reference and reuses already-frozen options without writing to them, while continuing to reject every independently supplied or different-credential transport. `ClientOptionsCanBeReusedWithTheSameCredential` constructs two clients from the identical frozen options and proves they share the credential's single token exchange. The clean NuGet-installed consumer also constructs and exercises two public clients from the same options on .NET 8/9/10.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,415 @@
...
+                    ? await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false)
+                    : _httpClient.Send(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+#else
+                response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
+#endif
+            }
+            catch (HttpRequestException) when (attempt + 1 < MaximumExchangeAttempts)
+            {
+                await DelayAsync(GetRetryDelay(attempt, retryAfter: null), async, cancellationToken).ConfigureAwait(false);
+                continue;
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8e73d66f476b86c4908a66d9497a6323473bceb9: the final `HttpRequestException` from token exchange is wrapped in a non-retryable `InvalidOperationException` after the third attempt, preventing `ClientRetryPolicy` from restarting the exchange loop. `NetworkExchangeRetriesAreNotMultipliedByPipelineRetries` injects an always-failing native `SocketsHttpHandler.ConnectCallback`, enables two outer pipeline retries, and asserts exactly three connection attempts. The separate packaged-client real-transport harness repeats the same bounded-network scenario on .NET 8/9/10.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -394,6 +419,30 @@ protected internal OpenAIClient(ClientPipeline pipeline, OpenAIClientOptions opt
...
+                return options;
+            }
+
+            throw new ArgumentException(
+                "X.509 workload identity must use the transport created from its credential's handler.",
+                nameof(options));
+        }
+
+        options.Transport = credential.Transport;
+        options.Endpoint ??= new Uri("https://mtls.api.openai.com/v1");
```
HAYDEN-OAI: **[P1] Reject non-HTTPS workload-identity API endpoints**

An explicit `options.Endpoint = new Uri("http://...")` survives this assignment, and `WorkloadIdentityAuthenticationPolicy` then unconditionally adds `Authorization: Bearer <federated token>` before the HTTP transport sends the request. The fixed HTTPS token exchange therefore succeeds and then transmits the newly minted token and request in plaintext; that API leg also cannot perform mutual TLS. The adjacent mTLS example already requires `Uri.UriSchemeHttps`, but this public constructor never applies that safeguard. Reject non-HTTPS explicit endpoints before attaching the workload-identity transport.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,422 @@
...
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            using HttpRequestMessage request = CreateExchangeRequest();
+            HttpResponseMessage response;
+
+            try
+            {
+#if NET8_0_OR_GREATER
+                response = async
+                    ? await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false)
```
HAYDEN-OAI, line 257: **[P1] Apply the configured network timeout to token exchange**

This direct token-exchange request uses an `HttpClient` whose timeout was explicitly set to infinite, but token acquisition runs in an authentication policy *before* `System.ClientModel`'s transport creates the cancellation token for `OpenAIClientOptions.NetworkTimeout` (100 seconds by default). Consequently neither the default nor a caller-configured network timeout applies here. With the usual `CancellationToken.None`, an identity endpoint that accepts a connection and stalls blocks forever while holding `_refreshLock`, so every concurrent and subsequent SDK request also hangs. Pass the message's configured network timeout into token acquisition and apply a bounded linked timeout covering the exchange and response-body reads.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,422 @@
...
+#if NET8_0_OR_GREATER
+        using Stream stream = async
+            ? await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false)
+            : response.Content.ReadAsStream(cancellationToken);
+#else
+        using Stream stream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
+#endif
+        using JsonDocument document = async
+            ? await JsonDocument.ParseAsync(stream, cancellationToken: cancellationToken).ConfigureAwait(false)
+            : JsonDocument.Parse(stream);
```
HAYDEN-OAI: **[P2] Preserve caller cancellation while synchronously reading token responses**

After `HttpClient.Send(..., ResponseHeadersRead, cancellationToken)` returns headers, the .NET HTTP implementation removes its request-level cancellation registration. This synchronous `JsonDocument.Parse(stream)` has no cancellation token, and the synchronous error-body `stream.Read(...)` below has the same problem. If the token endpoint sends headers and then stalls its body, canceling the caller's token cannot interrupt the synchronous SDK request, which continues holding the shared refresh lock. This remains broken even if a timeout is added only around `Send`. Register cancellation to dispose the response stream, as `System.ClientModel`'s transport does, or otherwise use cancellation-aware body reads for both success and error responses.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,422 @@
...
+            {
+#if NET8_0_OR_GREATER
+                response = async
+                    ? await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false)
+                    : _httpClient.Send(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+#else
+                response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
+#endif
+            }
+            catch (HttpRequestException exception)
```
HAYDEN-OAI: **[P2] Keep connection-timeout and response-body failures inside the exchange retry bound**

Catching only `HttpRequestException` does not cover real token-exchange transport failures: `SocketsHttpHandler.ConnectTimeout` produces an `OperationCanceledException` even when the caller was not canceled, and a truncated `200`/`503` response body throws `HttpIOException` from code outside this `try` block. `System.ClientModel` treats both non-user cancellations and `IOException` as retryable, so these failures escape to the outer pipeline and restart authentication four times by default, or arbitrarily more with caller-configured retries, violating `MaximumExchangeAttempts`. Handle non-user connection timeouts and response-body I/O failures within the same bounded exchange loop, preserve genuine caller cancellation, and wrap exhausted failures in a non-retryable exception.

HAYDEN-OAI (commented):
Reviewed the X.509 transport and authentication flow against the exact System.ClientModel dependency and .NET runtime implementations. Four actionable issues affect plaintext endpoint handling, token-exchange timeouts, synchronous cancellation, and bounded retries; concrete failure modes and fixes are inline.

OpenAI/src/Custom/OpenAIClient.cs
jbeckwith-oai: Fixed by rejecting non-absolute/non-HTTPS X.509 `OpenAIClientOptions.Endpoint` values before assigning the credential transport, so no token can be exchanged or attached for an insecure API endpoint. Added `WorkloadIdentityRejectsNonHttpsApiEndpoints` for HTTP/FTP plus `ApiKeyClientsContinueToAllowCustomHttpEndpoints` to prove existing API-key localhost/custom-endpoint behavior remains unchanged, and documented the X.509-only HTTPS requirement.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Fixed by carrying the configured `OpenAIClientOptions.NetworkTimeout` into `WorkloadIdentityAuthenticationPolicy`, honoring `PipelineMessage.NetworkTimeout` overrides, and applying a linked cancellation deadline across the entire shared token exchange and response-body processing; the SDK's default is 100 seconds. Timed-out exchanges become non-retryable `InvalidOperationException` failures, preventing outer pipeline retry multiplication. Added stalled 200/503 response-body coverage and a per-message-timeout override regression.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Fixed by registering cancellation to dispose the response stream during synchronous token JSON parsing and bounded synchronous error-body draining, then normalizing cancellation-triggered stream/SSL failures back into caller cancellation. Added real HTTPS regressions for stalled 200 and 503 bodies that cancel a synchronous request, verify prompt completion, and prove the refresh lock is released by successfully issuing the next request.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Fixed by moving response-body reads/drains inside the credential's bounded exchange retry loop and treating `HttpRequestException`, response-body `IOException`/`HttpIOException`, and non-caller-canceled `OperationCanceledException` connection timeouts as transient exchange failures. Exhaustion is converted into non-retryable `InvalidOperationException` so the outer SDK retry policy cannot multiply attempts. Added regressions for connection timeouts plus truncated 200 and 503 response bodies with an intentionally aggressive outer retry policy; every case proves exactly three exchange attempts.

HAYDEN-OAI (approved):
Re-reviewed the exact updated head and verified the previously reported HTTPS, exchange-timeout, cancellation, and bounded-retry fixes. Authentication, transport ownership, refresh/replay behavior, and existing client compatibility look good.

jsquire (changes requested):
Blocking to ensure that we don't have a accidental merge before .NET core team review.   As this proposes both public API changes, it'll be subject to additional review and discussion.

jbeckwith-oai:
Addressed all three P1 findings from https://github.com/openai/openai-dotnet/pull/1292#issuecomment-5318243813 in commit 1bd29f92ecd2d95a06dd28ef8a957ead232956c3:

1. **Explicit Host / HTTP/2 :authority.** The workload-identity authentication policy now validates every explicit `Host` value with `TryGetValues` *before acquiring a token*, and the transport validates the actual final `HttpRequestMessage` again after caller policies run. Both checkpoints reject duplicate/comma-joined values, malformed authorities, userinfo/path/query/fragment, and mismatched HTTPS host/port. Coverage: `PublicPipelineRejectsSpoofedHostAuthorityBeforeTokenAcquisition`, `BeforeTransportPoliciesCannotSpoofTheOutgoingHostAuthority`, and `AnExplicitMatchingHostAuthorityRemainsSupported`; the freshly packaged real-mTLS consumer separately verifies both early and final-sink rejection.

2. **401 replay and non-seekable content.** The specific reported reproducer cannot be constructed: `System.ClientModel.BinaryContent.Create(nonSeekableStream)` throws `ArgumentException` because the framework requires a seekable stream. The new regression `FrameworkRejectsNonSeekableStreamsBeforeWorkloadIdentityCanSendThem` proves this before any network request. Replay remains restricted to known framework-owned repeatable content; existing regressions `UnauthorizedNonReplayableRequestIsNeverRetried`, `UnauthorizedNonReplayableRequestInvalidatesTokenForNextRequest`, and `UnauthorizedSeekableStreamIsSafelyReplayed` verify arbitrary non-replayable content is never resent while genuinely seekable content remains supported.

3. **Automatic cookies.** Both supported native handlers now fail closed unless `UseCookies = false`; validation is repeated at the token-exchange/API transport boundaries, eliminating the mutable-cookie-container race. Explicit request-scoped API `Cookie` headers remain supported and cannot leak into the separate exchange request. Coverage: `ConstructorRejectsAutomaticCookieHandling` for both `HttpClientHandler` and `SocketsHttpHandler`, `ApiCookiesCannotCrossIntoASubsequentTokenExchange`, and `ExplicitCookiesScopedOnlyToTheApiRequestRemainSupported`, plus the packaged CA-signed mTLS cookie-isolation scenario.

Validation: 114 focused X.509 security/reliability tests; full playback suites (.NET 8: 1,678 passed, .NET 9: 1,690 passed, .NET 10: 1,692 passed; 87 existing skips each); 19/19 freshly packaged real-mTLS scenarios on each runtime; clean code generation, generator tests, API export, documentation snippets, analyzer-enabled Release build, zero-vulnerability .NET dependency audit, and APICompat with no breaking changes.

tests/MutualTls/X509WorkloadIdentitySecurityTests.cs
```diff
@@ -0,0 +1,642 @@
...
+        Assert.That(server.ApiCount, Is.EqualTo(1));
+    }
+
+    [Test]
+    public void FrameworkRejectsNonSeekableStreamsBeforeWorkloadIdentityCanSendThem()
+    {
+        using NonSeekableMemoryStream stream = new(Encoding.UTF8.GetBytes("single-use payload"));
+
+        ArgumentException exception = Assert.Throws<ArgumentException>(
+            () => System.ClientModel.BinaryContent.Create(stream));
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8d2b873460ff5cfbe39ba7639f886fb45d713c3a. The Assert.Throws lambda now declares `using System.ClientModel.BinaryContent content = System.ClientModel.BinaryContent.Create(stream);`, so an unexpectedly constructed BinaryContent is deterministically disposed while the expected non-seekable ArgumentException behavior remains unchanged. Verified with the full 114-test X.509 suite, analyzer-enabled Release build, changed-file formatting, and a one-file thermo review; no suppression or production/API change was added.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -394,6 +424,46 @@ protected internal OpenAIClient(ClientPipeline pipeline, OpenAIClientOptions opt
...
+                return options;
+            }
+
+            throw new ArgumentException(
+                "X.509 workload identity must use the transport created from its credential's handler.",
+                nameof(options));
+        }
+
+        options.Transport = credential.Transport;
+        options.Endpoint ??= X509WorkloadIdentityCredential.DefaultApiEndpoint;
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 931f7956c9ba72ec84492fdf85bf8191453c96e8. X.509 construction now uses the framework-provided `ClientPipelineOptions.Clone()` to create an independent mutable pipeline configuration, installs the credential-owned transport only on that private clone, and creates a separate typed endpoint/options snapshot for child clients. The caller's original `OpenAIClientOptions` is never mutated or frozen by the X.509 client, custom caller policies and timeout settings survive cloning, and the original options remain usable by API-key clients whether the API-key client is constructed before or after workload identity. `SharedOptionsKeepApiKeyAndWorkloadIdentityAuthenticationIndependent(false/true)` reproduces both orderings against real mTLS, and `ClientOptionsCanBeReusedWithTheSameCredential` now additionally asserts the original endpoint/transport remain unset. Verified on .NET 8/9/10, including the clean packaged real-mTLS consumer and full playback suites.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -132,6 +133,27 @@ public OpenAIClient(ApiKeyCredential credential, OpenAIClientOptions options) :
...
+    /// <param name="workloadIdentityCredential">The X.509 workload identity credential.</param>
+    public OpenAIClient(X509WorkloadIdentityCredential workloadIdentityCredential)
+        : this(workloadIdentityCredential, new OpenAIClientOptions())
+    {
+    }
+
+    /// <summary>Initializes an HTTP client authenticated with X.509 workload identity.</summary>
+    /// <param name="workloadIdentityCredential">The X.509 workload identity credential.</param>
+    /// <param name="options">The client options, which must not specify a separate transport.</param>
+    public OpenAIClient(X509WorkloadIdentityCredential workloadIdentityCredential, OpenAIClientOptions options)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 931f7956c9ba72ec84492fdf85bf8191453c96e8. Both public `OpenAIClient(X509WorkloadIdentityCredential)` and `OpenAIClient(X509WorkloadIdentityCredential, OpenAIClientOptions)` constructors now explicitly carry `[Experimental("OPENAI001")]`, matching the credential/options types and the neighboring experimental authentication-policy constructors. `WorkloadIdentityConstructorsAreExplicitlyExperimental(false/true)` verifies both overloads by reflection; the net8.0 and net10.0 exported API manifests were regenerated, and the repository's `Test-ExperimentalAttributes.ps1`, complete code generation, API export, analyzer-enabled build, and ApiCompat checks all pass. The netstandard2.0 export intentionally omits experimental attributes consistently with all existing netstandard APIs.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,587 @@
...
+
+        TimeSpan lifetime = TimeSpan.FromSeconds(lifetimeSeconds);
+        if (lifetime <= TimeSpan.Zero)
+        {
+            throw new InvalidOperationException("X.509 workload identity token exchange returned an invalid token lifetime.");
+        }
+
+        TimeSpan maximumBuffer = TimeSpan.FromTicks(lifetime.Ticks / 2);
+        TimeSpan effectiveBuffer = _refreshBuffer < maximumBuffer ? _refreshBuffer : maximumBuffer;
+        return new CachedToken(accessToken, GetTimestamp(), lifetime - effectiveBuffer);
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 209bc0adf17bd1eed83023991992e3faee916460. `ExchangeTokenAsync` now captures the existing monotonic `GetTimestamp()` immediately when `ResponseHeadersRead` completes, carries that same issuance origin through `ReadTokenAsync` into `CachedToken`, and uses the existing centralized `IsFresh` check before publishing any newly exchanged token into the cache or returning its bearer. This gives first use, cached reuse, and the half-TTL-clamped refresh buffer one coherent monotonic lifetime; response-body time can no longer extend `expires_in`, an already-expired response never reaches the API or poisons the cache, and the next legitimate exchange recovers normally.

Fail-first regressions: `DelayedExchangeBodiesCannotCacheOrSendExpiredTokens(false/true)` flushes response headers, delays a 100 ms token body by 250 ms, verifies both synchronous and asynchronous clients reject the redacted expired token before any API request, and verifies clean subsequent reacquisition. `ExchangeBodyDelayConsumesTheSameMonotonicRefreshWindowAsCachedReuse` proves body delay is counted against the configured refresh boundary while immediate fresh-token reuse remains valid.

Validation: full .NET 8/9/10 suites passed 1,685 / 1,697 / 1,699 tests; all 21 clean packaged CA-signed real-mTLS attack scenarios passed on each runtime, including sync/async delayed-body expiry; analyzer-enabled multi-target build, NuGet ApiCompat/package checks, formatting, generator tests, experimental API validation, zero-vulnerability NuGet audit, and a sealed immutable Codex Security diff scan (`eab579df-f584-40b1-b94a-e0ae43bf861d`) all passed.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,595 @@
...
+        IWebProxy? proxy = configuredProxy ?? HttpClient.DefaultProxy;
+#else
+        IWebProxy? proxy = configuredProxy
+            ?? typeof(HttpClient).GetProperty("DefaultProxy")?.GetValue(null) as IWebProxy
+            ?? WebRequest.DefaultWebProxy;
+#endif
+        if (proxy is not null
+            && !proxy.IsBypassed(destination)
+            && proxy.GetProxy(destination) is Uri proxyEndpoint
+            && !string.Equals(proxyEndpoint.Scheme, Uri.UriSchemeHttp, StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in c9b512a0 by installing a validating IWebProxy directly on the same caller-owned HttpClientHandler/SocketsHttpHandler before first use. Its GetProxy validates the exact URI returned to the native handler, so stateful bypass/HTTP→HTTPS changes cannot create an outer TLS proxy connection; post-construction proxy replacement is rejected. Explicit HTTP CONNECT proxy credentials and implicit DefaultProxyCredentials remain supported, including subsequent credential updates, without transferring handler/certificate/key ownership. Added fail-first regressions for both handler types, actual handler-time proxy changes, proxy replacement, and scoped/default proxy credentials. Independently packaged real CA-signed mTLS consumers exercised a hostile HTTPS proxy requesting client certificates on .NET 8/9/10 and verified zero disclosed certificates while ordinary HTTP CONNECT continued to work.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,595 @@
...
+                    TimeSpan delay = GetRetryDelay(attempt, response.Headers.RetryAfter);
+                    await DrainErrorBodyAsync(response, async, cancellationToken).ConfigureAwait(false);
+                    await DelayAsync(delay, async, cancellationToken).ConfigureAwait(false);
+                    continue;
+                }
+
+                if (!response.IsSuccessStatusCode)
+                {
+                    await DrainErrorBodyAsync(response, async, cancellationToken).ConfigureAwait(false);
+                    throw new InvalidOperationException(
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in c9b512a0 by moving unsuccessful-response handling into one status-owned failure path: bounded error-body drainage remains best-effort for a received non-transient HTTP status, and truncated/reset IOException/HttpRequestException cannot replace or retry the authoritative 400/403. Genuine cancellation remains cancellation, while malformed/truncated successful or transient 200/503 responses retain the existing exact three-attempt exchange retry bound. Added fail-first real-loopback regressions for both 400 and 403 on both synchronous and asynchronous public pipelines, with an aggressive outer retry policy; each proves one exchange, the original status in the redacted exception, and zero API requests. Packaged CA-signed real-mTLS consumers independently verified both failures on .NET 8/9/10.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,689 @@
...
+
+    private sealed class ValidatingProxy(IWebProxy proxy, ICredentials? defaultCredentials) : IWebProxy
+    {
+        public ICredentials? Credentials
+        {
+            get => defaultCredentials ?? proxy.Credentials;
+            set
+            {
+                defaultCredentials = null;
+                proxy.Credentials = value;
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 91b93a3f by separating proxy routing from credential ownership inside the existing validating IWebProxy. When the original native handler has Proxy == null, the wrapper snapshots the handler/default proxy credentials and owns all subsequent credential updates locally; it never writes to the shared HttpClient.DefaultProxy. When the caller explicitly supplied a proxy, that same caller-owned proxy remains the live credential owner and setter updates still propagate normally. Added fail-first regressions for both HttpClientHandler and SocketsHttpHandler proving rotated credentials leave the shared default unchanged and cannot be observed by an unrelated handler, plus explicit-proxy mutation controls. Independently built NuGet consumers on .NET 8/9/10 exercised a real 407-challenging HTTP CONNECT proxy with CA-signed mutual TLS: workload requests sent only the rotated handler-local credentials, and a separate ordinary HttpClient sent only the original shared proxy credentials. All full runtime suites, API compatibility/package validation, and a fresh sealed zero-finding security scan passed.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -394,6 +423,57 @@ protected internal OpenAIClient(ClientPipeline pipeline, OpenAIClientOptions opt
...
         => OpenAIClientUtilities.CreateApiKeyAuthenticationPolicy(credential);
 
+    private static ClientPipeline CreateWorkloadIdentityPipeline(
+        X509WorkloadIdentityCredential credential,
+        OpenAIClientOptions options)
+    {
+        Argument.AssertNotNull(credential, nameof(credential));
+        options ??= new OpenAIClientOptions();
+
+        Uri endpoint = options.Endpoint ?? X509WorkloadIdentityCredential.DefaultApiEndpoint;
```
HAYDEN-OAI, line 460: **[P1] Reject provider-owned endpoints before attaching X.509 credentials**

`new OpenAIClient(credential, new OpenAIClientOptions { Endpoint = new Uri("https://tenant.openai.azure.com/openai/v1") })` passes the HTTPS/userinfo checks and establishes that Azure origin as trusted. The first request then exchanges the workload certificate at `https://mtls.auth.openai.com/oauth/token`, attaches the resulting transferable OpenAI service-account bearer to the Azure request, and can also present the same workload certificate to that server. Reusing one credential across the normal OpenAI origin and this endpoint even reuses the same cached bearer. The adjacent mTLS example already explicitly allowlists the global/EU OpenAI authorities, but this public constructor does not apply an equivalent provider boundary. Reject Azure/other provider-owned origins before token acquisition, or require explicit trusted-origin validation for any intentionally supported custom gateway.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,704 @@
...
+                native.UseCookies,
+                native.UseProxy,
+                native.Proxy);
+            return;
+        }
+
+#if NET8_0_OR_GREATER
+        if (handler.GetType() == typeof(SocketsHttpHandler))
+        {
+            SocketsHttpHandler native = (SocketsHttpHandler)handler;
```
HAYDEN-OAI, line 652: **[P1] Bind token caching to the effective client-certificate identity**

This native-handler branch accepts `SslOptions.LocalCertificateSelectionCallback` without a pinned `ClientCertificateContext`, yet .NET invokes that callback separately with the TLS target host for each connection. A supported handler configured as `(_, host, _, _, _) => host == "mtls.auth.openai.com" ? certificateA : certificateB` therefore exchanges under certificate A and then presents certificate B to `mtls.api.openai.com` while sending A's cached service-account bearer. This occurs with the two ordinary trusted OpenAI hosts, requires no redirect or custom endpoint, and violates the promised same-certificate identity on both TLS legs; rotating/context-sensitive selectors can likewise retain a bearer from a previous identity. Pin or verify one effective certificate identity across exchange, API transport, and token cache, or reject dynamic selectors when that binding cannot be enforced.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,704 @@
...
+            IWebProxy? configuredProxy = socketsHandler.Proxy;
+            IWebProxy proxy = GetEffectiveProxy(configuredProxy)
+                ?? throw new ArgumentException("The workload identity HTTP handler proxy is unavailable.", nameof(handler));
+            ICredentials? defaultCredentials = configuredProxy is null
+                ? socketsHandler.DefaultProxyCredentials
+                : null;
+            socketsHandler.Proxy = new ValidatingProxy(
+                proxy,
+                configuredProxy,
+                defaultCredentials ?? proxy.Credentials);
```
HAYDEN-OAI: **[P2] Preserve native default-proxy credential precedence**

For an implicit proxy (`handler.Proxy == null`), .NET normally resolves proxy authentication as `HttpClient.DefaultProxy.Credentials ?? handler.DefaultProxyCredentials`. This wrapper reverses that precedence to `defaultCredentials ?? proxy.Credentials` in both supported-handler branches. Configure the shared default proxy with corporate credentials and `handler.DefaultProxyCredentials` with separate tenant-private credentials: an ordinary native handler authenticates to the shared proxy with the corporate account, but this credential instead sends the tenant-private username/password as `Proxy-Authorization` after a 407 challenge. Preserve native precedence (`proxy.Credentials ?? defaultCredentials`) while retaining wrapper-local credential ownership.

HAYDEN-OAI (commented):
Independent security review of exact head `91b93a3f1e5b7302aea72133cbbd2d0e08ac35a1` identified three concrete credential-isolation defects: provider-owned API endpoints can receive OpenAI workload credentials; host-dependent native certificate selection can mix mTLS identities with cached bearer tokens; and implicit-proxy wrapping reverses .NET proxy-credential precedence. The inline comments include the triggering configurations and fix directions.

OpenAI/src/Custom/OpenAIClient.cs
jbeckwith-oai: Fixed in d57eef833c1642800dcb920445dea15cee0f7a61. X.509 client construction now performs one centralized, canonical-IDN, dot-boundary provider-origin trust decision before token acquisition or pipeline mutation. Azure (including sovereign domains), AWS, and Google-owned origins fail closed; intentionally trusted customer OpenAI gateways and existing Azure API-key clients remain supported. Added public-client fail-first regressions for mixed-case/trailing-dot and provider-owned endpoints plus positive gateway/API-key controls, and verified provider rejection through the real packaged CA-signed mTLS consumer on .NET 8, 9, and 10.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Fixed in d57eef833c1642800dcb920445dea15cee0f7a61. Both supported native handlers now normalize certificate-selection safety through their existing single validation boundary: reject host-dependent LocalCertificateSelectionCallback (including when ClientCertificateContext is also configured), HttpClientHandler automatic selection, and multiple candidate client certificates. Validation is repeated before exchange and final dispatch, so a selector installed after construction also fails before any token request; one fixed caller-owned certificate/context remains supported. Added regressions with two distinct real generated certificates, both handler types, pinned-context variants, and late callback mutation; all three packaged real-mTLS runtime consumers passed.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Fixed in d57eef833c1642800dcb920445dea15cee0f7a61. The existing handler-time ValidatingProxy now owns one canonical initial-credential precedence rule, `proxy.Credentials ?? defaultCredentials`, for both HttpClientHandler and SocketsHttpHandler; implicit proxy credential rotation remains wrapper-local and explicit caller-owned proxies retain their live credential semantics. Added both-handler regressions for shared-corporate-before-handler precedence, handler fallback when the shared proxy has none, and non-mutation of HttpClient.DefaultProxy. A real hermetic 407 HTTP CONNECT proxy additionally verified that only corporate credentials are sent when both are configured, while rotated credentials stay isolated from unrelated HttpClient instances, across .NET 8/9/10 packaged consumers.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -394,6 +441,79 @@ protected internal OpenAIClient(ClientPipeline pipeline, OpenAIClientOptions opt
...
+    {
+        string host = endpoint.IdnHost.TrimEnd('.');
+        foreach (string domain in s_externalProviderDomains)
+        {
+            if (string.Equals(host, domain, StringComparison.OrdinalIgnoreCase)
+                || host.EndsWith("." + domain, StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+        }
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 706740486ae7c84719706d1b38207b687942e0de. Replaced the manual foreach/conditional return in IsExternalProviderEndpoint with the requested single `s_externalProviderDomains.Any(...)` predicate and added the required `System.Linq` import. The existing `string.Equals(..., StringComparison.OrdinalIgnoreCase)` and `EndsWith(..., StringComparison.OrdinalIgnoreCase)` checks are unchanged, preserving canonical host normalization, provider suffix boundaries, trusted custom gateways, and existing Azure API-key behavior. Verified 148/148 X.509/security tests on each of .NET 8/9/10; full runtime suites passed 1,712/1,724/1,726 tests, all-TFM package ApiCompat passed, and the freshly packaged real-mTLS consumer passed all 26 adversarial scenarios.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,712 @@
...
+
+        TimeSpan lifetime = TimeSpan.FromSeconds(lifetimeSeconds);
+        if (lifetime <= TimeSpan.Zero)
+        {
+            throw new InvalidOperationException("X.509 workload identity token exchange returned an invalid token lifetime.");
+        }
+
+        TimeSpan maximumBuffer = TimeSpan.FromTicks(lifetime.Ticks / 2);
+        TimeSpan effectiveBuffer = _refreshBuffer < maximumBuffer ? _refreshBuffer : maximumBuffer;
+        return new CachedToken(accessToken, issuedAt, lifetime - effectiveBuffer);
```
g2vinay: This treats the returned token as an ordinary bearer token. Using the same certificate for both TLS connections does not guarantee that the token is bound to it. Please document and test the server-side certificate binding, and carry the binding certificate with the cached token as Azure.Core does.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,712 @@
...
+
+    internal static Uri DefaultApiEndpoint { get; } = new("https://mtls.api.openai.com/v1");
+
+    private readonly string _identityProviderId;
+    private readonly string _serviceAccountId;
+    private readonly HttpMessageHandler _handler;
+    private readonly HttpClient _httpClient;
+    private readonly SemaphoreSlim _refreshLock = new(1, 1);
+    private readonly TimeSpan _refreshBuffer;
+    private CachedToken? _cachedToken;
```
g2vinay, line 42: A credential has one token cache but can be reused with different API endpoints. This could send the same token to multiple origins. Please bind each credential to one endpoint and audience, or include those values in the cache key and token request.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,712 @@
...
+    private static void ValidateHandler(HttpMessageHandler handler, Uri destination)
+    {
+        if (handler.GetType() == typeof(HttpClientHandler))
+        {
+            HttpClientHandler native = (HttpClientHandler)handler;
+            ValidateHandlerConfiguration(
+                handler,
+                destination,
+                native.ClientCertificateOptions != ClientCertificateOption.Manual
+                    || native.ClientCertificates.Count > 1,
```
g2vinay, line 640: The validation rejects multiple certificates but allows no certificate at all. It also does not prevent the configured certificate from changing while an old token remains cached. Please require and capture exactly one certificate identity when constructing the credential.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,712 @@
...
+        throw new InvalidOperationException("X.509 workload identity token exchange exhausted its retry attempts.");
+    }
+
+    private HttpRequestMessage CreateExchangeRequest()
+    {
+        using MemoryStream buffer = new();
+        using (Utf8JsonWriter writer = new(buffer))
+        {
+            writer.WriteStartObject();
+            writer.WriteString("grant_type", "urn:ietf:params:oauth:grant-type:token-exchange");
```
g2vinay, line 311: This request uses the RFC 8693 token-exchange grant, but it omits the required `subject_token` and sends JSON instead of form data. Please either use a standard RFC 8705 client-credentials flow or define a clearly named OpenAI-specific grant.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -394,6 +442,72 @@ protected internal OpenAIClient(ClientPipeline pipeline, OpenAIClientOptions opt
...
+            pipelineOptions,
+            options.UserAgentApplicationId,
+            options.OrganizationId,
+            options.ProjectId);
+    }
+
+    private static bool IsExternalProviderEndpoint(Uri endpoint)
+    {
+        string host = endpoint.IdnHost.TrimEnd('.');
+        return s_externalProviderDomains.Any(domain =>
```
g2vinay, line 505: Blocking known Azure, AWS, and Google domains does not make every other HTTPS endpoint trustworthy. Please allow known OpenAI mTLS endpoints by default and require explicit opt-in for custom gateways.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,712 @@
...
+
+            acquired = true;
+            ValidateEndpoint(s_exchangeEndpoint);
+            token = Volatile.Read(ref _cachedToken);
+            if (token is not null && IsFresh(token))
+            {
+                return token.Value;
+            }
+
+            token = await ExchangeTokenAsync(async, deadline.Token).ConfigureAwait(false);
```
g2vinay: The semaphore combines successful refreshes, but each waiter can retry independently after a failure. During an outage, N callers could produce up to 3N token requests. Please share the in-flight failure and add a short credential-level backoff.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,712 @@
...
+            IWebProxy proxy = GetEffectiveProxy(configuredProxy)
+                ?? throw new ArgumentException("The workload identity HTTP handler proxy is unavailable.", nameof(handler));
+#if NET8_0_OR_GREATER
+            ICredentials? defaultCredentials = configuredProxy is null
+                ? clientHandler.DefaultProxyCredentials
+                : null;
+#else
+            ICredentials? defaultCredentials = null;
+#endif
+            clientHandler.Proxy = new ValidatingProxy(proxy, configuredProxy, defaultCredentials);
```
g2vinay, line 739: Credential construction permanently replaces the proxy on the caller-owned handler. That is surprising if the handler is shared elsewhere. Please require a dedicated handler and document that it is mutated, or accept a factory for transport instances owned by the credential.

g2vinay (commented):
Leaving my comments in-line

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Addressed in 25c2f52e2087b1c45877652bc511060453c4096d with an authoritative contract clarification, not certificate retention. The X.509 WIF service intentionally returns an ordinary bearer token with no DPoP/cnf certificate binding; bearer authorization and mTLS admission are independent API checks. The SDK therefore must not carry or retain a binding certificate/fingerprint. README and the mTLS example now state that contract explicitly, while the existing transport checks still require one stable native TLS selection path and reject host-dependent/multiple selectors.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Fixed in 25c2f52e2087b1c45877652bc511060453c4096d. A credential now atomically binds to the first configured HTTPS API origin and rejects a second client with a different scheme/IDN host/port before token acquisition or dispatch, so one cached bearer cannot be reused across origins. Same-origin child clients still share the pipeline/cache. Regression: SharedCredentialRejectsDifferentConfiguredHttpsOrigins; the security pass also found and fixed construction ordering so a rejected independent Transport cannot poison the bound origin.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Addressed in 25c2f52e2087b1c45877652bc511060453c4096d by clarifying the native-transport contract rather than capturing certificate identity. The SDK must not retain certificate bytes, fingerprints, or a binding certificate, and requiring a visible collection entry would reject native OS/HSM-backed transport configurations. Existing validation rejects automatic, host-dependent, and multiple selectable certificates; missing/invalid client presentation remains an authoritative TLS/token-endpoint admission failure. Docs now state that callers configure one fixed native certificate and rotate by rebuilding the dedicated handler/credential/client group.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Addressed in 25c2f52e2087b1c45877652bc511060453c4096d with explicit documentation of the approved OpenAI-specific wire contract. The service requires JSON with `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`, the X.509 subject-token-type URN, provider/service-account IDs, and structural omission of `subject_token` because the authenticated TLS connection is the subject credential. Changing to form encoding, adding `subject_token`, or switching to RFC 8705 would break the service contract; tests continue to assert the exact host/path/body and absence of `subject_token`.

OpenAI/src/Custom/OpenAIClient.cs
jbeckwith-oai: Addressed in 25c2f52e2087b1c45877652bc511060453c4096d. The default remains the known global mTLS API origin; explicitly setting `OpenAIClientOptions.Endpoint` is the existing opt-in for one trusted custom OpenAI gateway. X.509 endpoints still require HTTPS/no userinfo and reject provider-owned Azure/AWS/Google origins before acquisition. The credential is now additionally bound to that one origin, preventing cached bearer reuse with a second endpoint. README/example text now makes the explicit opt-in and one-origin rule unambiguous.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Fixed in 25c2f52e2087b1c45877652bc511060453c4096d. The credential now stores a recent non-cancellation refresh failure for a one-second monotonic backoff window; queued waiters rethrow that same sanitized failure instead of each consuming another three-attempt exchange budget. Successful refresh clears the failure, and caller cancellation/network deadlines are not shared or allowed to poison other waiters. Regression ConcurrentWaitersShareRecentFailedRefresh starts eight concurrent callers against a delayed 503 endpoint and proves exactly three exchange requests, not 24.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
jbeckwith-oai: Addressed in 25c2f52e2087b1c45877652bc511060453c4096d by making the existing ownership contract explicit. The caller still owns and disposes the native handler/certificates/pool; the credential owns only its HttpClient wrapper, but the supplied handler must be dedicated to this credential because the SDK installs the authoritative proxy-validation wrapper needed to prevent HTTPS-proxy certificate disclosure and mutable-proxy TOCTOU bypasses. XML docs, README, and the mTLS example now state that mutation and lifetime requirement; no handler factory or second transport was introduced, preserving the approved native mTLS/HSM ownership model.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,777 @@
...
+                && !cancellationToken.IsCancellationRequested)
+            {
+                Volatile.Write(ref _refreshFailure, new RefreshFailure(exception, GetTimestamp()));
+                throw;
+            }
+
+            if (!IsFresh(token))
+            {
+                throw new InvalidOperationException(
+                    "X.509 workload identity token exchange returned a token that expired before it could be used.");
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in b8db599d3d15da76dda5464001e896d86b55f696. The post-read freshness rejection now runs inside the credential's single refresh-failure boundary, so a successful exchange whose token is already unusable records the same monotonic _refreshFailure as other non-cancellation refresh failures. Queued callers now observe that shared cooldown/failure instead of serially consuming more exchange attempts.

Regression: ConcurrentWaitersShareUnusableSuccessfulRefresh delays the body of a 0.1-second token, starts eight callers, and proves one exchange, zero API calls, and no token leakage in the error. The adjacent delayed-body recovery test now also proves recovery occurs after the bounded cooldown. Focused mTLS suites passed 157/169/169 on .NET 8/9/10; full suites passed 1715/1727/1729; package, API export, snippets, ApiCompat, zero-warning solution build, security review, and thermo review passed.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,777 @@
...
+            try
+            {
+                token = await ExchangeTokenAsync(async, deadline.Token).ConfigureAwait(false);
+            }
+            catch (Exception exception) when (
+                exception is not OperationCanceledException
+                && !cancellationToken.IsCancellationRequested)
+            {
+                Volatile.Write(ref _refreshFailure, new RefreshFailure(exception, GetTimestamp()));
+                throw;
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in b8db599d3d15da76dda5464001e896d86b55f696. CachedToken now keeps monotonic absolute ExpiresAfter separately from buffered RefreshAfter. If proactive refresh fails, the credential returns only the exact still-current cached token while it remains before absolute expiry; a concurrent 401 invalidation cannot revive it, and after absolute expiry the shared failure is rethrown without extending token life.

Regression: FailedProactiveRefreshUsesCachedTokenUntilAbsoluteExpiry primes a two-second bearer, moves it into the refresh window, starts eight callers through a 503 refresh outage, and proves they share the three-attempt failure while all use the still-valid bearer; after absolute expiry it proves no API call and no extra exchange occur during the cooldown. Focused mTLS suites passed 157/169/169 on .NET 8/9/10; full suites passed 1715/1727/1729; package, API export, snippets, ApiCompat, zero-warning solution build, security review, and thermo review passed.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,819 @@
...
+                        "X.509 workload identity token exchange returned a token that expired before it could be used.");
+                }
+
+                Volatile.Write(ref _cachedToken, refreshedToken);
+                Volatile.Write(ref _refreshFailure, null);
+                return refreshedToken.Value;
+            }
+            catch (Exception exception) when (
+                exception is not OperationCanceledException
+                && !cancellationToken.IsCancellationRequested)
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 1a9babffdeba11a4b888266f40d54cecbd5b414d. The credential refresh owner now distinguishes caller cancellation from its linked internal deadline: only deadline-originated OperationCanceledException with an uncanceled caller token is normalized into the existing network-timeout InvalidOperationException, recorded in the credential-level refresh-failure cooldown, and allowed to use the exact still-current cached bearer before absolute expiry. Caller-requested cancellation still propagates unchanged.

Regression: ProactiveRefreshDeadlineUsesCachedTokenAndSharedFailure runs both sync and async public pipeline paths, primes a two-second token, stalls proactive refresh past a 200 ms internal deadline, proves fallback succeeds, then proves the next request shares the failure without another stalled exchange. Focused mTLS passed 159/171/171 on .NET 8/9/10; full suites passed 1717/1729/1731; package, ApiCompat, zero-warning solution build, formatting, security review, and thermo review passed.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -101,6 +120,7 @@ internal TopLevelResponsesClient(ClientPipeline pipeline, ResponsesClientOptions
 
     private readonly OpenAIClientOptions _options;
     private readonly ApiKeyCredential _keyCredential;
+    private readonly X509WorkloadIdentityCredential _workloadIdentityCredential;
```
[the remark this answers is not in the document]
jbeckwith-oai: Investigated on exact head 1a9babffdeba11a4b888266f40d54cecbd5b414d; no code change is needed. X509WorkloadIdentityCredential is a non-partial sealed type with exactly one constructor, at OpenAI/src/Custom/X509WorkloadIdentityCredential.cs:50, and that constructor assigns every readonly credential field before returning. AuthenticationPolicy and pipeline constructors are constructors of different types, not additional constructors for this credential.

Compiler evidence: the analyzer-enabled Release solution build compiled OpenAI for netstandard2.0, net8.0, and net10.0 plus tests/examples with 0 warnings and 0 errors; focused mTLS passed 159/171/171, full suites passed 1717/1729/1731, package creation succeeded for all shipped TFMs, and ApiCompat found no breaks. CS0171 is therefore not reachable on this type.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -394,6 +442,74 @@ protected internal OpenAIClient(ClientPipeline pipeline, OpenAIClientOptions opt
...
+        credential.ValidateEndpoint(endpoint);
+
+        if (options.Transport is not null)
+        {
+            throw new ArgumentException(
+                "X.509 workload identity must use the transport created from its credential's handler.",
+                nameof(options));
+        }
+
+        credential.BindApiEndpoint(endpoint);
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in f8534cf3d5c9f223b90f145fb16d7e9d7b8bcbac. Pipeline creation now completes before the credential commits its API origin, so a telemetry/UserAgentApplicationId construction failure leaves the credential unbound. The existing Interlocked.CompareExchange binding remains the single commit point, so concurrent different-origin constructors still permit exactly one successful origin.

Regressions: FailedPipelineCreationDoesNotBindCredentialOrigin proves an overlong UserAgentApplicationId can fail for one endpoint and a corrected client can then bind/use a different endpoint; ConcurrentClientConstructionBindsOnlyOneCredentialOrigin races two origins and proves exactly one construction succeeds. The focused 17-test security slice passed on .NET 8/9/10; MutualTls passed 170/182/182; full suites passed 1728/1740/1742.

OpenAI/src/Custom/X509WorkloadIdentityCredential.cs
```diff
@@ -0,0 +1,837 @@
...
+    private static void ValidateHandler(HttpMessageHandler handler, Uri destination)
+    {
+        if (handler.GetType() == typeof(HttpClientHandler))
+        {
+            HttpClientHandler native = (HttpClientHandler)handler;
+            ValidateHandlerConfiguration(
+                handler,
+                destination,
+                native.ClientCertificateOptions != ClientCertificateOption.Manual
+                    || native.ClientCertificates.Count > 1,
```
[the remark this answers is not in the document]
jbeckwith-oai: Investigated on f8534cf3d5c9f223b90f145fb16d7e9d7b8bcbac; no code change is appropriate. The approved Phase-1 boundary deliberately leaves certificates, private keys, HSM/native key handles, and TLS admission with the caller-owned native handler and the fixed mTLS token endpoint. A visible collection count is not a reliable proof of a usable identity: one entry can still have an unusable key/EKU/chain, while native TLS configuration can use a certificate context or non-exportable key material without the SDK retaining identity material. The existing SDK checks the enforceable local invariant—one stable selection path by rejecting automatic, host-dependent, and multiple selectable certificates—and the token endpoint authoritatively rejects missing/invalid presentation before any bearer exists.

Requiring a visible collection entry would add a false local ownership check and reject valid native/HSM-backed configurations while still not proving handshake success. Existing regressions cover automatic/host-dependent/multiple selectors, and the real mTLS tests prove both HTTPS legs present the same caller-owned certificate when configured. The focused 17-test security slice, MutualTls 170/182/182, full 1728/1740/1742, package, ApiCompat, codegen, API export, docs snippets, and zero-warning build all passed.

OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -86,6 +87,24 @@ namespace OpenAI;
...
+        "azure.cn",
+        "azure.us",
+        "azure-api.net",
+        "azurewebsites.net",
+        "chinacloudapi.cn",
+        "usgovcloudapi.net",
+        "microsoft.com",
+        "microsoftonline.com",
+        "amazonaws.com",
+        "amazonaws.com.cn",
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in f8534cf3d5c9f223b90f145fb16d7e9d7b8bcbac. The one centralized X.509 provider-origin classifier now includes AWS dual-stack and partition suffix families: api.aws, api.amazonwebservices.com.cn, c2s.ic.gov, sc2s.sgov.gov, cloud.adc-e.uk, csp.hci.ic.gov, amazonaws.eu, and api.amazonwebservices.eu. Matching still uses the existing IDN-normalized, trailing-dot-trimmed, dot-delimited OrdinalIgnoreCase predicate, so intentional non-provider HTTPS OpenAI gateways and non-X.509 clients are unchanged.

Regression coverage adds commercial dual-stack, case/trailing-dot/port, China dual-stack, isolated partitions, and EU partition endpoints to ClientRejectsProviderOwnedEndpointsBeforeTokenAcquisition; each is rejected before token acquisition or API dispatch. The focused 17-test security slice passed on .NET 8/9/10; MutualTls passed 170/182/182; full suites passed 1728/1740/1742.

PR #1293, opened 2026-08-16
Use shared pagination option models in TypeSpec

### LINKED ISSUE #1110: Use shared pagination option models instead of redeclaring pagination parameters in TypeSpec

wei-core:
Closes #1110

## Summary

- Reuse the shared collection option models across the 13 affected pagination operations instead of redeclaring their query parameters.
- Keep REST-facing names (`limit`, `after`, and `before`) in the base TypeSpec and apply .NET names (`pageSizeLimit`, `afterId`, and `beforeId`) to the shared option models and generated internal plumbing.
- Preserve the legacy public protocol-overload parameter names (`limit`, `after`, and `before`) so existing C# callers that use named arguments remain source-compatible.
- Preserve the original wire/query ordering and the existing positional signatures of the model factory helpers.
- Regenerate the affected clients and API listings, remove obsolete duplicate pagination declarations, and add coverage for visitor behavior, model-factory ordering, query serialization order, and named-argument source compatibility.

This follows up on #1152 and incorporates the review feedback from that attempt, including keeping wire-name aliases out of the options-replacement path in `PaginationVisitor`, retaining the vector-store query decorators and suppressions, preserving model-factory parameter order, and retaining the existing public named-argument surface.

## API impact

The shared pagination option types use the .NET property names `PageSizeLimit`, `AfterId`, and `BeforeId`. Public protocol overloads retain their existing parameter types, positions, and legacy names `limit`, `after`, and `before`, so both binary compatibility and C# named-argument source compatibility are preserved. Existing model-factory positional ordering remains unchanged.

## Contribution roles

I led the contribution direction and acceptance criteria, reviewed and approved
the final design, and reviewed every changed line. I understand the change and
own the final design decision and the submission. AI assistance was used for
repository research, root-cause analysis, implementation preparation, and
automated validation execution.

## Validation

- TypeSpec compilation and code generation completed successfully on the latest `main`
- Generator tests: 28 passed
- Legacy pagination named-argument compile test: passed
- `OpenAI.Tests` compiled successfully for net10.0, net9.0, and net8.0
- Experimental attribute validation: passed
- API listings regenerated for net10.0, net8.0, and netstandard2.0
- Pagination query-order mock test: 1 passed
- Full net10.0 Playback suite from the rebased PR: 1,579 passed, 0 failed, 87 skipped


sylvesterkaczmarek (commented):
The overload types/positions stay the same, but renaming public parameters is source-breaking for C# callers that use named arguments (`limit:`, `after:`, etc.). The binary/API compatibility checks won’t catch those call sites. Can the old named-argument surface be preserved somehow, or should this be staged/documented as a breaking change?

wei-core:
@sylvesterkaczmarek Good catch — the parameter-name changes were source-breaking for named-argument callers even though the binary signatures were unchanged.

Fixed in [`9a416c2`](https://github.com/openai/openai-dotnet/pull/1293/commits/9a416c217c1798922f5e2a67767acaba4c4c9fa8):

- Public protocol overloads now retain the legacy `limit`, `after`, and `before` parameter names.
- Shared pagination option types still use `PageSizeLimit`, `AfterId`, and `BeforeId`, and the generated internal plumbing keeps the new SDK-side names.
- The generator rule is scoped to the exact affected public client methods.
- Added a compile-level regression test that calls every affected sync/async overload with the legacy named arguments, plus generator visitor coverage.

Validation: all 28 generator tests pass; the named-argument regression passes; the test project compiles for net10.0, net9.0, and net8.0; TypeSpec regeneration, experimental-attribute validation, and API export all completed successfully.

Could you take another look?

sylvesterkaczmarek (commented):
Re-checked the named-argument compatibility concern. The affected public protocol overloads retain the legacy `limit`/`after`/`before` names while the generated internal plumbing and shared option models use the new SDK names, and the generator has compile-level named-argument coverage. My source-compatibility finding is resolved.

wei-core:
Maintenance check: the named-argument compatibility concern has been fixed and the reviewer confirmed it is resolved. GitHub reports the PR as mergeable with no unresolved review threads; the remaining blocker is maintainer review/approval.

Could one of the requested reviewers take a final look when convenient? I am happy to refresh from main and regenerate if the branch needs another update.

joseharriaga (approved):

joseharriaga:
Thank you very much for your contribution, @wei-core ! Approved and merged. 🙂

PR #1299, opened 2026-08-17
[EngSys] Update .NET SDK Version

jsquire:
The focus of these changes is to bump the version of the .NET SDK required by the repository to mitigate a CVE in earlier versions of the SDK.

### DIFF
--- a/.github/workflows/live-test.yml
+++ b/.github/workflows/live-test.yml
@@ -19,10 +19,13 @@ jobs:
     env:
       version_suffix_args: ${{ format('/p:VersionSuffix="alpha.{0}"', github.run_number) }}
     steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
       - name: Setup .NET 10
         uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
         with:
-          dotnet-version: '10.0.400'
+          global-json-file: global.json
 
       - name: Setup .NET 9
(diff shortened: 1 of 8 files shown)


jsquire:
@copilot:  Keep global.json pinned to .NET SDK 10.0.400. Fix the Build workflow by updating the “Restore tools” step in .github/workflows/main.yml to run `dotnet tool restore` without `--add-source`. The Azure SDK development feed is already declared and correctly mapped in nuget.config, and NuGet 10 rejects command-line sources when package source mapping is enabled. Make no unrelated changes, then run the tool restore and relevant build/playback tests to verify the fix.

jsquire:
@copilot:  Ensure that the repository documentation (contributing guide, readme, etc) are updated to reflect the .NET SDK version bump, where an older version is referenced.

.github/workflows/live-test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Setup .NET 10
         uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
         with:
-          dotnet-version: '10.x'
+          dotnet-version: '10.0.400'
```
jsquire: @copilot:  Explain why these overrides are needed.   If `global.json` already specifies this at the minimum level, the task should automatically default to the latest - which in this case is .400.

jsquire: @copilot:  We need an approach that does not require hardcoding in every action.  Propose a solution.

jsquire: @copilot:  Approved.  Implement the proposal

joseharriaga (approved):

PR #1346, opened 2026-08-28
Fix the nullability of the `Status` property of Responses items

joseharriaga:
The `status` properties of Responses items were not defined as optional in the base TypeSpec. This error in the spec remained hidden because the properties went through a series of overlapping customizations in client.tsp, the `OpenAILibraryVisitor`, and Custom code that correctly exposed the `Status` as nullable in the .NET library. However, these same customizations resulted in incorrect deserialization code that would fail when the value of `status` was null. In other words: Even though the property was nullable, our deserialization logic was not accepting `null` as valid. This change fixes the base TypeSpec and removes the overlapping customizations.


specification/client/responses.client.tsp
```diff
@@ -27,30 +27,16 @@ using TypeSpec.HttpClient.CSharp;
 @@usage(ItemResource, Usage.input | Usage.output);
 @@visibility(ItemResource.id, Lifecycle.Read);
 
-@@visibility(CodeInterpreterToolCallItemResource.status, Lifecycle.Read);
 @@clientName(CodeInterpreterToolImageOutput.url, "ImageUri");
```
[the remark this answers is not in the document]
joseharriaga: This is the expected behavior. The service accepts passing a `status` as an input.

codegen/generator/src/OpenAILibraryVisitor.cs
```diff
@@ -38,17 +38,6 @@ public class OpenAILibraryVisitor : ScmLibraryVisitor
             [
                 new("Id", "id", isCollection: false),
             ],
-        ["ApplyPatchCallItem"] = [_readonlyStatusReplacementInfo],
-        ["CodeInterpreterCallResponseItem"] = [_readonlyStatusReplacementInfo],
```
jsquire, line 42: 🟡 The status-specific entries have been removed from `TypeNameToWritePropertyNameAdditionalConditionMap`, which leaves `_readonlyStatusReplacementInfo` without any remaining references. Could we remove this field as part of the cleanup?

---

> [!NOTE]

joseharriaga: @copilot Please remove the `_readonlyStatusReplacementInfo` field from the `OpenAILibraryVisitor`.

jsquire (approved):

Issue #653, opened 2025-09-04
[FEATURE REQ] Observability for Responses API


tomasz-tomczykiewicz:
### Describe the feature or improvement you are requesting

It would be very useful to add observability features to the Responses API, including support for streaming.

jsquire:
Hi @tomasz-tomczykiewicz.   Thanks for reaching out and for your request.  Our intent is to instrument all operations for observability.   However, because AI observability itself is still in an experimental state and changing, this is something that isn't on our near-term roadmap.  We'll continue to monitor that status and reconsider prioritization over time as the spec becomes more stable and our library more mature.

PR #1347, opened 2026-08-28
Add Responses OpenTelemetry instrumentation

jsquire:
# Summary
 
 This change adds experimental OpenTelemetry instrumentation for non-streaming `ResponsesClient.CreateResponse` and `CreateResponseAsync` operations.
 
The implementation builds on the existing Chat telemetry infrastructure while using a separate `OpenAI.ResponsesClient` activity source and meter. It follows the current OpenTelemetry Generative AI semantic conventions and supports both the default convention behavior and `OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental`.
 
 Partially addresses #653. Streaming Responses telemetry remains outside the scope of this change and will be added in a follow-up pull request after this merges..
 
 ## Details
 
 The new instrumentation:
 
 - Creates client spans for strongly typed, non-streaming Responses operations.
 - Records `gen_ai.operation.name` with the `chat` operation.
 - Uses `chat {model}` as the span name when the requested model is available and `chat` when it is not.
 - Records request model, response model, response ID, supported request parameters, finish reasons, token usage, service tier, reasoning effort, conversation details, and compaction state where defined by the active semantic conventions.
 - Emits `gen_ai.client.operation.duration` and `gen_ai.client.token.usage` metrics through the `OpenAI.ResponsesClient` meter.
 - Marks failed and cancelled responses as errors and records low-cardinality error types without capturing provider-controlled error messages.
 - Makes sampling-relevant attributes available when the span is created.
 - Supports requests where the model is supplied implicitly, such as future saved-prompt scenarios, while preserving an explicitly supplied model as the requested model.
 
 Telemetry is attached only to the strongly typed non-streaming methods. Generated protocol methods, streaming response creation, stored-response retrieval, cancellation, deletion, and input-item listing are not instrumented by this change.
 
 ## Privacy
 
 The implementation does not record prompts, generated output, instructions, tool definitions, tool arguments or results, metadata, safety identifiers, end-user identifiers, prompt cache keys, multimodal payloads, or provider-controlled error messages.
 
 Enabling experimental OpenTelemetry support does not opt callers into content capture.
 
 ## Testing
 
 The synthetic telemetry coverage includes:
 
 - Synchronous and asynchronous response creation.
 - Default and latest experimental semantic conventions.
 - Metrics enabled without tracing.
 - Tracing enabled without metrics.
 - Request and response attributes.
 - Sampling-time attributes.
 - Duration and token usage metrics.
 - Requests without an explicit model.
 - HTTP failures and response status mapping.
 - Completed, incomplete, failed, and cancelled responses.
 - Unknown future incomplete reasons.
 - Response compaction.
 - Sensitive-data exclusion.
 - Protocol method exclusion.
 - Convenience overload deduplication.
 
 The observability documentation and changelog have also been updated to describe the new coverage and its current limitations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Thank you to our developer community members who helped to make the OpenAI clien
 - OpenAI.Chat:
   - Added opt-in support for the latest experimental OpenTelemetry GenAI semantic conventions supported by this library. Set `OTEL_SEMCONV_STABILITY_OPT_IN` to include `gen_ai_latest_experimental` to emit `gen_ai.provider.name` instead of `gen_ai.system`. The default remains compatible with OpenTelemetry GenAI Semantic Conventions v1.27.0. GenAI histograms now also advertise the recommended explicit bucket boundaries. _(A community contribution, courtesy of [trask](https://github.com/trask))_
 - OpenAI.Responses:
+  - Added experimental OpenTelemetry traces and metrics for non-streaming `CreateResponse` operations. The instrumentation follows the configured GenAI semantic convention version and does not capture prompts, generated content, instructions, tools, or end-user identifiers.
   - Added `ResponseReasoningContext` and exposed it through `ResponseReasoningOptions.Context`, allowing the amount of reasoning context preserved across turns to be controlled with `Auto`, `CurrentTurn`, or `AllTurns`. _(A community contribution, courtesy of [hogeheer499-commits](https://github.com/hogeheer499-commits))_
 
 ### Bugs Fixed
--- a/OpenAI.Responses/src/Custom/ResponsesClient.cs
+++ b/OpenAI.Responses/src/Custom/ResponsesClient.cs
@@ -1,4 +1,5 @@
 using Microsoft.TypeSpec.Generator.Customizations;
+using OpenAI.Telemetry;
 using System;
 using System.ClientModel;
 using System.ClientModel.Primitives;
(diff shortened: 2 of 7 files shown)


jsquire:
@lmolkova:   Is there any chance that you may be able to clarify the expected OpenTelemetry behavior for an OpenAI Responses request where the SDK does not receive a model value?

The base `gen_ai.inference.client` span defines `gen_ai.request.model` as conditionally required when available. The OpenAI-specific `openai.inference.client` refinement overrides that inherited requirement and marks the attribute as required:

```yaml
span_refinements:
  - id: openai.inference.client
    ref: gen_ai.inference.client
    attributes:
      - ref: gen_ai.request.model
        requirement_level: required
```

The generated OpenAI documentation also lists `gen_ai.request.model` as required and specifies the span name as `{gen_ai.operation.name} {gen_ai.request.model}`.

I'm not clear what the right behavior is when a valid Responses request omits the client-side `model` parameter from the client while still using a model selected by the service. For example, a saved prompt can configure the effective model server-side. The SDK knows the prompt identifier, but it does not know the model selected by that prompt when it starts the activity. Inferring a model from the prompt identifier would be inaccurate, and `gen_ai.response.model` is not available until after the request completes, which is too late for head-based sampling.

In this case, should client instrumentation:

1. Create the `openai.inference.client` span without `gen_ai.request.model`
2. Omit inference instrumentation for the valid operation
3. Use a different convention or attribute pattern

Is there existing guidance, or should the OpenAI refinement explicitly describe model selection delegated to a saved prompt or other server-side request state?  Thank you for any insights!

Sources:

- [`gen_ai.inference.client` and `openai.inference.client`](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/spans.yaml)
- [OpenAI GenAI semantic conventions](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/openai.md)

joseharriaga (approved):

PR #1354, opened 2026-09-02
Convert Realtime models to protocol models

joseharriaga:
- Added a `protected internal` constructor to the `RealtimeClientCommand` base class and exposed its `Kind` property of a new extensible enum type called `RealtimeClientCommandKind` to support extensibility.
- Added a `protected internal` constructor to the `RealtimeServerUpdate` base class and exposed its `Kind` property of a new extensible enum type called `RealtimeServerUpdateKind` to support extensibility.
- Added public setters to the applicable properties of Realtime models.
- Added public default constructors to the applicable Realtime models.
- Renamed `RealtimeLogProbabilityDetails` to `RealtimeTokenLogProbabilityDetails` for consistency.
- Renamed the `Logprobs` property of `RealtimeServerUpdateConversationItemInputAudioTranscriptionCompleted` to `TranscriptionTokenLogProbabilities` for consistency.
- Renamed the `Logprobs` property of `RealtimeServerUpdateConversationItemInputAudioTranscriptionDelta` to `TranscriptionTokenLogProbabilities` for consistency.


OpenAI/src/Generated/OpenAIModelFactory.cs
```diff
@@ -2764,51 +2764,51 @@ public static RealtimeMaxOutputTokenCount RealtimeMaxOutputTokenCount(RealtimeDe
 
         public static RealtimeServerUpdate RealtimeServerUpdate(string kind = default)
         {
-            return new InternalUnknownRealtimeServerEventGA(new InternalRealtimeServerEventTypeGA(kind), default);
+            return new InternalUnknownRealtimeServerEventGA(new RealtimeServerUpdateKind(kind), default);
```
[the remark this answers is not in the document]
joseharriaga: The model factory is internal and not needed because we're using protocol models.

codegen/generator/src/Visitors/ProtocolModelVisitor.cs
```diff
@@ -19,6 +19,7 @@ public class ProtocolModelVisitor : ScmLibraryVisitor
     {
         "OpenAI.Containers",
         "OpenAI.Conversations",
+        "OpenAI.Realtime",
```
jsquire, line 22: 🟡 Should we add `OpenAI.Realtime` to the test cases in `ProtocolModelVisitorTests` and cover the `VisitType` constructor visibility path? This allowlist now drives setters and public parameterless constructors for the Realtime protocol models. The existing tests enumerate the other protocol namespaces but omit Realtime, and `VisitType`, which produces much of the new API surface in this PR, does not have direct coverage. A focused visitor test would help ensure that future generator changes do not silently alter this API shape.

---

> [!NOTE]

OpenAI/src/Custom/Realtime/GA/ServerUpdate/RealtimeServerUpdateConversationItemInputAudioTranscriptionCompleted.cs
```diff
@@ -20,4 +21,7 @@ namespace OpenAI.Realtime;
 [CodeGenType("RealtimeServerEventConversationItemInputAudioTranscriptionCompletedGA")]
 public partial class RealtimeServerUpdateConversationItemInputAudioTranscriptionCompleted
 {
+    // CUSTOM: Renamed.
+    [CodeGenMember("Logprobs")]
+    public IList<RealtimeTokenLogProbabilityDetails> TranscriptionTokenLogProbabilities { get; }
```
jsquire, line 26: 🟡 Should we add a serialization test that deserializes the `logprobs` wire member into `TranscriptionTokenLogProbabilities`, then writes the model and confirms that the wire name remains `logprobs`? This PR renames both the CLR type and property while preserving the protocol field. The generated mapping looks correct, but no Realtime test exercises deserialization or round-trip serialization through the renamed members. We should probably cover both the completed and delta update shapes.

---

> [!NOTE]

sylvesterkaczmarek (commented):
These renames remove existing public API names (`RealtimeLogProbabilityDetails` and `Logprobs`). That breaks source callers immediately, and removing the old type/property can also break binaries compiled against the current package. Could we keep obsolete compatibility aliases/forwarders for the old names while exposing the new ones?

jsquire:
@sylvesterkaczmarek:   The RealTime feature is still in an experimental state and, as a result, does not come with a backwards compatibility guarantee.  While we do not make breaking changes lightly, until an API surface graduates from experimental to stable it's a signal that we're iterating on the design and there will be changes that may be breaking.   Once stable, we provide a strong backwards-compatibility guarantee and would introduce compat shims such as you're suggesting.   More information can be found in [Feature lifecycle in the OpenAI .NET client library](https://github.com/openai/openai-dotnet/blob/main/docs/FeatureLifecycle.md#stable-and-experimental-apis)

In this case, RealTime is based on legacy patterns that we need to update in order to align with the current API direction and is expected to continue to evolve in non-trivial ways over the next couple of months.

sylvesterkaczmarek:
Understood, thanks for the lifecycle clarification. Given Realtime is explicitly experimental and does not carry the stable API compatibility guarantee yet, I withdraw the compatibility-shim request for these renames.

codegen/generator/src/Visitors/ProtocolModelVisitor.cs
joseharriaga: Fixed!

OpenAI/src/Custom/Realtime/GA/ServerUpdate/RealtimeServerUpdateConversationItemInputAudioTranscriptionCompleted.cs
joseharriaga: Fixed!

tests/Realtime/RealtimeTests.cs
```diff
@@ -476,6 +475,73 @@ public async Task TranscriptionOnlyWorks()
...
+
+            if (deltaUpdate is not null && completedUpdate is not null)
+            {
+                break;
+            }
+        }
+
+        Assert.That(deltaUpdate, Is.Not.Null);
+        Assert.That(deltaUpdate!.TranscriptionTokenLogProbabilities, Has.Count.GreaterThan(0));
+        Assert.That(deltaUpdate.TranscriptionTokenLogProbabilities[0].Token, Is.Not.Null.And.Not.Empty);
```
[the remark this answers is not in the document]
joseharriaga: There is an assert that checks for null above.

tests/Realtime/RealtimeTests.cs
```diff
@@ -476,6 +475,73 @@ public async Task TranscriptionOnlyWorks()
...
+            IncludedProperties = { RealtimeIncludedProperty.ItemInputAudioTranscriptionLogProbabilities },
+        };
+
+        await sessionClient.ConfigureTranscriptionSessionAsync(options, CancellationToken);
+
+        // Sending the audio in a delayed stream allows us to validate bidirectional behavior, i.e.
+        // transcription data arriving while audio is still being sent.
+        string inputPath = Path.Join("Assets", "realtime_api_description_pcm16_24khz_mono.wav");
+        using TestDelayedFileReadStream inputStream = new(inputPath, TimeSpan.FromMilliseconds(50), readsBeforeDelay: 2);
+        _ = sessionClient.SendInputAudioAsync(inputStream, CancellationToken);
```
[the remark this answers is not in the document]
joseharriaga: Fixed.

OpenAI/src/Generated/Models/Realtime/RealtimeTranscriptionTokenUsage.cs
```diff
@@ -31,6 +31,6 @@ internal RealtimeTranscriptionTokenUsage(InternalTranscriptionTokenUsageBaseType
         }
 #pragma warning restore SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
 
-        public RealtimeTranscriptionInputTokenUsageDetails InputTokenDetails { get; }
+        public RealtimeTranscriptionInputTokenUsageDetails InputTokenDetails { get; set; }
     }
```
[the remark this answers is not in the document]
joseharriaga: This is a generator issue tracked here: https://github.com/microsoft/typespec/issues/11896

OpenAI/src/Custom/Realtime/GA/ClientCommand/RealtimeClientCommand.cs
```diff
@@ -7,6 +7,8 @@ namespace OpenAI.Realtime;
 /// </summary>
 // CUSTOM: Renamed.
 [CodeGenType("RealtimeClientEventGA")]
+[CodeGenVisibility(nameof(RealtimeClientCommand), CodeGenVisibility.ProtectedInternal, typeof(RealtimeClientCommandKind))]
```
jsquire, line 10: 🔴 Should we preserve patch data when serializing subclasses enabled by the new constructors?

A custom subclass can supply its discriminator and set `Patch.Set("$.extra"u8, "123"u8)`, but typed `SendCommandAsync` emits only the discriminator. I reproduced the field being silently omitted. The subclass must override `JsonModelWriteCore` and explicitly write the patch to retain it. The new `RealtimeServerUpdate` extensibility path behaves the same way.

We should probably cover this scenario with an offline regression test.

tests/Realtime/RealtimeTests.cs
```diff
@@ -476,6 +475,71 @@ public async Task TranscriptionOnlyWorks()
         Assert.That(gotSessionUpdated, Is.True);
     }
 
+    [Test]
+    public async Task TranscriptionTokenLogProbabilitiesWork()
```
jsquire, line 479: 🟡 Should we add the round-trip serialization coverage requested in the earlier comment?

This test checks returned tokens and bytes, but never serializes either update or verifies that the renamed property still writes `logprobs`. It is also live-only. A fixture-based test for both completed and delta updates would cover the wire mapping without a service connection.

Both shapes round-trip correctly in local probes, so this is a coverage gap rather than a demonstrated mapping defect.

OpenAI/src/Custom/Realtime/GA/ClientCommand/RealtimeClientCommand.cs
joseharriaga: If I'm understanding the concern correctly, this is expected. All sub-classes (custom or not) must override `JsonModelWriteCore` and call `Patch.WriteTo(writer)`.

Here's an example in one of our own sub-classes of `RealtimeClientCommand` called `RealtimeClientCommandConversationItemCreate`: 

https://github.com/openai/openai-dotnet/blob/f4249d63edfbfb90d83f8d5543fe420cd4e24021/OpenAI/src/Generated/Models/Realtime/RealtimeClientCommandConversationItemCreate.Serialization.cs#L92

tests/Realtime/RealtimeTests.cs
joseharriaga: I think this test (which exercises the deserialization of `logprobs` along with other features related to it) is sufficient. The change in question is a simple rename achieved through codegen, and codegen has its own tests to validate general renaming behavior.

jsquire (approved):

sylvesterkaczmarek:
Thanks for the lifecycle clarification earlier, and glad to see this landed cleanly. The experimental/stable distinction was useful context for reviewing this area.

PR #1357, opened 2026-09-04
docs: fix function calling links in ChatTool

z0rimo:
## Summary

Fixes four malformed function calling documentation links in `ChatTool.cs`.

The existing links include an extra `/api-reference/chat` path segment:

`https://platform.openai.com/docs/api-reference/chat/docs/guides/function-calling`

This updates them to the current function calling guide:

`https://platform.openai.com/docs/guides/function-calling`

## Validation

- Ran `git diff --check`
- Verified all four malformed URLs were replaced
- Verified the corrected URL matches the function calling documentation used by OpenAI's other SDKs
- Documentation-only change; no runtime behavior or public API surface changes


jsquire (approved):
Thank you for your contribution, @z0rimo, and for helping to improve the OpenAI developer experience.   The proposed changes look good.  I'll get them merged in.

PR #1360, opened 2026-09-04
fix(chat): preserve patches on single content parts

### LINKED ISSUE #1359: [BUG] Single-part message content is collapsed to a string on write, dropping the part's Patch properties (e.g. cache_control)
**Service:** Chat Completions (`OpenAI.Chat`)
**Package:** OpenAI 2.12.0 (also observed through Microsoft.Extensions.AI.OpenAI 10.8.3, which delegates to the same serializer)

### Describe the bug

`ChatMessageContentPart.Patch` lets a caller add properties a compatible endpoint understands (in our case Anthropic-style `"cache_control": {"type": "ephemeral"}`, which OpenRouter forwards to Anthropic for prompt caching). The patched property is written correctly **only when the message has two or more content parts**. When a message's content is a single text part, the serializer collapses the content to a plain JSON string and the part's `Patch` properties are silently dropped.

A message-level `Patch.Set("$.content", …)` does not help: the patched value is written next to the original `content` property, producing a duplicate key.

The single-part case is the common one (every `ToolChatMessage`, most system prompts), so there is currently no way to carry a per-part extension on those messages.

### Steps to reproduce

```csharp
#pragma warning disable SCME0001
using System.ClientModel.Primitives;
using OpenAI.Chat;

static BinaryData Ephemeral() => BinaryData.FromString("{\"type\":\"ephemeral\"}");

// 1. one text part with a Patch property -> collapsed to a string, Patch dropped
var single = ChatMessageContentPart.CreateTextPart("You are a helpful assistant.");
single.Patch.Set("$.cache_control"u8, Ephemeral());
Console.WriteLine(ModelReaderWriter.Write(new SystemChatMessage(single)));

// 2. two text parts, second with the same Patch -> array, Patch kept
var a = ChatMessageContentPart.CreateTextPart("part one");
var b = ChatMessageContentPart.CreateTextPart("part two");
b.Patch.Set("$.cache_control"u8, Ephemeral());
Console.WriteLine(ModelReaderWriter.Write(new UserChatMessage(a, b)));

// 3. tool message, one text part with a Patch property -> collapsed, Patch dropped
var tool = ChatMessageContentPart.CreateTextPart("{\"price\":1}");
tool.Patch.Set("$.cache_control"u8, Ephemeral());
Console.WriteLine(ModelReaderWriter.Write(new ToolChatMessage("call_1", tool)));

// 4. message-level Patch on $.content -> duplicate key
var msg = new SystemChatMessage("You are a helpful assistant.");
msg.Patch.Set("$.content"u8, BinaryData.FromString("[{\"type\":\"text\",\"text\":\"You are a helpful assistant.\",\"cache_control\":{\"type\":\"ephemeral\"}}]"));
Console.WriteLine(ModelReaderWriter.Write(msg));
```

Output:

```
1: {"role":"system","content":"You are a helpful assistant."}
2: {"role":"user","content":[{"type":"text","text":"part one"},{"type":"text","text":"part two","cache_control":{"type":"ephemeral"}}]}
3: {"role":"tool","tool_call_id":"call_1","content":"{\"price\":1}"}
4: {"role":"system","content":"You are a helpful assistant.","content":[{"type":"text","text":"You are a helpful assistant.","cache_control":{"type":"ephemeral"}}]}
```

### Expected behavior

Cases 1 and 3 should keep the array form whenever a part carries `Patch` properties, i.e.

```
{"role":"system","content":[{"type":"text","text":"You are a helpful assistant.","cache_control":{"type":"ephemeral"}}]}
```

The collapse to a plain string is a fine default for an unpatched single text part (byte-identical to what callers sent before), but when the caller has explicitly patched the part, dropping the patch defeats the purpose of `Patch`. Alternatively, a message-level `Patch` on `$.content` should replace the property rather than duplicate it.

### Context

This matters for OpenAI-compatible endpoints such as OpenRouter, where `cache_control` on content parts is the documented way to get Anthropic prompt caching, and the natural place for a breakpoint is the last tool result of a turn — always a single-part message. With the current behaviour, users of `Microsoft.Extensions.AI.OpenAI` who need this have to bypass the SDK's request serialization entirely.

Happy to open a PR if the maintainers agree on the direction (keep the array form when any part has `Patch` properties).

sylvesterkaczmarek:
### Summary

Preserve `ChatMessageContentPart.Patch` properties when a chat message contains exactly one text content part.

The serializer previously collapsed every single text part to a plain JSON string. That compact representation cannot carry part-level extension fields, so explicitly patched properties such as `cache_control` were silently dropped. Single text parts with patch operations now retain the array/object representation, while ordinary unpatched single text messages keep the existing string serialization.

### Tests

- Added SystemChatMessage and ToolChatMessage regressions for a patched single text part.
- Existing unpatched single-message serialization remains covered.
- `dotnet build OpenAI.slnx` — succeeded with 0 warnings and 0 errors.
- `ChatSmokeTests` on net10.0 — 74 passed.
- Focused patched/unpatched serialization tests on net10.0 — 4 passed.
- `git diff --check` — passed.

Fixes #1359.

### DIFF
--- a/OpenAI/src/Custom/Chat/ChatMessageContent.Serialization.cs
+++ b/OpenAI/src/Custom/Chat/ChatMessageContent.Serialization.cs
@@ -12,7 +12,7 @@ public partial class ChatMessageContent
         {
             writer.WriteNullValue();
         }
-        else if (Count == 1 && this[0].Kind == ChatMessageContentPartKind.Text)
+        else if (Count == 1 && this[0].Kind == ChatMessageContentPartKind.Text && !HasPatchOperations(this[0]))
         {
             writer.WriteStringValue(this[0].Text);
         }
@@ -26,6 +26,13 @@ public partial class ChatMessageContent
             writer.WriteEndArray();
         }
     }
+
+#pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
+    private static bool HasPatchOperations(ChatMessageContentPart part)
+    {
+        return part.Patch.ToBinaryData().ToMemory().Length > 2;
+    }
+#pragma warning restore SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
 
     internal static ChatMessageContent DeserializeChatMessageContent(JsonElement element, ModelReaderWriterOptions options = null)
     {
--- a/OpenAI/src/Custom/Chat/ChatMessageContent.Serialization.cs
+++ b/OpenAI/src/Custom/Chat/ChatMessageContent.Serialization.cs
@@ -30,7 +30,7 @@ public partial class ChatMessageContent
 #pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
     private static bool HasPatchOperations(ChatMessageContentPart part)
     {
-        return part.Patch.ToBinaryData().ToMemory().Length > 2;
+        return part.Patch.Contains("$"u8) || part.Patch.ToBinaryData().ToMemory().Length > 2;
     }
 #pragma warning restore SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
 
--- a/OpenAI/src/Custom/Chat/ChatMessageContentPart.Serialization.cs
+++ b/OpenAI/src/Custom/Chat/ChatMessageContentPart.Serialization.cs
@@ -23,37 +23,40 @@ public partial class ChatMessageContentPart : IJsonModel<ChatMessageContentPart>
 #pragma warning restore SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
 
         writer.WriteStartObject();
-        writer.WritePropertyName("type"u8);
-        writer.WriteStringValue(instance._kind.ToSerialString());
+#pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
+        if (!instance.Patch.Contains("$.type"u8))
+        {
+            writer.WritePropertyName("type"u8);
+            writer.WriteStringValue(instance._kind.ToSerialString());
+        }
 
-        if (instance._kind == ChatMessageContentPartKind.Text)
+        if (instance._kind == ChatMessageContentPartKind.Text && !instance.Patch.Contains("$.text"u8))
         {
             writer.WritePropertyName("text"u8);
             writer.WriteStringValue(instance._text);
         }
-        else if (instance._kind == ChatMessageContentPartKind.Refusal)
+        else if (instance._kind == ChatMessageContentPartKind.Refusal && !instance.Patch.Contains("$.refusal"u8))
         {
             writer.WritePropertyName("refusal"u8);
             writer.WriteStringValue(instance._refusal);
         }
-        else if (instance._kind == ChatMessageContentPartKind.Image)
+        else if (instance._kind == ChatMessageContentPartKind.Image && !instance.Patch.Contains("$.image_url"u8))
         {
             writer.WritePropertyName("image_url"u8);
             writer.WriteObjectValue(instance._imageUri, options);
         }
-        else if (instance._kind == ChatMessageContentPartKind.InputAudio)
+        else if (instance._kind == ChatMessageContentPartKind.InputAudio && !instance.Patch.Contains("$.input_audio"u8))
         {
             writer.WritePropertyName("input_audio"u8);
             writer.WriteObjectValue(instance._inputAudio, options);
         }
-        else if (instance._kind == ChatMessageContentPartKind.File)
+        else if (instance._kind == ChatMessageContentPartKind.File && !instance.Patch.Contains("$.file"u8))
         {
             writer.WritePropertyName("file"u8);
             writer.WriteObjectValue(instance._fileFile, options);
         }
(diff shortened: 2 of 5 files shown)


OpenAI/src/Custom/Chat/ChatMessageContent.Serialization.cs
```diff
@@ -27,6 +27,13 @@ internal void WriteTo(Utf8JsonWriter writer, ModelReaderWriterOptions options =
         }
     }
 
+#pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
+    private static bool HasPatchOperations(ChatMessageContentPart part)
+    {
+        return part.Patch.ToBinaryData().ToMemory().Length > 2;
+    }
```
jsquire, line 34: 🔴 Can we please check for retained patch entries rather than whether the serialized JSON Patch document contains operations?

`ToBinaryData()` omits operations whose values match the original JSON. A deserialized text part can therefore retain `cache_control` while this helper returns false. I confirmed that writing a patched message, reading it back, and writing it again drops the extension and produces plain string content.

Please include that round trip in the regression coverage. The upstream query discussed in #1359 needs to account for these retained properties, even when they are unchanged from the original JSON.

OpenAI/src/Custom/Chat/ChatMessageContent.Serialization.cs
```diff
@@ -12,7 +12,7 @@ internal void WriteTo(Utf8JsonWriter writer, ModelReaderWriterOptions options =
         {
             writer.WriteNullValue();
         }
-        else if (Count == 1 && this[0].Kind == ChatMessageContentPartKind.Text)
+        else if (Count == 1 && this[0].Kind == ChatMessageContentPartKind.Text && !HasPatchOperations(this[0]))
```
jsquire, line 15: 🔴 Can we please preserve the existing string-or-null representation for `FunctionChatMessage`?

The shared writer also handles that deprecated role. Patching its text part now causes it to emit a content array, which the [current REST specification](https://github.com/openai/openai-openapi/blob/2c5215a6a7c08f01cf8c9f8c76a7df5f34a8b30d/openapi.yaml#L31387-L31409) does not permit. For example, this now writes an array instead of a string.

```csharp
var message = new FunctionChatMessage("get_price", "1");
message.Content[0].Patch.Set("$.custom_property"u8, true);
var json = ModelReaderWriter.Write(message);
```

Please add a legacy function-message regression alongside the system and tool cases.

jsquire, line 15: 🔴 Can we please also update the content-part writer to honor patch precedence when taking this path?

The object writer currently writes modeled properties before appending their patched replacements. Sending a single text part through it exposes that behavior for the following case.

```csharp
var part = ChatMessageContentPart.CreateTextPart("original");
part.Patch.Set("$.text"u8, "replacement");
var json = ModelReaderWriter.Write(new SystemChatMessage(part));
```

The resulting content contains both values.

```json
[{"type":"text","text":"original","text":"replacement"}]
```

The missing precedence checks predate this PR, but need to be addressed before routing single patched text parts through this writer. Please include coverage for replacement, removal, and null, with assertions that each property name occurs only once.

OpenAI/src/Custom/Chat/ChatMessageContent.Serialization.cs
```diff
@@ -27,6 +27,13 @@ internal void WriteTo(Utf8JsonWriter writer, ModelReaderWriterOptions options =
         }
     }
 
+#pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
+    private static bool HasPatchOperations(ChatMessageContentPart part)
+    {
+        return part.Patch.ToBinaryData().ToMemory().Length > 2;
```
jsquire: 🟡 Can we please use the proposed System.ClientModel query rather than serialize the patch to determine whether it is empty?

This check runs for every single-text message, including ordinary unpatched messages. In a warmed diagnostic against this branch, 1,000 empty-patch checks allocated approximately 16 MiB, before serializing any message content. Patched messages also serialize their extension data here and then again when writing the part.

An inexpensive state query would let us preserve the existing single-text optimization without adding this allocation cost.

jsquire (changes requested):
🟡 Can we please also include the message-level `$.content` replacement fix in this PR?

The PR fixes the constructed single-part examples, but a message-level content patch still produces duplicate `content` properties. This is pre-existing behavior, rather than a regression from this change, but it is in the scope of #3159. It can be corrected independently of the upstream patch-state query.

OpenAI/src/Custom/Chat/ChatMessageContent.Serialization.cs
sylvesterkaczmarek: Addressed in `9d14930`. `FunctionChatMessage` now keeps its string/null content representation even when its content part carries patch data, with regression coverage.

sylvesterkaczmarek: Addressed in `9d14930`. Modeled content-part properties now defer to patch values; regression coverage includes replacement, removal, null, and duplicate-property checks.

sylvesterkaczmarek:
Pushed `9d14930`: message-level `$.content` precedence, content-part patch precedence, and the legacy `FunctionChatMessage` shape are covered. The remaining efficient patch-entry/round-trip case still depends on the System.ClientModel query discussed in #1359.

PR #1364, opened 2026-09-07
fix(realtime): implement synchronous ReceiveUpdates

### LINKED ISSUE #1363: [BUG] RealtimeSessionClient.ReceiveUpdates throws NotImplementedException
### Describe the bug

The synchronous `RealtimeSessionClient.ReceiveUpdates` APIs are part of the public API surface, but both currently throw `NotImplementedException`.

The affected overloads are:

- `ReceiveUpdates(CancellationToken)`
- `ReceiveUpdates(RequestOptions)`

The corresponding `ReceiveUpdatesAsync` overloads are implemented and use the existing WebSocket receive pipeline.

This is also inconsistent with other synchronous Realtime APIs such as `SendCommand`, which already use sync-over-async because `ClientWebSocket` does not expose a synchronous send operation.

The synchronous `ReceiveUpdates` methods are present in the released API surface, so callers can discover and invoke them normally, but they currently fail immediately.

I confirmed this on both OpenAI 2.13.0 and the current `main` branch.

If implementing the synchronous paths using the existing asynchronous WebSocket receive pipeline is the intended approach, I'd be happy to contribute a fix and regression coverage.

### Steps to reproduce

1. Use OpenAI 2.13.0 or the current `main` branch.
2. Obtain a `RealtimeSessionClient`.
3. Call either synchronous `ReceiveUpdates` overload:
   - `ReceiveUpdates(CancellationToken)`
   - `ReceiveUpdates(RequestOptions)`
4. Observe that the method throws `NotImplementedException`.
5. Compare with the corresponding `ReceiveUpdatesAsync` overloads, which are implemented.

### Code snippets

```C#
#pragma warning disable OPENAI002

using System.ClientModel;
using OpenAI.Realtime;

var session = new TestRealtimeSessionClient();

// Throws NotImplementedException.
_ = session.ReceiveUpdates();

sealed class TestRealtimeSessionClient : RealtimeSessionClient
{
    public TestRealtimeSessionClient()
        : base(
            new ApiKeyCredential("test"),
            new Uri("wss://example.com"),
            "gpt-realtime",
            intent: null,
            parentClient: null)
    {
    }
}
```

### OS

Windows 11

### .NET version

10.0.109

### Library version

OpenAI 2.13.0 and current main (19d0a3cb8e0cf0f3137a5c56c3c70a0c3f6c96f5)

sylvesterkaczmarek:
## Summary

- implement both synchronous `RealtimeSessionClient.ReceiveUpdates` overloads
- reuse the existing WebSocket receive pipeline and typed update deserialization
- avoid synchronization-context capture in the receive path used by sync-over-async
- add regressions for both protocol and typed synchronous enumeration

Fixes #1363.

## Validation

- `dotnet test tests/OpenAI.Tests.csproj --filter FullyQualifiedName~RealtimeUnitTests --no-restore`
- 12/12 passed on .NET 8, .NET 9, and .NET 10
- `git diff --check`

### DIFF
--- a/OpenAI/src/Custom/Realtime/Internal/AsyncWebsocketMessageEnumerator.cs
+++ b/OpenAI/src/Custom/Realtime/Internal/AsyncWebsocketMessageEnumerator.cs
@@ -38,7 +38,7 @@ internal partial class AsyncWebsocketMessageResultEnumerator : IAsyncEnumerator<
         WebsocketPipelineResponse websocketPipelineResponse = new();
         for (int partialMessageCount = 1; !websocketPipelineResponse.IsComplete; partialMessageCount++)
         {
-            WebSocketReceiveResult receiveResult = await _webSocket.ReceiveAsync(new(_receiveBuffer), _cancellationToken);
+            WebSocketReceiveResult receiveResult = await _webSocket.ReceiveAsync(new(_receiveBuffer), _cancellationToken).ConfigureAwait(false);
             if (receiveResult.CloseStatus.HasValue)
             {
                 Current = null;
--- a/OpenAI/src/Custom/Realtime/RealtimeSessionClient.Protocol.cs
+++ b/OpenAI/src/Custom/Realtime/RealtimeSessionClient.Protocol.cs
@@ -103,7 +103,7 @@ public partial class RealtimeSessionClient
         {
             _receiveCollectionResult ??= new(WebSocket, options?.CancellationToken ?? default);
         }
-        await foreach (ClientResult result in _receiveCollectionResult)
+        await foreach (ClientResult result in _receiveCollectionResult.ConfigureAwait(false))
         {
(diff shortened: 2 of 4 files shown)


jsquire:
Hi @sylvesterkaczmarek. Thank you for your contribution and your interest in improving the OpenAI developer experience.

Our position from #1227 has not changed. We do not intend to support a synchronous API surface for Realtime, so we will not be moving forward with this implementation. The existing synchronous stubs are remnants of an earlier design, rather than an indication that synchronous receiving is intended to be supported.

We recognize that leaving these methods discoverable in the released package is confusing. We have opened #1370 to track removing the synchronous surface and aligning Realtime with its asynchronous-only direction. For receiving updates today, callers should use `ReceiveUpdatesAsync`.

Thank you, again, for taking the time to investigate the issue and contribute regression coverage.

---
> [!NOTE]

PR #1372, opened 2026-09-10
Remove synchronous Realtime APIs

### LINKED ISSUE #1370: Remove synchronous APIs from OpenAI.Realtime
## Context

Realtime is intended to have an asynchronous-only API surface. The decision recorded in #1227 still stands.

The released experimental `OpenAI.Realtime` surface retains synchronous operations from an earlier design. Both `RealtimeSessionClient.ReceiveUpdates` overloads throw `NotImplementedException`, while other synchronous operations are implemented, including sync-over-async wrappers. This makes the intended support difficult for callers to understand and has led to #1363 and the proposed implementation in #1364.

We should remove the synchronous operation surface rather than implement the receive stubs.

## Scope

- Remove synchronous service and session operations from `RealtimeClient` and `RealtimeSessionClient`, covering both typed and protocol overloads. This includes session creation, client-secret creation, receiving updates, sending commands and audio, and the synchronous session configuration and conversation helpers.
- Remove the synchronous connection path and supporting sync-only implementation that is no longer needed.
- Preserve the asynchronous operations and their existing behavior, including cancellation, update delivery, and resource cleanup.
- Retain ordinary synchronous object functionality such as constructors, properties, events, model helpers, and `IDisposable.Dispose`. These are not the operation surface being removed.

## Completion criteria

- The synchronous operation methods are absent from the public and protected API surface across supported target frameworks.
- API baselines and any applicable generation customizations are updated so regeneration does not restore the removed operations.
- Tests and samples use the asynchronous surface, and existing asynchronous coverage continues to pass.
- Release notes and migration guidance explain the removal of the previously released experimental methods and point callers to their asynchronous equivalents.

Related discussions are #1227, #1363, and #1364. This issue tracks the API removal separately from the report that the synchronous receive methods throw.

---
> [!NOTE]

z0rimo:
## Summary

Removes the synchronous operation surface from `OpenAI.Realtime`, making the service and session operations on `RealtimeClient` and `RealtimeSessionClient` async-only.

This includes:

* removing synchronous session creation, command/audio sending, session configuration, conversation item, response, and update-receiving APIs
* removing the synchronous Realtime connection path and sync-only semaphore helper
* suppressing the generated synchronous `CreateRealtimeClientSecret` overloads so they are not restored by code generation
* migrating the remaining Realtime tests to the async client-secret API
* updating the in-progress API baselines and migration guidance

Async APIs and ordinary synchronous object functionality such as constructors, properties, events, and `IDisposable.Dispose` are unchanged.

Fixes #1370.

## Validation

* `Invoke-CodeGen.ps1 -Clean` completed successfully, including experimental API validation
* code generation does not restore the removed synchronous `CreateRealtimeClientSecret` overloads
* `Export-Api.ps1` completed successfully
* `api/released/**` is unchanged
* library and test projects build with 0 errors and 0 warnings
* Realtime tests (`net10.0`): 19 passed, 0 failed, 40 skipped (`LiveOnly`)
* codegen generator tests: 16 passed, 0 failed
* `git diff --check` passes

The skipped Realtime tests require live/recorded service infrastructure that was not available in the local environment.

### DIFF
--- a/tests/Realtime/RealtimeTests.cs
+++ b/tests/Realtime/RealtimeTests.cs
@@ -19,6 +19,7 @@ namespace OpenAI.Tests.Realtime;
 #pragma warning disable OPENAI002
 
 [LiveOnly(Reason = "Test framework doesn't support recording with web sockets yet")]
+[TestFixture(true)]
 public class RealtimeTests : RealtimeTestFixtureBase
 {
     public enum TestAudioSendType { WithAudioStreamHelper, WithManualAudioChunks }
--- a/tests/Realtime/RealtimeTests.cs
+++ b/tests/Realtime/RealtimeTests.cs
@@ -1262,7 +1262,7 @@ public class RealtimeTests : RealtimeTestFixtureBase
             SessionOptions = conversationSessionOptions,
         };
 
-        CreateClientSecretResult result = client.CreateRealtimeClientSecret(createClientSecretOptions);
+        CreateClientSecretResult result = await client.CreateRealtimeClientSecretAsync(createClientSecretOptions, CancellationToken);
 
         Assert.That(result, Is.Not.Null);
         Assert.That(result.Value, Is.Not.Null.And.Not.Empty);
@@ -1384,7 +1384,7 @@ public class RealtimeTests : RealtimeTestFixtureBase
             SessionOptions = transcriptionSessionOptions,
         };
 
-        CreateClientSecretResult result = client.CreateRealtimeClientSecret(createClientSecretOptions);
+        CreateClientSecretResult result = await client.CreateRealtimeClientSecretAsync(createClientSecretOptions, CancellationToken);
 
         Assert.That(result, Is.Not.Null);
         Assert.That(result.Value, Is.Not.Null.And.Not.Empty);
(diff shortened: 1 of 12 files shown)


tests/Realtime/RealtimeTests.cs
```diff
@@ -1262,7 +1262,7 @@ public async Task CreateConversationSessionClientSecret()
             SessionOptions = conversationSessionOptions,
         };
 
-        CreateClientSecretResult result = client.CreateRealtimeClientSecret(createClientSecretOptions);
+        CreateClientSecretResult result = await client.CreateRealtimeClientSecretAsync(createClientSecretOptions, CancellationToken);
```
jsquire, line 1266: 🟡 Please update the Realtime live tests so that they no longer run under both `isAsync` fixture variants. `GetTestClient()` returns a direct client for these `[LiveOnly]` tests rather than the test framework’s sync/async proxy. Now that these final synchronous calls have been converted, both fixture variants execute the same asynchronous path and duplicate every live WebSocket test. We should run only the async fixture so that the test count and coverage signal reflect the new API surface.

---
> [!NOTE]

z0rimo: Updated both Realtime live test fixtures to run only with the async fixture variant.

Verified with:

`dotnet test tests/OpenAI.Tests.csproj -f net10.0 --filter "TestCategory=Realtime"`

The test suite now discovers 20 Realtime tests instead of the previous duplicated 40; all 20 are skipped in Playback mode as expected.

jsquire (approved):
Thank you for your contribution, @z0rimo!    The proposed change looks good.   I'll get this merged in once CI completes.

z0rimo:
> Thank you for your contribution, @z0rimo! The proposed change looks good. I'll get this merged in once CI completes.

Thanks for the review and for getting this merged! I also appreciate the feedback on the Realtime test coverage.

PR #1377, opened 2026-09-14
fix: preserve API keys in Realtime client factory

jbeckwith-oai:
## Bug

`OpenAIClient.GetRealtimeClient()` reused the parent's HTTP pipeline but did not retain its API-key credential. Realtime WebSocket connections read the credential directly, so starting a conversation or transcription session through the factory-created client could fail authentication despite a valid parent API key, unless the caller supplied a session client secret or custom authentication headers.

## Fix

Pass the original credential reference through an internal constructor while continuing to reuse the HTTP pipeline and client options. Existing public signatures and session authentication overrides are preserved.

## Validation

- All 80 Realtime unit tests pass on .NET 10 using installed SDK 10.0.301; the repository pins 10.0.400.
- Library builds for `net10.0`, `net8.0`, and `netstandard2.0` with zero warnings or errors.
- The new credential regression test fails against the original factory behavior and passes with the fix.
- Regression coverage also checks shared pipeline identity, endpoint propagation, and policy-only construction.
- `git diff --check` passes. Formatting introduces no new diagnostics; the test file has eight existing whitespace diagnostics on the new base.


markstuart-oai (approved):
Reviewed the complete change at `8842ae8`. Passing the original credential reference through the internal constructor fixes the WebSocket authentication path while retaining the shared HTTP pipeline, existing options, credential rotation, and session-secret/header precedence. The focused factory tests cover credential and pipeline identity plus policy-only construction. This is a small, cohesive fix; independent review found no actionable correctness or structural issues.

Verified current-head hosted playback tests: 1,608 passed on .NET 9, 1,596 on .NET 8, and 1,610 on .NET 10; 89 skipped per target. All 11 reported checks passed. Local .NET was unavailable; no live WebSocket handshake was run.

PR #1381, opened 2026-09-16
Add new response models and update tool choice handling

jairmyree:
- Introduced ResponseCustomWebSearchToolChoice and ResponseDefaultToolChoice models for enhanced tool choice options.
- Updated serialization logic in ResponseResult to accommodate new tool choice structures.
- Implemented ResponseToolChoice serialization and deserialization methods.
- Refactored existing tests to utilize new response models and ensure compatibility with updated tool choice handling.
- Removed deprecated tool choice creation methods in favor of new constructors for better clarity and maintainability.


specification/base/typespec/responses/models.tsp
```diff
@@ -536,6 +536,15 @@ union ToolChoiceOptions {
   "required",
 }
 
+@access(Access.public)
+@usage(Usage.input | Usage.output)
```
joseharriaga, line 540: I believe these decorators are not needed here. For reference, these are only needed in cases where the model or union is not referenced by an operation (or by a model that is referenced by an operation, and so on), which I don't think is the case here.

OpenAI.Responses/src/Generated/Models/ResponseToolChoice.cs
```diff
@@ -0,0 +1,40 @@
...
+            CustomToolChoice = customToolChoice;
+            _patch = patch;
+            _patch.SetPropagators(PropagateSet, PropagateGet);
+        }
+#pragma warning restore SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
+
+        [JsonIgnore]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Experimental("SCME0001")]
+        public ref JsonPatch Patch => ref _patch;
```
joseharriaga, line 38: This PR made some small changes to how `RealtimeToolChoice` works:
🔗 https://github.com/openai/openai-dotnet/pull/1375

For example, it made `Patch` internal (among other things). Please take a look and integrate the same changes here. 

Additionally, you will also notice that each union in that PR also got a set of unit tests; I recommend adding that same set of tests for this union too.

specification/base/typespec/responses/models.tsp
```diff
@@ -586,6 +595,13 @@ model ToolChoiceObjectFunction extends ToolChoiceObject {
   name: string;
 }
 
+@access(Access.public)
+@usage(Usage.input | Usage.output)
+model DotNetResponseToolChoice {
+  default_tool_choice?: DotNetResponseDefaultToolChoice;
+  custom_tool_choice?: ToolChoiceObject;
```
joseharriaga, line 602: I think the current definition of `ToolChoiceObject` is missing a few extensions, such as "custom", "apply_patch", "allowed_tools", etc.

